PR #7, opened 2024-11-21
Add Surface Code

justinlietz:
This PR adds the `cudaq::qec::surface_code` namespace which includes the `surface_code` type and the `stabilizer_grid` helper class. The `stabilizer_grid` generates the 2d layout of the stabilizers, and which data qubits each stabilizer has support on. Lastly it generates the stabilizers and observables as vectors of `cudaq::spin_op` which the `qec::code` expects.


libs/qec/lib/codes/surface_code_device.cpp
```diff
@@ -0,0 +1,87 @@
...
+    z<cudaq::ctrl>(logicalQubitA.data[i], logicalQubitB.data[i]);
+  }
+}
+
+// Transversal state prep, turn on stabilizers after these ops
+__qpu__ void prep0(patch logicalQubit) {
+}
+
+__qpu__ void prep1(patch logicalQubit) {
+  prep0(logicalQubit);
```
amccaskey, line 36: do you need this prep0 call?

justinlietz: Not currently, but at the moment, all of our logical stateprep assumes starting from fresh qubits set to |0>. We should change the behavior of all of these so that they could work in the middle of an larger kernel execution. This means that the existing `prep0` should all have a round of resets at the beginning, then `prep1` should call `prep0` then the logical X gate.

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+private:
+  /// @brief Generates this->roles
+  void generate_grid_roles();
+  /// @brief Generates {x,z}_stab_coords and indices
+  void generate_grid_indices();
+  /// @brief Generates {x,z}_stabilizers
+  void generate_stabilizers();
+
+public:
+  /// @brief The distance of the code,
```
bmhowe23: Nit:
```suggestion
  /// @brief The distance of the code
```

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+#include "cudaq/qec/code.h"
+#include "cudaq/qec/patch.h"
+
+using namespace cudaqx;
+
+namespace cudaq::qec::surface_code {
+
+/// @brief enumerates the role of a grid site in the surface codes stabilizer
+/// grid
+enum surface_role { amx, amz, empty };
```
bmhowe23, line 19: Just curious: is this "a measure x" and "a measure z", or something else?

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+  void generate_grid_roles();
+  /// @brief Generates {x,z}_stab_coords and indices
+  void generate_grid_indices();
+  /// @brief Generates {x,z}_stabilizers
+  void generate_stabilizers();
+
+public:
+  /// @brief The distance of the code,
+  // determines the number of data qubits per dimension
+  uint32_t distance;
```
bmhowe23: It would be good to initialize this for cases where the default constructor is used.
```suggestion
  uint32_t distance = 0;
```

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+
+  /// @brief length of the stabilizer grid
+  // for distance = d data qubits,
+  // the stabilizer grid has length d+1
+  uint32_t grid_length;
```
bmhowe23: It would be good to initialize this for cases where the default constructor is used.
```suggestion
  uint32_t grid_length= 0;
```

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+/// \pure_device_kernel
+///
+/// @brief Perform stabilizer measurements on a surface_code patch
+/// @param p The patch to measure
+/// @param x_stabilizers Indices of X stabilizers to measure
+/// @param z_stabilizers Indices of Z stabilizers to measure
+/// @return Vector of measurement results
+__qpu__ std::vector<cudaq::measure_result>
+stabilizer(patch p, const std::vector<std::size_t> &x_stabilizers,
+           const std::vector<std::size_t> &z_stabilizers);
```
bmhowe23, line 229: I didn't think we supported pass-by-reference going into a `__qpu__` function?

libs/qec/unittests/test_qec.cpp
```diff
@@ -502,6 +503,99 @@ TEST(QECCodeTester, checkRepetition) {
...
+    cudaqx::tensor<uint8_t> parity = surf_code->get_parity();
+    cudaqx::tensor<uint8_t> parity_x = surf_code->get_parity_x();
+    cudaqx::tensor<uint8_t> parity_z = surf_code->get_parity_z();
+    cudaqx::tensor<uint8_t> observables =
+        surf_code->get_pauli_observables_matrix();
+    cudaqx::tensor<uint8_t> Lx = surf_code->get_observables_x();
+    cudaqx::tensor<uint8_t> Lz = surf_code->get_observables_z();
+
+    {
+      // This is just a regression check, this has not been hand tested
```
bmhowe23, line 523: What exactly does this comment mean? Where did the data come from?

libs/qec/unittests/test_qec.cpp
```diff
@@ -502,6 +503,99 @@ TEST(QECCodeTester, checkRepetition) {
...
+
+      std::vector<std::size_t> expected_shape{8, 18};
+      cudaqx::tensor<uint8_t> t(expected_shape);
+      t.borrow(data.data());
+      for (std::size_t i = 0; i < 8; i++)
+        for (std::size_t j = 0; j < 18; j++)
+          EXPECT_EQ(t.at({i, j}), parity.at({i, j}));
+    }
+    {
+      // This is just a regression check, this has not been hand tested
```
bmhowe23, line 542: Same question here.

libs/qec/unittests/test_qec.cpp
```diff
@@ -502,6 +503,99 @@ TEST(QECCodeTester, checkRepetition) {
...
+
+      std::vector<std::size_t> expected_shape{4, 9};
+      cudaqx::tensor<uint8_t> t(expected_shape);
+      t.borrow(data.data());
+      for (std::size_t i = 0; i < 4; i++)
+        for (std::size_t j = 0; j < 9; j++)
+          EXPECT_EQ(t.at({i, j}), parity_x.at({i, j}));
+    }
+    {
+      // This is just a regression check, this has not been hand tested
```
bmhowe23, line 556: Same question here.

bmhowe23 (commented):
Thanks, Justin. I have a few comments/questions below.

libs/qec/include/cudaq/qec/codes/surface_code.h
justinlietz: That is a fine interpretation, although in my head I was thinking "ancilla for mx/z".

libs/qec/include/cudaq/qec/codes/surface_code.h
justinlietz: What is the expected behavior of passing a ref in? This seems to work here as well as in the [Steane code](https://github.com/NVIDIA/cudaqx/blob/8a7f74e14300e41e1b5705175c945040eee3f4c5/libs/qec/lib/codes/steane_device.cpp#L73).

amccaskey: In this case, either pass by value or pass by ref get you the same thing - a subtlety of the ASTBridge MLIR generation. No matter what the signature is, if you pass a std::vector it is modeled as a cc.stdvec which lowers to a span-like type (a pointer and a size). You are always passing by reference no matter what. Eric can elaborate more if needed.

libs/qec/unittests/test_qec.cpp
justinlietz: The data is only generated from itself, just scraped from the terminal and put into a vector here. Other instances of checking against hardcoded bit strings were hand checked to be correct. This one was not, as there are not printed instances of surface code PCMs I could find.

This comment is to make us aware that these are not magic numbers that we must always match. Various permutations would all be valid PCMs of the surface code.

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+};
+
+vec2d operator+(const vec2d &lhs, const vec2d &rhs);
+vec2d operator-(const vec2d &lhs, const vec2d &rhs);
+bool operator==(const vec2d &lhs, const vec2d &rhs);
+bool operator<(const vec2d &lhs, const vec2d &rhs);
+
+/// @brief Generates and keeps track of the 2d grid of stabilizers in the
+/// rotated surface code
+// Grid layout is arranged from left to right, top to bottom (row major storage)
```
bmhowe23: We need to double check that the docs are rendering as desired for `//` vs `///`.

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+// (0,0)   (0,1)   (0,2)   (0,3)
+// (1,0)   (1,1)   (1,2)   (1,3)
+// (2,0)   (2,1)   (2,2)   (2,3)
+// (3,0)   (3,1)   (3,2)   (3,3)
+//
+// Each entry on the grid can be an X stabilizer, Z stabilizer,
+// or empty, as is needed on the edges.
+// The grid length of 4 corresponds to a distance 3 surface code, which results
+// in: e(0,0)  e(0,1)  Z(0,2)  e(0,3) X(1,0)  Z(1,1)  X(1,2)  e(1,3) e(2,0)
+// X(2,1)  Z(2,2)  X(2,3) e(3,0)  Z(3,1)  e(3,2)  e(3,3)
```
bmhowe23: Double-check formatting.

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+// and to get the familiar visualization of the distance three surface code, the
+// `print_stabilizer_indices` results in:
+//         Z0
+// X0  Z1  X1
+//     X2  Z2  X3
+//     Z3
+//
+// The data qubits are located at the four corners of each of the weight-4
+// stabilizers. They are also organized with index increasing from left to
+// right, top to bottom: d0  d1  d2 d3  d4  d5 d6  d7  d8
```
bmhowe23: Double check formatting (two spaces vs single spac)

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -0,0 +1,274 @@
...
+  vec2d(int row_in, int col_in);
+};
+
+vec2d operator+(const vec2d &lhs, const vec2d &rhs);
+vec2d operator-(const vec2d &lhs, const vec2d &rhs);
+bool operator==(const vec2d &lhs, const vec2d &rhs);
+bool operator<(const vec2d &lhs, const vec2d &rhs);
+
+/// @brief Generates and keeps track of the 2d grid of stabilizers in the
+/// rotated surface code
```
bmhowe23: Perhaps consider adding link to a reference paper who's convention you're following w.r.t. the grid.

libs/qec/lib/codes/surface_code.cpp
```diff
@@ -0,0 +1,396 @@
+/****************************************************************-*- C++ -*-****
```
bmhowe23: Nit: 
```suggestion
/*******************************************************************************
```

libs/qec/lib/codes/surface_code_device.cpp
```diff
@@ -0,0 +1,87 @@
+/****************************************************************-*- C++ -*-****
```
bmhowe23: Nit:
```suggestion
/*******************************************************************************
```

bmhowe23 (commented):
Notes from our meeting today.

bmhowe23:
I don't know if you're seeing this in your local builds or not, but the CI is reporting some potentially relevant warnings:
https://github.com/NVIDIA/cudaqx/actions/runs/12943122467/job/36166508865?pr=7#step:10:232

E.g., this is repeated many times:
```
In file included from /__w/cudaqx/cudaqx/libs/qec/unittests/test_qec.cpp:14:
/__w/cudaqx/cudaqx/libs/qec/include/cudaq/qec/codes/surface_code.h:164:23: warning: ‘annotate’ attribute directive ignored [-Wattributes]
  164 | __qpu__ void x(patch p);
```

justinlietz:
> I don't know if you're seeing this in your local builds or not, but the CI is reporting some potentially relevant warnings: https://github.com/NVIDIA/cudaqx/actions/runs/12943122467/job/36166508865?pr=7#step:10:232
> 
> E.g., this is repeated many times:
> 
> ```
> In file included from /__w/cudaqx/cudaqx/libs/qec/unittests/test_qec.cpp:14:
> /__w/cudaqx/cudaqx/libs/qec/include/cudaq/qec/codes/surface_code.h:164:23: warning: ‘annotate’ attribute directive ignored [-Wattributes]
>   164 | __qpu__ void x(patch p);
> ```

I do not see this locally, but I build with `-DCMAKE_CXX_FLAGS="-Wno-attributes"`. Turning this flag off, I do now see this warning, but I'm not sure why `surface_code` is any different than `steane` in this regard.

bmhowe23:
> > I don't know if you're seeing this in your local builds or not, but the CI is reporting some potentially relevant warnings: https://github.com/NVIDIA/cudaqx/actions/runs/12943122467/job/36166508865?pr=7#step:10:232
> > E.g., this is repeated many times:
> > ```
> > In file included from /__w/cudaqx/cudaqx/libs/qec/unittests/test_qec.cpp:14:
> > /__w/cudaqx/cudaqx/libs/qec/include/cudaq/qec/codes/surface_code.h:164:23: warning: ‘annotate’ attribute directive ignored [-Wattributes]
> >   164 | __qpu__ void x(patch p);
> > ```
> 
> I do not see this locally, but I build with `-DCMAKE_CXX_FLAGS="-Wno-attributes"`. Turning this flag off, I do now see this warning, but I'm not sure why `surface_code` is any different than `steane` in this regard.

It turns out this warning is only coming from the `test_qec.cpp` file, not our actual libraries. Our libs are built with `-Wno-attributes`, but our tests are not. `test_qec.cpp` does not `#include` the Steane code, so the tests are not currently reporting this warning. The prior QEC tests just used the pure QEC interfaces. I think we should add `add_compile_options(-Wno-attributes)` to `libs/qec/unittests/CMakeLists.txt` (similar to how we already do that in `libs/qec/lib/CMakeLists.txt`).

bmhowe23 (approved):
Thanks, @justinlietz - these changes look great!

PR #9, opened 2024-11-22
[ci] Fixes to PR and PR cleanup workflows.

boschmitt:
* Fix paths that triggers different library testing jobs.
* Fix condition to run library testing jobs.
* There is no need to install GitHub CLI tool on github runners.


.github/workflows/pr_workflow.yaml
```diff
@@ -109,21 +116,31 @@ jobs:
           save-build: true
           save-ccache: false
 
+  # In the following, we need the `always()` fucntion in order to guarantee
+  # that these jobs will run when their conditions are met, even if the
+  # `build-cudaq` job is skipped.
```
bmhowe23: I could be wrong, but I believe this means they will always run, even if build-cudaq runs and fails. Is that intentional?

boschmitt: Indeed, while these would still "work", it would be a waste of resources. I do think we need the `always()`, otherwise the condition would not even be evaluated. However, I think it should be possible to improve with something like:

```
needs.build-cudaq.result == 'skipped' || 
needs.build-cudaq.result == 'success'
```

Edit: or `!failure() && !cancelled()`.

PR #13, opened 2024-11-25
Add a get_operator_pool function in C++ to mirror the Python

amccaskey:


libs/solvers/lib/operators/operator_pools/operator_pool.cpp
```diff
@@ -9,3 +9,10 @@
 #include "cudaq/solvers/operators/operator_pool.h"
 
 INSTANTIATE_REGISTRY_NO_ARGS(cudaq::solvers::operator_pool)
+
+namespace cudaq::solvers {
+std::vector<cudaq::spin_op>
+get_operator_pool(const std::string &name, const heterogeneous_map &options) {
+  return cudaq::solvers::operator_pool::get(name)->generate(options);
```
bmhowe23, line 16: At first, I was concerned that this could lead to a null pointer dereference if `name` wasn't found, but I see that we throw an exception before that would happen, so I guess that is not an issue.

bmhowe23 (approved):
LGTM 👍

docs/sphinx/examples/solvers/cpp/adapt_h2.cpp
```diff
@@ -24,16 +24,16 @@ int main() {
   auto h = molecule.hamiltonian;
 
   // Create the operator pool
-  auto pool = cudaq::solvers::operator_pool::get("spin_complement_gsd");
-  auto poolList = pool->generate({{"num-orbitals", h.num_qubits() / 2}});
+  auto opPool = cudaq::solvers::get_operator_pool(
+      "spin_complement_gsd", {{"num-orbitals", h.num_qubits() / 2}});
```
boschmitt: Consider not using `auto` here. Its use can be quite handy on certain situations, e.g., when declaring lambda functions or taking iterators from containers, but makes code review much harder.

In this PR, its easy to get by because there is a change that shows the signature of `get_operator_pool`, thus if I look further down I can see its a `std::vector<cudaq::spin_op>`.

libs/solvers/unittests/test_operator_pools.cpp
```diff
@@ -26,6 +26,17 @@ TEST(UCCSDTest, GenerateWithDefaultConfig) {
   }
 }
 
+TEST(UCCSDTest, GenerateFromAPIFunction) {
+  auto operators = cudaq::solvers::get_operator_pool(
```
boschmitt, line 30: Consider not using `auto` here.

melody-ren (approved):
Looks good to me

boschmitt (approved):

PR #15, opened 2024-11-27
[ci] Add workflow to check c++ code formatting.

### LINKED ISSUE #14: Add a clang-format check to the CI

boschmitt:
Closes #14 

EDIT: This PR depends on #21

### DIFF
--- a/.github/workflows/pr_sanity_checks.yaml
+++ b/.github/workflows/pr_sanity_checks.yaml
@@ -1,0 +1,103 @@
+name: PR sanity checks
+
+on:
+  pull_request:
+    branches: 
+      - main
+    types: [opened, synchronize, reopened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  LLVM_VERSION: 16
+
+jobs:
+  check-changes:
+    name: Check changes
+    runs-on: ubuntu-latest
+    outputs:
+      check-cpp: ${{ steps.filter.outputs.check-cpp }}
+      check-all-cpp: ${{ steps.filter.outputs.check-all-cpp }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          set-safe-directory: true
+
+      - name: Check what needs testing
+        uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            check-all-cpp:
+              - '.github/workflows/pr_sanity_checks.yaml'
+              - '.clang-format'
+            check-cpp:
+              - '**/*.cpp'
+              - '**/*.h'
+
+  check-clang-format:
+    name: Check C++ code formatting
+    needs: [check-changes]
+    if: needs.check-changes.outputs.check-cpp == 'true' || needs.check-changes.outputs.check-all-cpp == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          set-safe-directory: true
+
+      - name: Install clang-format
+        run: |
+          # Requirements
+          sudo apt-get update
+          sudo apt-get install -y wget software-properties-common gpg
+
+          # Obtain VERSION_CODENAME and UBUNTU_CODENAME
+          source /etc/os-release
+
+          wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
+          sudo add-apt-repository "deb http://apt.llvm.org/${UBUNTU_CODENAME}/  llvm-toolchain-${UBUNTU_CODENAME}-${LLVM_VERSION} main"
+          sudo apt-get update && sudo apt-get install -y --no-install-recommends clang-format-${LLVM_VERSION}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_sanity_checks.yaml
```diff
@@ -0,0 +1,84 @@
...
+    branches: 
+      - main
+    types: [opened, synchronize, reopened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  LLVM_VERSION: 19
```
bmhowe23: I'd recommend keeping this in sync with the LLVM version that we're currently using (16).

.github/workflows/pr_sanity_checks.yaml
```diff
@@ -0,0 +1,84 @@
...
+    outputs:
+      check-cpp: ${{ steps.filter.outputs.check-cpp }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          set-safe-directory: true
+
+      - name: Check what needs testing
+        uses: dorny/paths-filter@v3
```
bmhowe23, line 30: Is there any reason this can't be done with a find command (rather than taking on a new dependency)?

.github/workflows/pr_sanity_checks.yaml
```diff
@@ -0,0 +1,84 @@
...
+        DIFF_COMMIT_SHA=$(git rev-parse origin/${{ github.base_ref }})
+
+        git config clangFormat.binary clang-format-$LLVM_VERSION
+        if ! git clang-format $DIFF_COMMIT_SHA; then
+          git diff --ignore-submodules > clang-format.patch
+          echo "::error::Clang-format found formatting problems." \
+                "Check the uploaded artifact."
+          exit 1
+        fi
+        echo "Clang-format found no formatting problems"
```
bmhowe23: It would probably be helpful to add some of these status-related items to the GITHUB_STEP_SUMMARY so that they show up on the summary page.

bmhowe23 (commented):
Thanks for pulling this together, Bruno.

.github/workflows/pr_sanity_checks.yaml
boschmitt: The `::error` command will show in the summary.

.github/workflows/pr_sanity_checks.yaml
boschmitt: This step is not simply looking for `*.cpp` and `*.h` files in the repository. It is looking for these files _among the list of files changed by the PR_. We could hand roll the functionality. but it is a bit error prone.

bmhowe23: I believe cuda-quantum has already implemented and tested this for us: cuda-quantum/.github/workflows/repo_checks.yml. Or we could simply process the entire repo (do `git ls-files` and filter for the C++ files similar to find). I believe running clang-format is quite fast and running it on more files than strictly necessary would likely have no noticeable impact on CI time.

.github/workflows/pr_sanity_checks.yaml
bmhowe23: OK, that's good. Could we please also add the above message to the summary file when everything runs cleanly?

.github/workflows/pr_sanity_checks.yaml
boschmitt: The two main differences between this workflow and CUDAQ's similar one are: the use `dorny/paths-filter` and the use of `git clang-format`. CUDAQ's chose to hand-roll similar functionalities. IMHO, doing like here has more advantages:

- **It saves times**: Yes, we could get the same initial saving by reusing CUDAQ's one but there is also a question of maintenance.
- **Maintenance**: Using a well-maintained tools ensures bug fixes and improvements without requiring manual updates to custom scripts. (Keep in mind that `dorny/paths-filter` is part of a very restricted list of trusted actions allowed to be used in our workflows.)
- **It reduces complexity**: These tools are widespread in the community and well documented.
- **Leverages community-tested code** 

In the case of `dorny/paths-filter`, we also have the advantage of reusability. We already use the action in the PR-testing workflow. Using it here again promotes consistency and reduce code duplication (I'm assuming that removing it from here implies removing from there as well).

In the case of `git clang-format`, we also get the benefit of only ensuring correct formatting of the changes. This minimizes unintended changes to code a PR didn't modify, reducing noise in the diff and making code reviews easier. (IMO, CUDAQ already provides an example where its formatting check would be annoying, i.e., if we fix the `.clang-format` file to substitute `qoda` by `cudaq`, then changes to a C++ file that don't touch the include files might need to fix their formatting.)

bmhowe23: I am fine with keeping it if it is pre-approved.

bmhowe23: Just echoing something from an offline discussion - I would like for our general policy to be that if we make changes to .clang-format, that we reformat the entire repo (and telling git to ignore the formatting commit when blaming) either in that same commit or in the very next commit. This current approach of using `dorny/paths-filter` for only analyzing the diffs for the per-PR processing will allow us to accidentally defer formatting of files to future PRs, and I don't want those future PR's to have mixes of formatting changes + content changes.

bmhowe23 (approved):
Thanks!

PR #16, opened 2024-12-02
Make DecoderResult tuple-like in python

### LINKED ISSUE #4: Make DecoderResult tuple-like in Python
Update the DecoderResult python binding to support "structured binding" capture in Python. I.e. 
```python
converged, result = decoder.decode(...) 
```

Look at the QAOAResult in the solvers bindings for an example on how to do this.

melody-ren:
Resolves #4

### DIFF
--- a/libs/qec/python/tests/test_decoder.py
+++ b/libs/qec/python/tests/test_decoder.py
@@ -36,11 +36,11 @@ def test_decoder_result_structure():
 
 def test_decoder_result_values():
     decoder = qec.get_decoder('example_byod', H)
-    result = decoder.decode(create_test_syndrome())
-    
-    assert result.converged is True
-    assert all(isinstance(x, float) for x in result.result)
-    assert all(0 <= x <= 1 for x in result.result)
+    convergence, result = decoder.decode(create_test_syndrome())
+
+    assert convergence is True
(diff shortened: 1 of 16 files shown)


bmhowe23:
Thanks, @melody-ren! Could you please update the description to "Fixes #4" or "Resolves #4" so that when this PR is merged, it will automatically close the GitHub issue? (It will also link the PR to the issue so that anybody who views the issue will see that this is a possible fix for the issue, even before it is merged.)

libs/qec/python/tests/test_decoder.py
```diff
@@ -80,13 +80,13 @@ def test_decoder_reproducibility():
     decoder = qec.get_decoder('example_byod', H)
     
     np.random.seed(42)
-    result1 = decoder.decode(create_test_syndrome())
+    convergence1, result1 = decoder.decode(create_test_syndrome())
     
     np.random.seed(42)
-    result2 = decoder.decode(create_test_syndrome())
+    convergence2, result2 = decoder.decode(create_test_syndrome())
```
bmhowe23, line 86: Can you please keep one of the tests using the old interface so that we ensure both methods (tuple-like and struct-like) are supported?

bmhowe23 (approved):
Thanks, @melody-ren!

PR #23, opened 2024-12-05
Add an option to set the tolerance for jordan_wigner

### LINKED ISSUE #3: Add threshold option to Jordan-Wigner fermion_compiler

melody-ren:
Resolves #3

### DIFF
--- a/libs/solvers/lib/operators/molecule/fermion_compilers/jordan_wigner.cpp
+++ b/libs/solvers/lib/operators/molecule/fermion_compilers/jordan_wigner.cpp
@@ -390,7 +390,7 @@ cudaq::spin_op jordan_wigner::generate(const double constant,
   std::size_t nqubit = hpq.shape()[0];
 
   double tolerance =
-      options.get({"tolerance", "tol", "TOLERANCE", "TOL"}, 1e-15);
+      options.get<double>(std::vector<std::string>{"tolerance", "tol"}, 1e-15);
 
   for (auto p : cudaq::range(nqubit)) {
     auto coef = hpq.at({p, p});
--- a/libs/solvers/python/bindings/solvers/py_solvers.cpp
+++ b/libs/solvers/python/bindings/solvers/py_solvers.cpp
@@ -238,6 +238,7 @@ core_energy : float, optional
     The core energy of the system when using active space Hamiltonian, nuclear energy otherwise. Default is 0.0.
 tolerance : float, optional
     The threshold value for ignoring small coefficients.
+    Can also be specified using 'tol'.
     Coefficients with absolute values smaller than this tolerance are considered as zero.
     Default is 1e-15.
 
@@ -311,6 +312,7 @@ core_energy : float, optional
     The core energy of the system. Default is 0.0.
 tolerance : float, optional
     The threshold value for ignoring small coefficients.
+    Can also be specified using 'tol'.
     Coefficients with absolute values smaller than this tolerance are considered as zero.
     Default is 1e-15.
 


melody-ren:
I think there are still some questions about whether move-assignment was intentionally disabled in the Heterogeneous map class. If so, we should document the reason. 

Thoughts, Alex? @amccaskey

libs/core/include/cuda-qx/core/heterogeneous_map.h
```diff
@@ -65,8 +65,10 @@ class heterogeneous_map {
   /// @param _other The map to assign from
   /// @return Reference to this map
   heterogeneous_map &operator=(const heterogeneous_map &_other) {
-    clear();
-    items = _other.items;
+    if (this != &_other) {
+      clear();
+      items = _other.items;
+    }
```
melody-ren, line 67: This was added to prevent self assignment

libs/core/include/cuda-qx/core/heterogeneous_map.h
```diff
@@ -76,20 +78,14 @@ class heterogeneous_map {
...
-      if constexpr (is_bounded_char_array<T>{}) {
-        items.insert({key, std::string(value)});
-        return;
-      }
-
-      items.insert({key, value});
-      return;
+      items.insert_or_assign(key, std::string(value));
+    } else {
+      items.insert_or_assign(key, value);
```
melody-ren, line 83: Just some minor updates to improve readability

libs/solvers/lib/operators/operator_pools/uccsd_operator_pool.cpp
```diff
@@ -18,7 +18,8 @@ using excitation_list = std::vector<std::vector<std::size_t>>;
 std::vector<cudaq::spin_op>
 uccsd::generate(const heterogeneous_map &config) const {
 
-  auto numQubits = config.get<int>({"num-qubits", "num_qubits"});
+  auto numQubits =
+      config.get<int>({"num-qubits", "num_qubits", "n-qubits", "n_qubits"});
   auto numElectrons = config.get<int>({"num-electrons", "num_electrons"});
```
melody-ren, line 23: Added two more variations of "n_qubits" to match that of QAOA's definition to keep the user interface consistent

libs/solvers/lib/operators/molecule/fermion_compilers/jordan_wigner.cpp
```diff
@@ -382,12 +382,16 @@ cudaq::spin_op two_body(std::size_t p, std::size_t q, std::size_t r,
...
+                                       const tensor<> &hpqrs,
+                                       const heterogeneous_map &options) {
   assert(hpq.rank() == 2 && "hpq must be a rank-2 tensor");
   assert(hpqrs.rank() == 4 && "hpqrs must be a rank-4 tensor");
   auto spin_hamiltonian = constant * cudaq::spin_op();
   std::size_t nqubit = hpq.shape()[0];
-  double tolerance = 1e-15;
+
+  double tolerance =
+      options.get({"tolerance", "tol", "TOLERANCE", "TOL"}, 1e-15);
```
bmhowe23: I think we should probably reduce down to just supporting `"tolerance"` and maybe `"tol"`, but I think we should remove the upper-case versions. In addition, if we are going to support both "tolerance" and "tol" keywords, we should document both of them in the associated pydoc strings.

melody-ren: Good point about the docstring. Will update. Also no problem with reducing down to lowercase only.

bmhowe23 (approved):
Looks good...thanks, Melody!

PR #31, opened 2024-12-12
Change default decoder float type to double

bmhowe23:
1. Change default decoder float type to double.
2. Define equality and inequality operators for `decoder_result`.
3. Cleanup CMakeLists.txt to use var instead of repeated name.


wsttiger (approved):
LGTM

PR #32, opened 2024-12-12
Add CUDA build support to libraries

bmhowe23:
This comes from https://github.com/NVIDIA/cuda-quantum/blob/105c05a15c5533cf2aadb0ede263f110f44426a0/CMakeLists.txt#L497-L533


boschmitt:
Since libraries can be built separately, we should add this on the libraries. (When built separately, libs don't access to the top CMakeLists.)

amccaskey (approved):

PR #42, opened 2024-12-18
Add an example of loading a decoder .so

melody-ren:


libs/qec/lib/decoders/plugins/example/decoder_plugins_demo.cpp
```diff
@@ -0,0 +1,113 @@
...
+  auto Lz = steane->get_observables_x();
+  std::cout << "Lz:\n";
+  Lz.dump();
+
+  double p = 0.2;
+  size_t nShots = 5;
+
+  // load the decoder plugins
+  load_plugins(".");
+  // create a decoder from the plugins
```
melody-ren, line 50: Loading the plugins still requires a loader header with some dlopen code. I haven't figured out how to bundle this nicely such that we don't need this "load_plugins" in the client code. Suggestions?

amccaskey: Should be noted that the following code will run and print your example decoder plugin without the `load_plugins()` function call 
```cpp
#include "cudaq/qec/decoder.h"

int main() {
  for (auto &name : cudaq::qec::decoder::get_registered())
    printf("Decoder: %s\n", name.c_str());
}
```
```bash
nvq++ -lcudaq-qec -lcudaq-qec-example  plugin.cpp
./a.out
  Decoder: single_error_lut_example
  Decoder: single_error_lut
```
Note this requires installing cudaqx (which we by default want to be installed to the cudaq install prefix), and using the installed nvq++ (not one from a build directory). This is what I meant by the decoder already exposing a default static registry pattern. 

Now there are a few things we could do next. First, maybe we are ok with this extra library on the compile line? I mean we already require someone pass the cudaq-qec library. Second, we could write a bash utility (we did historically have one, but dropped it before the release) that automates the addition of these libraries, e.g. 
```bash
nvq++ -lcudaq-qec $(cudaqx-util --add-decoder-plugins) plugin.cpp
```
Third, we could put some statically constructed code in the cudaq-qec library that runs at app startup and does something similar to your `load_plugins()` function.

amccaskey: For the third approach, something like this works in `qec/lib/decoder.cpp`
```cpp

__attribute__((constructor)) void acquireDecoderPlugins() {
  namespace fs = std::filesystem;
  bool success = true;
  for (const auto &entry : fs::directory_iterator("/home/.cudaq/lib/decoder-plugins")) {
    if (entry.path().extension() == ".so") {
      std::cout << "Loading plugin: " << entry.path() << std::endl;

      // Open the shared library
      void *handle = dlopen(entry.path().c_str(), RTLD_NOW);
      if (!handle) {
        std::cerr << "Failed to load plugin: " << entry.path()
                  << "Error: " << dlerror() << std::endl;
        success = false;
      } else {
        // Close the shared library
        dlclose(handle);
      }
    }
  }
}
```
Note I just hardcoded the path and I changed the install directory in the Cmake file to this `lib/decoder-plugins`. I would think this would make the Python side work automatically, but I haven't tested that . 

You may want to store the `handles` somewhere and add a `destructor` function that cleans them up.

bmhowe23: I'd vote for the 3rd approach.

melody-ren: Python side works automatically with the 3rd approach

CMakeLists.txt
```diff
@@ -67,7 +67,10 @@ add_custom_target(cudaqx-pymodules)
 
 if (CUDAQX_INCLUDE_TESTS)
   include(CTest)
+  # Create the directory for decoder plugins before GoogleTest Discovery is run
+  file(MAKE_DIRECTORY ${CUDAQ_INSTALL_DIR}/lib/decoder-plugins)
```
bmhowe23: Is there any harm in just requiring that the plugins be placed in the same `lib` directory as all the other .so's? (In the short term, I think that is preferrable, and in the long term, we could search something like `$HOME/.cudaqx/decoder-plugins`, but I would be ok saving that for later.)

amccaskey: It's easier to just dlopen the plugin so's if they are in a separate directory. If we start in lib we'll have to have logic in the loader function that screens out existing libraries.

melody-ren: Agreed. This way it's easier to load the right .so files (decoder, code and so on)

bmhowe23 (commented):
Thanks, @melody-ren. I think this looks great. Could you please add a Python test that demonstrates this working as well? I think you said it should work without any further changes?

libs/qec/lib/decoders/plugins/example/decoder_plugins_demo.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <vector>
+
+#include "cudaq.h"
+#include "cudaq/qec/decoder.h"
+#include "cudaq/qec/experiments.h"
+
+int main() {
+  auto steane = cudaq::qec::get_code("steane");
+  auto Hz = steane->get_parity_z();
+  cudaqx::heterogeneous_map params;
```
amccaskey: is this used anywhere? looks like it's not needed

amccaskey (approved):

libs/qec/lib/decoders/plugins/example/decoder_plugins_demo.cpp
melody-ren: Indeed not used. Will remove

bmhowe23 (approved):
This LGTM. The CI timed out for what looks to be an unrelated reason, so I kicked it off to see if the failure is repeatable.

PR #43, opened 2024-12-19
Update build defaults to support typical dev needs

bmhowe23:
Make some of the typically used items be "opt out" rather than "opt in". All existing options are still supported if people need to disable specific portions of the build.


melody-ren (approved):
Looks good

amccaskey (approved):

PR #49, opened 2025-01-10
Updates to the build wheels scripts

bmhowe23:
Successfully ran here: https://github.com/NVIDIA/cudaqx/actions/runs/12716632282

1/10 Update: https://github.com/NVIDIA/cudaqx/actions/runs/12719268312

1/13 Update: https://github.com/NVIDIA/cudaqx/actions/runs/12753763150

### DIFF
--- a/.github/workflows/scripts/build_wheels.sh
+++ b/.github/workflows/scripts/build_wheels.sh
@@ -69,6 +69,7 @@ parse_options "$@"
 
 python_version=3.10
 python=python${python_version}
+ARCH=$(uname -m)
 
 # We need to use a newer toolchain because CUDA-QX libraries rely on c++20
 source /opt/rh/gcc-toolset-11/enable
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/build_wheels.sh
```diff
@@ -82,21 +83,16 @@ export CXX=g++
 
 cd libs/qec
 
-SKBUILD_CMAKE_ARGS="-DCUDAQ_DIR=$cudaq_prefix/lib/cmake/cudaq;-DCMAKE_CXX_COMPILER_EXTERNAL_TOOLCHAIN=/opt/rh/gcc-toolset-11/root/usr/lib/gcc/x86_64-redhat-linux/11/" \
+SKBUILD_CMAKE_ARGS="-DCUDAQ_DIR=$cudaq_prefix/lib/cmake/cudaq"
+SKBUILD_CMAKE_ARGS+=";-DCMAKE_CXX_COMPILER_EXTERNAL_TOOLCHAIN=/opt/rh/gcc-toolset-11/root/usr/lib/gcc/${ARCH}-redhat-linux/11/"
+SKBUILD_CMAKE_ARGS+=";-DCMAKE_BUILD_TYPE=$build_type"
+export SKBUILD_CMAKE_ARGS
```
bmhowe23, line 89: Other than line wrapping changes, the main change in this section was to set `-DCMAKE_BUILD_TYPE=$build_type`.

.github/workflows/scripts/build_wheels.sh
```diff
@@ -82,21 +83,16 @@ export CXX=g++
...
-  --exclude libcudaq-python-interop.so \
-  --exclude libcudaq-ensmallen.so \
-  --exclude libcudaq-common.so \
-  --exclude libcudaq-platform-default.so \
-  --exclude libnvqir-qpp.so \
-  --exclude libnvqir.so \
-  --exclude libcudaq.so \
-  --exclude libcudaq-spin.so \
-  --exclude libcudaq-nlopt.so \
+$python -m auditwheel -v repair dist/*.whl $CUDAQ_EXCLUDE_LIST \
```
bmhowe23, line 95: I made this `exclude` list change because the build process was failing if the wheels were built on a machine with GPUs because the default CUDA-Q target on a machine with GPUs is the `nvidia`, which means the default linking was to `libnvqir-custatevec...` rather than `libnvqir-qpp`, and therefore the prior exclusion list was incomplete. I could've made it complete by adding `libnvqir-custatevec...` to the exclusion list but figured it would be more thorough to simply exclude ALL CUDA-Q libraries from our wheels.

melody-ren (approved):

PR #55, opened 2025-01-14
[DRAFT] Build CUDA-Q wheels

bmhowe23:
Something like this will be needed to properly test our CUDA-QX wheels with the desired CUDA-Q wheels on an ongoing basis.

(This currently includes changes from another PR as well.)


bmhowe23:
Closed in lieu of https://github.com/NVIDIA/cudaqx/pull/56

PR #56, opened 2025-01-17
Update build wheels job to test the wheels

bmhowe23:
## Primary Changes
* Update build_wheels.yaml to test CUDA-QX wheels after building them.
* Update build_wheels.yaml to build CUDA-Q wheels rather than using released PyPi CUDA-Q wheels. This was necessary because the current 0.9.1 CUDA-Q wheels have an ARM bug that were impacting our ability to do testing. This was fixed in https://github.com/NVIDIA/cuda-quantum/pull/2504, which has not been released yet.
* Fix RPATH of decoder plugin example (thanks for the suggestion, @melody-ren) to fix a wheel packaging problem where `auditwheel` as duplicating (aka grafting) the `libcudaq-qec.so` file, causing `extension_point` registration anomalies and decoder plugin lookup failures.

## Secondary Changes
* In .github/actions/get-cudaq-version/action.yaml, skip `jq` install if it is already installed
* Add get-cudaq-wheels action; currently only invoked on manually commanded build_wheels.yaml workflow but we may decide to invoke this on a per-PR basis in the future.
* Change retention period of all uploaded artifacts to 14 days
* Update `getCUDAQXLibraryPath` to work for both QEC and Solvers
* Change decoder-plugins path from a build-time constant (which will not work for wheels) to use `getCUDAQXLibraryPath` instead.
* Change use of shared_ptr to unique_ptr in decoder plugin handling
* Move decoder plugin helper functions into namespaces rather than global namespace
* Change cudaqx wheels to be depending on cuda-quantum >= 0.9 rather than ~= 0.9.0 for better futureproofing of our released wheels

Sample run result w/ custom CUDA-Q wheels: https://github.com/NVIDIA/cudaqx/actions/runs/12855925162
Sample run result w/ released PyPI CUDA-Q wheels: https://github.com/NVIDIA/cudaqx/actions/runs/12856394702


libs/core/include/cuda-qx/core/library_utils.h
```diff
@@ -20,9 +20,12 @@
 
 namespace cudaqx::__internal__ {
 
+enum class CudaQXLibType { Solvers, QEC };
```
melody-ren: Nit: Other instances of `CUDAQXLib*` seem to be using all cap for the `CUDAQX` part and the entire word of "library"

bmhowe23: Thanks ... updated in 74837ea.

melody-ren (approved):

PR #65, opened 2025-01-30
UCCSD operator pool correction and integration in adapt simulator

### LINKED ISSUE #28: fix uccsd pool operator for the adapt-vqe

kvmto:
# UCCSD operator pool correction and integration in adapt simulator (Issue #28)
This PR fixes #28 .

## Problem
Adapt VQE did not provide correct results. The problem could be the creation of the UCCSD operators pool.

## Changes
Changes so far have been applied to:
- The UCCSD operator pool 'generate' function.
- The Testing of the UCCSD operator pool, cpp and py files.
- Tested adap_vqe for final results.

## Additional notes 
Further testing is needed for making sure that adapt VQE works correctly, especially in terms of molecules.


libs/solvers/lib/operators/operator_pools/uccsd_operator_pool.cpp
```diff
@@ -39,8 +39,10 @@ uccsd::generate(const heterogeneous_map &config) const {
     for (std::size_t i = p + 1; i < q; i++)
       o *= cudaq::spin::z(i);
 
-    ops.emplace_back(cudaq::spin::y(p) * o * cudaq::spin::x(q));
-    ops.emplace_back(cudaq::spin::x(p) * o * cudaq::spin::y(q));
+    cudaq::spin_op t = cudaq::spin::y(p) * o * cudaq::spin::x(q) -
```
amccaskey: looks like you can drop this line since you just emplace_back right after

kvmto: Okay, noted. Thanks.

libs/solvers/unittests/test_operator_pools.cpp
```diff
@@ -46,12 +46,20 @@ TEST(UCCSDTest, GenerateWithCustomCoefficients) {
   auto operators = pool->generate(config);
```
amccaskey, line 50: I think you can just delete this GenerateWithCustomCoefficients test. Early on we had the idea that we'd allow someone to set the coefficients when they created the pull, but then decided to just let them do it manually with something like Python `zip()`

amccaskey:
Let's add a test to `test_adapt.cpp` that mimics the checkSimpleAdapt but uses the uccsd operator instead of the spin_complement_gsd one. Note that you'll need to update the coefficients on the operators to make them imaginary, since we assume that in adapt_simulator.cpp (line 152).

libs/solvers/unittests/test_operator_pools.cpp
kvmto: I think I will leave it for now. The presence of imaginary coeffcients is fundamental to the correct execution. I modiefied the test to verify that the coeff.s are imaginary :)

bmhowe23:
I would recommend adding the phrase "fixes #28" or "resolves #28" to the description in order to automatically link this GitHub PR with the issue. If you do this, then when the PR is merged, the issue will automatically be closed.

Ref: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

bmhowe23:
> Let's add a test to `test_adapt.cpp` that mimics the checkSimpleAdapt but uses the uccsd operator instead of the spin_complement_gsd one. Note that you'll need to update the coefficients on the operators to make them imaginary, since we assume that in adapt_simulator.cpp (line 152).

I think I'm missing something ... why we are updating the generic UCCSD operator pools from real to imaginary to accommodate a single assumption in adapt_simulator.cpp? Are any docs updates needed for this change?

marwafar:
The uccsd operator pool is correct. However, it might not be a good idea to multiply the imaginary number "i" by the uccsd operator pool. It will not make it generic. We might use this uccsd operator pool with other algorithm such as GQE and we will need to modify this again. I would suggest multiply the imaginary number "i" somewhere within the adapt-vqe algorithm. 

In addition, I would suggest to run the adapt-vqe for these two geometry with sto3g basis. 
```
H 0.0 0.0 0.0
H 0.0 0.0 0.7474
```

```
Li  0.0  0.0  0.0
H   0.0  0.0  1.5 
```
We need to make sure that it work for something beyond the simple H2 molecule with 4 qubits and 3 operator pools.

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -21,13 +21,15 @@ namespace cudaq::solvers::adapt {
...
-               const spin_op &H, const std::vector<spin_op> &pool,
+               const spin_op &H, const std::vector<spin_op> &poolOp,
                const optim::optimizer &optimizer, const std::string &gradient,
                const heterogeneous_map options) {
 
-  if (pool.empty())
+  if (poolOp.empty())
     throw std::runtime_error("Invalid adapt input, operator pool is empty.");
 
+  std::vector<spin_op> pool =
```
amccaskey: instead of making this copy, could you not retain the code that was there and just update the commutator creation to
```cpp
std::complex<double> j = 1.; 
if (isImaginary) j = std::complex<double>{0.0, 1.0};
commutators.push_back(H * j * pool[i] - j * pool[i] * H); 
```

libs/solvers/lib/stateprep/uccsd.cpp
```diff
@@ -507,18 +508,18 @@ __qpu__ void uccsd(cudaq::qview<> qubits, const std::vector<double> &thetas,
...
+        for (int s = r + 1; s < numVirtAlpha; s++)
           counter++;
 
   std::vector<std::size_t> doublesAlpha(4 * counter);
   counter = 0;
-  for (std::size_t p = 0; p < numOccAlpha - 1; p++)
-    for (std::size_t q = p + 1; q < numOccAlpha; q++)
-      for (std::size_t r = 0; r < numVirtAlpha - 1; r++)
-        for (std::size_t s = r + 1; s < numVirtAlpha; s++) {
+  for (int p = 0; p < numOccAlpha - 1; p++)
```
amccaskey: aren't these changes from Anna's PR? do you need her changes duplicated here?

libs/solvers/lib/adapt/adapt_simulator.cpp
kvmto: In line 159 (auto op = pool[maxOpIdx];), the code selects the best operator and manipulates it. We could assume that each operator in the pool contains either only real or only imaginary coefficients, in which case we would only need to perform the check on the first item. Alternatively, if we can’t make that assumption, we would need to store the result of the check for each operator and provide it in line 159. The copy is just a proposal to accommodate different assumptions. I would be happy with either.

libs/solvers/lib/stateprep/uccsd.cpp
kvmto: I don't need the duplication.
I just fetched and rebased, not sure how it ended up in here... I am not sure how changes happened in between.
I can reset the commit if needed.

libs/solvers/python/tests/test_adapt.py
```diff
@@ -1,69 +1,35 @@
...
-# ============================================================================ #
+#== == == == == == == == == == == == == == == == == == == == == == == == == == \
+    == == == == == == == == == == == == #
+#Copyright(c) 2024 NVIDIA Corporation &Affiliates.#
+#All rights reserved.#
+# #
+#This source code and the accompanying materials are made available under #
+#the terms of the Apache License 2.0 which accompanies this distribution.#
+#== == == == == == == == == == == == == == == == == == == == == == == == == == \
+    == == == == == == == == == == == == #
```
bmhowe23: It looks like the formatter went wild on this one. Did you accidentally run clang-format on this file?

libs/solvers/unittests/test_adapt.cpp
```diff
@@ -12,37 +12,160 @@
...
+  auto [energy3, thetas3, ops3] = cudaq::solvers::adapt_vqe(
+      hartreeFock2Electrons, hamhh, poolList,
+      {{"grad_norm_tolerance", 1e-2}, {"verbose", true}});
+  EXPECT_NEAR(energy3, -1.13, 1e-2);
+
+  poolList = pool->generate({{"num-orbitals", hamli.num_qubits() / 2}});
+  auto [energy2, thetas2, ops2] = cudaq::solvers::adapt_vqe(
+      statePrep4Electrons, hamli, poolList,
+      {{"grad_norm_tolerance", 0.5}, {"verbose", true}});
+  EXPECT_NEAR(energy2, -7.88, 1e-2);
```
bmhowe23: This test is too computationally intensive to run as part of our regular CI when no GPU is available. I will post a recommended change.

libs/solvers/unittests/test_adapt.cpp
```diff
@@ -12,37 +12,160 @@
...
+    hamhh = hh.hamiltonian;
+  }
+};
+
+TEST_F(SolversTester, checkSimpleAdapt) {
   auto pool = cudaq::solvers::operator_pool::get("spin_complement_gsd");
+
   auto poolList = pool->generate({{"num-orbitals", h.num_qubits() / 2}});
-  auto [energy, thetas, ops] = cudaq::solvers::adapt_vqe(
+  auto [energy1, thetash1, ops1] = cudaq::solvers::adapt_vqe(
```
bmhowe23: Numbering the variables as 1, 2, and 3 is unclear. I think this test should be split out for the different molecules and the 1, 2, and 3 should be removed. I will post a suggestion for this test. Will you please make the similar changes for the other tests?

libs/solvers/unittests/test_adapt.cpp
```diff
@@ -12,37 +12,179 @@
...
+  }
+  return static_cast<bool>(gpu_available);
+}
+
+class SolversTester : public ::testing::Test {
+protected:
+  cudaq::spin_op h;
+  cudaq::spin_op hamli;
+  cudaq::spin_op hamhh;
+  void SetUp() override {
```
bmhowe23: As I understand it, this method is called at the beginning of each test, but it looks like it only needs to be run once as far as I can tell. Perhaps it could be sped up with this change.
```suggestion
  void SetUp() override {
    if (!h.is_identity())
      return; // early return if we're already setup
```

libs/solvers/unittests/test_adapt.cpp
kvmto: Thanks for the suggestion!

libs/solvers/unittests/test_adapt.cpp
kvmto: Unfortunately the approach was not working. I used the proper of doing it in gtest :)

libs/solvers/include/cudaq/solvers/operators/molecule.h
```diff
@@ -103,6 +103,12 @@ struct molecule_options {
   /// default "RESTPySCFDriver"
   std::string driver = "RESTPySCFDriver";
 
+  /// @brief Fully qualified path to Python executable to use (if applicable).
```
bmhowe23: This change is not part of this PR...I think this is another case of the rebase not quite working for you. This makes it a bit difficult to review, so I think this needs to be fixed.

libs/solvers/unittests/test_adapt.cpp
```diff
@@ -210,16 +266,37 @@ TEST_F(SolversTester, checkSimpleAdaptGradientUCCSD_LiHSto3g) {
...
-      {{"grad_norm_tolerance", 0.5}, {"verbose", true}});
+      {{"grad_norm_tolerance", 1e-1}, {"verbose", true}});
   EXPECT_NEAR(energy, -7.88, 1e-2);
 
   for (std::size_t i = 0; i < thetas.size(); i++)
     printf("%lf -> %s\n", thetas[i], ops[i].to_string().c_str());
 }
+
+TEST_F(SolversTester, checkSimpleAdaptGradientUCCSD_BeH2Sto3g) {
+  if (!check_gpu_available())
```
amccaskey, line 473: If you look in the QEC unittests, you'll see we do have backend-specific tests in there. I think it would be good to add something like that here for the solvers.

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -51,9 +50,21 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
   std::size_t start = rank * elements_per_rank + std::min(rank, remainder);
   std::size_t end = start + elements_per_rank + (rank < remainder ? 1 : 0);
-  for (int i = start; i < end; i++) {
-    auto op = pool[i];
-    commutators.emplace_back(H * op - op * H);
+
+  // Check if operator has only imaginary coefficients
+  // checking the first one is enough, we assume the pool is homogeneous
+  const auto &c = pool[0].begin()->get_coefficient();
+  bool isImaginary = (c.real() == 0.0 && c.imag() != 0.0);
```
amccaskey: instead of direct equality checks like this for floating point we should do a check for is it zero within some precision, maybe something like 
```cpp
bool isImaginary = (std::abs(c.real()) <= epsilon) && (std::abs(c.imag()) > epsilon);
```

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -51,9 +50,21 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
+
+  if (!isImaginary) {
```
amccaskey: I wonder if this could be simplified to something like
```cpp
auto coeff = isImaginary ? std::complex<double>({0., 1.}) : std::complex<double>{1.,0.}; 
for (auto& op : pool) 
  commutators.push_back(H * (coeff * op) - (coeff * op) * H); 
```

libs/solvers/python/tests/test_operator_pools.py
```diff
@@ -69,25 +70,11 @@ def test_uccsd_operator_pool_correctness():
     pool_strings = [op.to_string(False) for op in pool]
 
     # Expected result
-    expected_pool = [
-        'YZXI', 'XZYI', 'IYZX', 'IXZY', 'XXXY', 'XXYX', 'XYYY', 'YXYY', 'XYXX',
-        'YXXX', 'YYXY', 'YYYX'
-    ]
+    expected_pool = ["XZYIYZXI", "IXZYIYZX", "XXXYXXYXXYXXXYYYYXXXYXYYYYXYYYYX"]
```
amccaskey: The op.to_string(False) is concatenating all the terms together into one string. I think it would be better to ensure that the terms are correct instead of this concatenated string.

libs/solvers/unittests/test_adapt.cpp
kvmto: I think this is coming in an another PR :)

libs/solvers/lib/adapt/adapt_simulator.cpp
kvmto: This is what I tried in the past, but it does not work. 
This works : `commutators.push_back(coeff * (H * op - op * H));`

libs/solvers/include/cudaq/solvers/operators/operator_pools/qaoa_operator_pool.h
```diff
@@ -12,16 +12,33 @@
 
 namespace cudaq::solvers {
 
-/// @brief Test
+/// @brief QAOA operator pool is a class with generates a pool
+/// of QAOA operators for use in quantum algorithms, like Adapt-QAOA.
+/// @details This class extends the operator_pool interface
+/// therefore inhers the extension_point template, allowing for
```
bmhowe23:
```suggestion
/// therefore inherits the extension_point template, allowing for
```

libs/solvers/include/cudaq/solvers/operators/operator_pools/qaoa_operator_pool.h
```diff
@@ -12,16 +12,33 @@
...
+  /// operator generation.
+  /// @return A vector of cudaq::spin_op objects representing the generated
+  /// operator pool.
   std::vector<cudaq::spin_op>
   generate(const heterogeneous_map &config) const override;
+
+  /// @brief Call to macro for defining the creator function for the qaoa_pool
+  /// extension
+  /// @details This function is used by the extension point mechanism to create
+  /// instances of the qaoa_pool class. The extension is registred with a name
```
bmhowe23:
```suggestion
  /// instances of the qaoa_pool class. The extension is registered with a name
```

libs/solvers/include/cudaq/solvers/operators/operator_pools/spin_complement_gsd.h
```diff
@@ -15,13 +15,31 @@ namespace cudaq::solvers {
 // adapted from
 // https://github.com/mayhallgroup/adapt-vqe/blob/master/src/operator_pools.py
 
-/// @brief Test
+/// @brief spin_complement_gsd operator pool is a class which generates a pool
+/// of operators with the spin complement ground state degeneracy method for use
+/// in quantum algorithms, like Adapt-VQE.
+/// @details This class extends the operator_pool interface
+/// therefore inhers extension_point template, allowing for
```
bmhowe23:
```suggestion
/// therefore inherits extension_point template, allowing for
```

libs/solvers/include/cudaq/solvers/operators/operator_pools/uccsd_operator_pool.h
```diff
@@ -12,14 +12,32 @@
 
 namespace cudaq::solvers {
 
-/// @brief Test
+/// @brief UCCSD operator pool is a class with generates a pool
+/// of UCCSD operators for use in quantum algorithms, like Adapt-VQE.
+/// @details This class extends the operator_pool interface
+/// therefore inhers the extension_point template, allowing for
```
bmhowe23:
```suggestion
/// therefore inherits the extension_point template, allowing for
```

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -101,15 +124,23 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
     // All ranks have a norm, need to reduce that across all
     if (mpi::is_initialized())
       norm = cudaq::mpi::all_reduce(norm, std::plus<double>());
 
     // All ranks have a max gradient and index
-    auto iter = std::max_element(gradients.begin(), gradients.end());
+    std::vector<double> roundedGradients(gradients.size());
+    for (size_t i = 0; i < gradients.size(); ++i) {
+      // Round to a specific precision (e.g., 10 decimal places)
+      roundedGradients[i] = std::round(gradients[i] * 1e16) / 1e16;
```
bmhowe23: Why is this here? Also, the code seems to conflict with the comments (16 digits vs 10 digits). I'm hoping this is old debugging code and no longer needed.

libs/solvers/python/tests/test_operator_pools.py
bmhowe23: I thought I heard @marwafar also ask for this, so should this conversation be reopened?

bmhowe23 (commented):
I'm still trying to get this to run in my environment (probably something wrong in my environment), but here are some initial comments.

marwafar:
Yes. This concatenation should change. You should have it like  [['XZY', 'YZX'], ['IXZY', 'IYZX'], ['XXYX', 'XXXY', 'XYYY', 'YYXY', 'XYXX', 'YXYY', 'YYYX', 'YXXX']] . You should make sure adapt_vqe() function returns final parameters, selected pool, and final energy. . For example:
Final parameters:  [0.11278279920936399]
Selected pools:  [['XXYX', 'XXXY', 'XYYY', 'YYXY', 'XYXX', 'YXYY', 'YYYX', 'YXXX']]
Final energy:  -1.1372838344885003

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -70,9 +87,15 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
   cudaq::state state = get_state(adapt_kernel, numQubits, initialState, thetas,
-                                 coefficients, pauliWords);
-  std::size_t count = 0;
+                                 coefficients, pauliWords, poolIndices);
+
+  auto step = 0;
   while (true) {
+    if (step >= 30) {
+      printf("TIMED OUT\n");
+      break;
```
bmhowe23: This `30` should be a parameter to this function, something like `maxIter`.

libs/solvers/unittests/test_adapt.cpp
```diff
@@ -12,19 +12,67 @@
 #include "cudaq.h"
 #include "nvqpp/test_kernels.h"
 #include "cudaq/solvers/adapt.h"
+#include "cudaq/solvers/operators.h"
```
bmhowe23, line 15: When I run the tests in this file w/ `CUDAQ_LOG_LEVEL=info` and just watch the stuff scroll by, there are *lots* of things that look really empty, like this:
```
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1204] Setting current circuit name to void (cudaq::state&)
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1024] Allocating 12 new qubits.
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 0 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 1 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 2 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 3 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 4 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 5 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 6 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 7 deallocation
[2025-03-13 19:50:21.962] [info] [CircuitSimulator.h:1045] Deferring qubit 8 deallocation
[2025-03-13 19:50:21.963] [info] [CircuitSimulator.h:1045] Deferring qubit 9 deallocation
[2025-03-13 19:50:21.963] [info] [CircuitSimulator.h:1045] Deferring qubit 10 deallocation
[2025-03-13 19:50:21.963] [info] [CircuitSimulator.h:1045] Deferring qubit 11 deallocation
[2025-03-13 19:50:22.053] [info] [CircuitSimulator.h:1189] Deallocated all qubits, reseting state vector.
[2025-03-13 19:50:22.105] [info] [CircuitSimulator.h:1204] Setting current circuit name to void (cudaq::state&)
[2025-03-13 19:50:22.105] [info] [CircuitSimulator.h:1024] Allocating 12 new qubits.
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 0 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 1 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 2 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 3 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 4 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 5 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 6 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 7 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 8 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 9 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 10 deallocation
[2025-03-13 19:50:22.106] [info] [CircuitSimulator.h:1045] Deferring qubit 11 deallocation
[2025-03-13 19:50:22.200] [info] [CircuitSimulator.h:1189] Deallocated all qubits, reseting state vector.
...
```
Is it possible we're running a bunch of empty circuits somehow?

libs/solvers/lib/adapt/adapt_simulator.cpp
kvmto: The rounding is necessary to avoid some repeatability issues. The 10th digits was the initial rounding, but then decided to go up again, I will put it back to 10. Thanks for spotting this.

libs/solvers/lib/adapt/adapt_simulator.cpp
kvmto: I think we can put that as a configuration options, together with the tolerance on the energy difference between iterations. If it is okay I will do that.

libs/solvers/lib/adapt/device/adapt.cpp
```diff
@@ -15,13 +16,20 @@ adapt_kernel(std::size_t numQubits,
              const cudaq::qkernel<void(cudaq::qvector<> &)> &statePrep,
              const std::vector<double> &thetas,
              const std::vector<double> &coefficients,
-             const std::vector<cudaq::pauli_word> &trotterOpList) {
+             const std::vector<cudaq::pauli_word> &trotterOpList,
+             const std::vector<std::size_t> poolIndices) {
   cudaq::qvector q(numQubits);
```
kvmto, line 21: Could this line the reason why we see the allocation and deallocation of qubits in the log? @bmhowe23

libs/solvers/python/tests/test_operator_pools.py
kvmto: I had some local changes that didn't get pushed, now they should be here as well.

libs/solvers/unittests/test_adapt.cpp
kvmto: The log since it repeats seems like the usage of the adapt_kernel, since it used iteratively during the optimization steps.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: If repeatability issues exist, this will not prevent them. One can always be *right on the edge* of a rounding decision (i.e. 0.100000000500001 and 0.100000000499999 are 2e-15 apart but will still round to different numbers when rounding to 10 digits), so it would be better to leave this out.

I am ok with non-repeatabilities as long as they go away when we run with `OMP_NUM_THREADS=1`.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: Yes, that sounds good to me.

libs/solvers/lib/adapt/device/adapt.cpp
bmhowe23: Yes, that would explain the allocation and deallocation, but I'm confused why I don't see any gates occurring. Could we be calling this kernel with an empty `trotterOpList`?

libs/solvers/lib/adapt/adapt_simulator.cpp
marwafar: Yes. This is max iteration, we can add a default value but allow user to change this. Same for the tolerance of the energy and the tolerance for the norm of gradients.

libs/solvers/unittests/test_adapt.cpp
marwafar: This is weird.

bmhowe23: It looks like they are all coming from https://github.com/NVIDIA/cudaqx/blob/9ec771bc892f4d4202f330372c4b2f8903b7c107/libs/solvers/lib/adapt/adapt_simulator.cpp#L111 (the `prepare_state` kernel).

It looks like that code is assuming that all the state vectors from all the commutators will fit in GPU memory. That seems like a dubious assumption. (This was pre-existing, not new for this PR.)

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -51,10 +62,16 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
+
+  // Check if operator has only imaginary coefficients
+  // checking the first one is enough, we assume the pool is homogeneous
+  const auto &c = pool[0].begin()->get_coefficient();
+  bool isImaginary =
+      (std::abs(c.real()) <= 1e-9) && (std::abs(c.imag()) > 1e-9);
+  auto coeff = (!isImaginary) ? std::complex<double>{0.0, 1.0}
+                              : std::complex<double>{1.0, 0.0};
+  for (auto &op : pool)
+    commutators.push_back(coeff * (H * op - op * H));
```
bmhowe23: @amccaskey @kvmto @marwafar - there are a significant number of terms in these spin ops that have coefficients = 0. Is that expected? I could be mistaken, but I suspect that means some wasted computation down below.

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -51,10 +62,16 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
-  for (int i = start; i < end; i++) {
-    auto op = pool[i];
-    commutators.emplace_back(H * op - op * H);
-  }
+
+  // Check if operator has only imaginary coefficients
+  // checking the first one is enough, we assume the pool is homogeneous
+  const auto &c = pool[0].begin()->get_coefficient();
+  bool isImaginary =
+      (std::abs(c.real()) <= 1e-9) && (std::abs(c.imag()) > 1e-9);
```
bmhowe23, line 70: Are you sure that `c` will never be 0+0j? If it is, then this `isImaginary` variable would potentially be incorrect.

libs/solvers/lib/adapt/adapt_simulator.cpp
marwafar: Also, the initial theta parameter default is zero. But, we should also let user change that if they want.

libs/solvers/lib/adapt/adapt_simulator.cpp
marwafar: The coefficients of operator pool will not be zero. They are either one or less than one.

libs/solvers/lib/adapt/adapt_simulator.cpp
marwafar: You mean the spin op in the commutator? Yes some some spin op in the commutator operator will be zero. Because of symmetry, some excitation are not allowed. Those will be zero. We should write a function after computing the commutator to remove terms with zero coefficient.

libs/solvers/lib/adapt/adapt_simulator.cpp
marwafar: Maybe also when simulation reached the maxIter, instead of only "printf("TIMED OUT\n")" You should also print or save in a file the last parameters, state , selected pools, and energy. User can later enhance the state with other classical simulation in classical post-processing Or we might also later allow user to restart simulation. Allow user to restart simulation will be a nice feature that we should add.

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -140,16 +171,22 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
     }
 
     // Convergence is reached if gradient values are small
-    if (std::sqrt(std::fabs(norm)) < tol || std::fabs(lastNorm - norm) < tol)
+    if (std::sqrt(std::fabs(norm)) < tol || std::fabs(lastNorm - norm) < tol ||
```
marwafar: You should not use the same tol value for norm and lastNorm-norm. You should provide different variable names. Also, you should allow user to change those threshold value.

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -140,16 +171,22 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
+        ediff < 1e-6)
```
marwafar: 1e-6 should be variable name like thresholdE. Add a default value but allow user to change these value if they want

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -140,16 +171,22 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
       break;
 
     // Use the operator from the pool
     auto op = pool[maxOpIdx];
+    if (!isImaginary)
+      op = std::complex<double>{0.0, 1.0} * pool[maxOpIdx];
+
     chosenOps.push_back(op);
     thetas.push_back(0.0);
```
marwafar: Default value should be zero. But allow user to change this too. Instead of "thetas.push_back(0.0);", you can do "thetas.push_back(initTheta)" and initTheta=0.0 default

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -101,15 +124,23 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
     double norm = 0.0;
     for (auto &g : gradients)
       norm += g * g;
+    norm = std::sqrt(norm);
```
marwafar, line 161: This part is not clean. You are returning in line 121 the absolute value of the gradient value, then compute norm \sqrt(\sum_i (grad_i)**2). 
I think you should do that and it will not change any result. return the gradient value and not the absolute value. Then compute norm as abs (\sqrt(\sum_i (grad_i)**2)). This is how we compute norm of a vector.

libs/solvers/lib/adapt/adapt_simulator.cpp
marwafar: Also, replace fabs(norm)) with just norm, norm is already the absolute value of the gradient vector

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -210,7 +239,12 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
     // Set the norm for the next iteration's check
     lastNorm = norm;
     state = get_state(adapt_kernel, numQubits, initialState, thetas,
-                      coefficients, pauliWords);
+                      coefficients, pauliWords, poolIndices);
+
+    if (groundEnergy < latestEnergy) {
```
marwafar: They will be always less. This if statement is redundant

libs/solvers/python/tests/test_operator_pools.py
marwafar: I do not see any change. They are still the same ["XZYIYZXI", "IXZYIYZX", "XXXYXXYXXYXXXYYYYXXXYXYYYYXYYYYX"]

kvmto: Sorry, I pushed the cpp and not the py, I will push it with the rest at this point. The test identical.

libs/solvers/unittests/test_adapt.cpp
kvmto: The component running empty circuits is "prepare_state". It just allocates and deallocates the qubits.

`__qpu__ void prepare_state(cudaq::state &state) { cudaq::qvector q{state}; } `

My guess is that in order to always run useful circuits, it could could be useful to remove the get_states, and use only adapt_kernel with the various observe

marwafar: You can always prepare the state from current ansatz using adapt_kernel. 

Is this issue only because how code is written, because of the commutator is a list, or in general when using get_state? I thought based on some old benchmark with hamiltonian simulation that using get_state could make code faster

kvmto: I will try to skip the two get_state and see if i can get a boost,
but they seem to be redundant at the moment :)
I need a moment to address everything.

libs/solvers/lib/adapt/adapt_simulator.cpp
kvmto: I need the absolute values in order to pick the max_element, somewhere I have to do the std::abs. If the order of operations does not bother I would keep it as is.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: > The coefficients of operator pool will not be zero. They are either one or less than one.

I believe `c` is being set to _the first term_ of the first spin op in the pool here. If any of the terms in the spin op can have 0-valued coefficients, then I believe this could be 0. I could be misunderstanding that, though ... looking for @kvmto to confirm that `c` is never 0 here.

kvmto: I confirm it is never 0 :)

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: > Maybe also when simulation reached the maxIter, instead of only "printf("TIMED OUT\n")" You should also print or save in a file the last parameters, state , selected pools, and energy. User can later enhance the state with other classical simulation in classical post-processing Or we might also later allow user to restart simulation. Allow user to restart simulation will be a nice feature that we should add.

These sound like nice features to add, but probably not part of this PR. For this PR, I think returning/printing an error if it runs longer than a new `maxIter` parameter should be sufficient.

libs/solvers/unittests/test_adapt.cpp
kvmto: @bmhowe23 any idea why it is not possible to do :
```
      resultHandles.emplace_back(
          observe_async(qpuCounter++, adapt_kernel, commutators[i], numQubits, initialState, thetas, coefficients, pauliWords, poolIndices));
```

bmhowe23: I looked into this a bit more. It looks like it is indeed applying the `commutators[i]` spin op at the end of each circuit (via https://github.com/NVIDIA/cuda-quantum/blob/50066dd99025c4eae2af150c327664328b8249cc/runtime/nvqir/custatevec/CuStateVecCircuitSimulator.cpp#L548) ... there are no logs in that function, so reading the logs may have led us down a wild goose chase here.

That being said - we may improve the efficiency by removing the terms that have 0-valued coefficients inside the commutators.

kvmto: Just done.:)
I am trying to avoid the double calculation of the state, do you happen to have an answer to the observe_async?

bmhowe23: > @bmhowe23 any idea why it is not possible to do :
> 
> ```
>       resultHandles.emplace_back(
>           observe_async(qpuCounter++, adapt_kernel, commutators[i], numQubits, initialState, thetas, coefficients, pauliWords, poolIndices));
> ```

Not off the top of my head ... what happens when you try that?

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: Kevin wrote https://github.com/NVIDIA/cuda-quantum/issues/2737, so I will mark this conversation as resolved.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: I see this was fixed in 6df87309542325df24048b3887bf9cb7332ac605, so I'm marking this as resolved.

libs/solvers/unittests/test_adapt.cpp
bmhowe23: As far as I can tell, everything in this thread was resolved, so I am marking this as resolved.  If there is anything else, let's open a new thread.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: I see this was updated in 6df87309542325df24048b3887bf9cb7332ac605, so marking resolved.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: I see this was updated in 6df87309542325df24048b3887bf9cb7332ac605, so I'm marking as resolved.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: I see this was changed in 6df87309542325df24048b3887bf9cb7332ac605, so I'm marking as resolved.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: I see this was updated in 6df87309542325df24048b3887bf9cb7332ac605, so I'm marking as resolved.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: I see this was updated in 6df87309542325df24048b3887bf9cb7332ac605 so I am marking this thread as resolved.

libs/solvers/lib/adapt/device/adapt.cpp
bmhowe23: Marking as resolved.

libs/solvers/lib/adapt/adapt_simulator.cpp
bmhowe23: I see that a new `cleanCommutator` was added in 6df87309542325df24048b3887bf9cb7332ac605, so marking as resolved.

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -70,9 +114,18 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
 
   // Start of with the initial |psi_n>
   cudaq::state state = get_state(adapt_kernel, numQubits, initialState, thetas,
-                                 coefficients, pauliWords);
-  std::size_t count = 0;
+                                 coefficients, pauliWords, poolIndices);
+
+  int step = 0;
   while (true) {
+    printf("Step %d\n", step);
```
bmhowe23:
```suggestion
    if (options.get("verbose", false))
      printf("Step %d\n", step);
```

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -30,8 +35,17 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
+  std::vector<std::size_t> poolIndices;
   std::vector<cudaq::spin_op> chosenOps;
+  double latestEnergy = std::numeric_limits<double>::max();
+  double ediff = std::numeric_limits<double>::max();
+
+  int maxIter = options.get<int>("maxIter", 30);
   auto tol = options.get<double>("grad_norm_tolerance", 1e-5);
+  auto tolNormDiff = options.get<double>("grad_norm_diff_tolerance", 1e-5);
+  auto thresholdE = options.get<double>("threshold_energy", 1e-6);
+  auto initTheta = options.get<double>("initial_theta", 0.0);
```
bmhowe23: Thanks for adding these needed parameters. Could you please add them to Doxygen (and Python pydocs) for `adapt_vqe`?

I believe that is:
* https://github.com/NVIDIA/cudaqx/blob/529ec6f57740619506a09a11f2f781b0245580df/libs/solvers/python/bindings/solvers/py_solvers.cpp#L965
* https://github.com/NVIDIA/cudaqx/blob/529ec6f57740619506a09a11f2f781b0245580df/libs/solvers/include/cudaq/solvers/adapt.h#L94
* https://github.com/NVIDIA/cudaqx/blob/529ec6f57740619506a09a11f2f781b0245580df/libs/solvers/include/cudaq/solvers/adapt.h#L112
* https://github.com/NVIDIA/cudaqx/blob/529ec6f57740619506a09a11f2f781b0245580df/libs/solvers/include/cudaq/solvers/adapt.h#L131

Note that the py_solvers.cpp one will _also need code changes_ to make Python honor these new options as well (e.g. https://github.com/NVIDIA/cudaqx/blob/529ec6f57740619506a09a11f2f781b0245580df/libs/solvers/python/bindings/solvers/py_solvers.cpp#L933C9-L933C19)

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -6,6 +6,12 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 
+#include <cmath>
+#include <cstddef>
+#include <cstdio>
+#include <fmt/core.h>
```
bmhowe23: Did these extra 4 #include's end up being needed? By my local testing, they aren't needed, and if not, we should remove them.

bmhowe23 (commented):
Thanks, Kevin ... I think we're getting pretty close. I have a few minor comments below.

My main remaining question is whether or not this is as performant as Marwa's code. If not, we may merge it anyway (just to fix the known issue), but we would have to allocate additional time to come back and analyze any remaining performance issues.

kvmto:
# C++ vs Python ADAPT-VQE: Tables, Code, and Settings for Reproducibility

## Performance Results for LiH, the minimal molecule with large difference in execution time

### C++ Implementation Performance

| Run | Real Time | User Time | System Time |
|-----|-----------|-----------|-------------|
| 1   | 5m36.918s | 3m9.283s  | 1m52.294s   |
| 2   | 5m38.389s | 3m7.260s  | 1m53.344s   |
| 3   | 5m35.464s | 3m8.628s  | 1m52.660s   |
| 4   | 5m29.788s | 3m7.512s  | 1m50.279s   |
| 5   | 5m52.388s | 3m7.170s  | 1m57.231s   |
| 6   | 6m14.200s | 3m9.194s  | 1m59.558s   |
| 7   | 6m17.362s | 3m10.504s | 2m1.507s    |
| 8   | 6m9.004s  | 3m11.837s | 2m2.394s    |
| 9   | 6m8.432s  | 3m10.065s | 1m58.100s   |
| 10  | 5m55.503s | 3m8.538s  | 1m55.011s   |

### Python Implementation Performance

| Run | Real Time | User Time | System Time |
|-----|-----------|-----------|-------------|
| 1   | 1m5.611s  | 0m42.805s | 0m16.678s   |
| 2   | 1m5.469s  | 0m42.706s | 0m16.800s   |
| 3   | 1m5.237s  | 0m42.439s | 0m16.634s   |
| 4   | 1m5.637s  | 0m42.932s | 0m16.414s   |
| 5   | 1m5.291s  | 0m42.255s | 0m16.668s   |
| 6   | 1m5.569s  | 0m43.090s | 0m16.262s   |
| 7   | 1m5.096s  | 0m42.463s | 0m16.651s   |
| 8   | 1m5.169s  | 0m42.874s | 0m16.622s   |
| 9   | 1m5.284s  | 0m42.701s | 0m16.400s   |
| 10  | 1m5.769s  | 0m42.930s | 0m16.518s   |

## Code Samples

### C++ Implementation

```cpp
#include <cmath>
#include <gtest/gtest.h>

#include "cudaq.h"
#include "cudaq/solvers/adapt.h"
#include "cudaq/solvers/operators.h"

__qpu__ void statePrep4Electrons(cudaq::qvector<> &q) {
  for (std::size_t i = 0; i < 4; i++)
    x(q[i]);
}

int main(){
  cudaq::solvers::molecular_geometry geometryLiH = {
    {"Li", {0.3925, 0., 0.}}, {"H", {-1.1774, 0., 0.0}}};

  auto lih = cudaq::solvers::create_molecule(
    geometryLiH, "sto-3g", 0, 0,
    {.casci = true, .ccsd = true, .verbose = true});


  auto hamli = lih.hamiltonian;

  auto pool = cudaq::solvers::operator_pool::get("uccsd");
  auto opt = cudaq::optim::optimizer::get("lbfgs");
  heterogeneous_map config;
  config.insert("num-qubits", hamli.num_qubits());
  config.insert("num-electrons", 4);
  auto poolList = pool->generate(config);
  auto [energy, thetas, ops] = cudaq::solvers::adapt_vqe(
    statePrep4Electrons, hamli, poolList, *opt, "central_difference",
    {{"grad_norm_tolerance", 1e-3},
    {"verbose", false},
    {"maxIter", 30},
    {"grad_norm_diff_tolerance", 1e-5},
    {"threshold_energy", 5e-6},
    {"initial_theta", 0.0},
    {"optimizer", "L-BFGS-B"},
    {"jac", "3-point"},
    {"tol", 1e-7},
    {"cutoff", 1e-15}});
}
```

## Environment and Build Settings

### CUDA Quantum Version
```
nvq++ Version proto-0.8.0 (https://github.com/NVIDIA/cuda-quantum d2f3a958a211fe2f096fc61378d2ca5b38a58703)
```

### C++ Build Configuration
```
cmake .. -G Ninja \
-DCUDAQX_ENABLE_LIBS="all" \
-DCUDAQX_INCLUDE_TESTS=ON \
-DCUDAQX_BINDINGS_PYTHON=ON \
-DCUDAQX_INSTALL_PYTHON=ON \
-DCUDAQ_DIR=$HOME/.cudaq/lib/cmake/cudaq \
-DCMAKE_CXX_FLAGS="-Wno-aRributes" \
-DCMAKE_BUILD_TYPE=Release \
-DCMAKE_INSTALL_PREFIX=$HOME/.cudaq

ninja install

nvq++ --enable-mlir -lcudaq-solvers main.cpp
```

### Python Environment
```
which python3 
/usr/bin/python3

echo $PYTHONPATH
/home/.cudaq:/home/workspaces/cudaqx/build/python

echo $OMP_NUM_THREADS
1
```

### Running C++ Implementation
```bash
for i in {1..10}; do 
  echo "Run $i:"; 
  time CUDA_VISIBLE_DEVICES=1 ./a.out > /dev/null 2>&1; 
  echo "--------------------"; 
done
```

### Running Python Implementation
```bash
for i in {1..10}; do 
  echo "Run $i:"; 
  time CUDA_VISIBLE_DEVICES=1 python3 adapt-vqe.py --obi-file LiH_one_body.dat --tbi-file LiH_two_body.dat --json-file LiH_metadata.json --verbose > /dev/null 2>&1; 
  echo "--------------------"; 
done
```

bmhowe23:
Thanks for the detailed timing numbers, Kevin. Could you please try the changes in https://gist.github.com/bmhowe23/1e342775adaa3216918d7fd6b92df29b and see if that resolves any of the performance discrepancy? It changes 3 things.
1. When numQpus is 1, it uses observe instead of observe_async. This is probably the least important of the 3 changes.
2. In each step, it initializes the theta parameters to the `optParams` from the prior time through the loop. The Python code (Marwa's code) was doing this but our C++ code was not.
3. It updates our "LBFGS" C++ optimizer to look for "tol" instead of "function_tolerance" to set the function tolerance for the optimizer. This was an important setting that was being ignored by the optimizer.

kvmto:
# New LiH Molecule Performance Comparison

New options introduced in order to improve the performance of adapt vqe.
Observe = When the numQPUs is equal to 1, all the observe async are just observe.
Tolerance = introduction of a tolerance threshold for the optimization of the thetas.
Memory of thetas  = new option for reusing the previously optimized thetas, which is a problem specific enhancement.  

## Configuration Summary

| Configuration | Observe | Tolerance 1e-7 | Memory of thetas | Avg. Real Time | Avg. User Time | Avg. Sys Time |
|---------------|:-------:|:--------------:|:----------------:|:--------------:|:--------------:|:-------------:|
| Config 1      | ✓       | ✓              | ✗                | 1m20.780s      | 0m40.003s      | 0m23.995s     |
| Config 2      | ✓       | ✓              | ✓                | 0m50.153s      | 0m25.734s      | 0m14.767s     |
| Config 3      | ✗       | ✓              | ✓                | 0m49.537s      | 0m31.423s      | 0m14.895s     |
| Config 4      | ✓       | ✗              | ✓                | 1m55.787s      | 0m56.671s      | 0m35.173s     |

## Performance Improvement Analysis

| Comparison | Time Reduction | Percentage Improvement |
|------------|----------------|------------------------|
| Config 1 → Config 2 | 30.627s | 37.9% |
| Config 3 → Config 2 | -0.616s | -1.2% |
| Config 4 → Config 2 | 65.634s | 56.6% |
| Config 4 → Config 3 | 66.250s | 57.2% |
| Config 4 → Config 1 | 35.007s | 30.2% |
| Config 1 → Config 3 | 31.243s | 38.7% |

libs/solvers/include/cudaq/solvers/adapt.h
```diff
@@ -84,6 +84,8 @@ class adapt_impl : public extension_point<adapt_impl> {
   ///  - "verbose" (bool): Enable detailed output logging [default: false]
   ///  - "shots" (int): Number of measurement shots (-1 for exact simulation)
   ///  [default: -1]
+  ///  - "opt_mode" (string): Optimization mode [default: "global"]
+  ///  - "tol" (double): Tolerance for optimization [default: 1e-6]
```
bmhowe23: This default value doesn't look right to me. The LBFGS optimizer has a default of 1e-12, doesn't it? (I didn't look at other optimizers.)

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -178,39 +238,43 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
-        // data.emplace_back(x, res, observe_execution_type::function);
-        // for (auto datum : gradient->data)
-        //   data.push_back(datum);
-
         return res.expectation();
       };
     }
 
-    // FIXME fix the const_cast
+    if (options.contains("opt_mode")) {
```
bmhowe23: I think these opt_mode references need to be changed to "dynamic_start", don't they?

libs/solvers/include/cudaq/solvers/adapt.h
```diff
@@ -71,6 +71,23 @@ class adapt_impl : public extension_point<adapt_impl> {
   /// @param optimizer Optimization algorithm
   /// @param gradient Gradient calculation method
   /// @param options Additional options for the algorithm
+  /// @par Supported Keys:
```
bmhowe23: This `@par` option is rendering the docs a bit strangely for me. (Comment applies to all places where `@par` was used. See below for my screenshot and suggested change. I'm open to other suggestions if you have any.

image

I think it would be better to do something like this so that the options are correctly listed with the `options` parameter.
```cpp
/// @param options Additional options for the algorithm. Supported keys:
///  - "maxIter" (int): Maximum number of iterations [default: 30]
///  - "grad_norm_tolerance" (double): Convergence tolerance for gradient norm [default: 1e-5]
///  - "grad_norm_diff_tolerance" (double): Tolerance for difference between gradient norms [default: 1e-5]
/// ...
```

libs/solvers/include/cudaq/solvers/adapt.h
```diff
@@ -71,6 +71,23 @@ class adapt_impl : public extension_point<adapt_impl> {
...
+  ///  - "maxIter" (int): Maximum number of iterations [default: 30]
```
bmhowe23: We should use `max_iter` (snake case) for consistency since all the other options are snake case.

libs/solvers/unittests/test_adapt.cpp
```diff
@@ -12,37 +12,483 @@
...
+
+class SolversTester : public ::testing::Test {
+protected:
+  static cudaq::spin_op h;
+  static cudaq::spin_op hamli;
+  static cudaq::spin_op hamhh;
+  static cudaq::spin_op hambeh2;
+
+  static void SetUpTestSuite() {
+    std::vector<double> h2_data{
```
bmhowe23, line 36: Nothing to change yet, but just FYI, this way of specifying spin ops will be going away with the new CUDA-Q spin op updates.

libs/solvers/lib/adapt/adapt_simulator.cpp
```diff
@@ -51,9 +62,39 @@ simulator::run(const cudaq::qkernel<void(cudaq::qvector<> &)> &initialState,
...
+  bool isImaginary =
+      (std::abs(c.real()) <= 1e-9) && (std::abs(c.imag()) > 1e-9);
+  auto coeff = (!isImaginary) ? std::complex<double>{0.0, 1.0}
+                              : std::complex<double>{1.0, 0.0};
+
+  auto cleanCommutator =
+      [](cudaq::spin_op commutator) -> std::tuple<cudaq::spin_op, std::size_t> {
+    cudaq::spin_op cleaned;
+    std::size_t numTerms = 0;
+    commutator.for_each_term([&](const auto &term) {
```
bmhowe23, line 78: Nothing to change yet, but just FYI, `for_each_term` will be going away with Bettina's spin op updates. On the bright side, a new `trim` help function will be created that cleans up zero-coefficient terms, so we won't need this `cleanCommutator` lambda then.

bmhowe23 (commented):
Thanks, Kevin. I really like the thorough tests and docs.

bmhowe23 (approved):
Thanks, Kevin!

PR #68, opened 2025-02-05
Add python binding for decoder apis

melody-ren:
Expose the following two functions to python:

`virtual std::future<decoder_result> decode_async(const std::vector<float_t> &syndrome);`
`virtual std::vector<decoder_result> decode_multi(const std::vector<std::vector<float_t>> &syndrome);`


libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -133,6 +133,21 @@ void bindDecoder(py::module &mod) {
           },
           "Decode the given syndrome to determine the error correction",
           py::arg("syndrome"))
+      .def(
+          "decode_async",
+          [](decoder &decoder, const std::vector<float_t> &syndrome) {
+            py::gil_scoped_release release;
```
melody-ren, line 155: This will be a blocking call on the python side. If we want truly non-blocking async interface on the python side as well, we can probably use python's asyncio. 

Let me know what you prefer

bmhowe23: Thanks, Melody. Would it be possible to do it like https://github.com/NVIDIA/cuda-quantum/blob/main/python/runtime/cudaq/algorithms/py_sample_async.cpp? That implementation is a non-blocking one, so it seems like we should be able to mimic that behavior.

melody-ren: I used a similar approach to implement it in a non-blocking manner. Originally, I intended to use `cudaq::async_result`, but then realized it was hardcoded for `sample_result`. That's why I've introduced a simple decoder wrapper `async_decoder_result`

libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -43,6 +43,29 @@ struct decoder_result {
   }
 };
 
+/// @brief Return type for asynchronous decoding results
+class async_decoder_result {
```
bmhowe23, line 47: It would be good to add a move constructor and a move assignment constructor so that users can transfer ownership of these `async_decoder_result` objects.

For example, see this (untested) code:

```cpp
  // Move constructor
  async_decoder_result(async_decoder_result &&other) noexcept
      : fut(std::move(other.fut)) {}

  // Move assignment
  async_decoder_result &operator=(async_decoder_result &&other) noexcept {
    if (this != &other)
      fut = std::move(other.fut);
    return *this;
  }
```

bmhowe23 (approved):
Thanks, Melody ... just one small comment.

PR #82, opened 2025-02-10
Refactoring and debugging of Jordan Wigner transform (Issue #67)

### LINKED ISSUE #67: Incorrect coefficients in Jordan-Wigner Transform
### Description

There is a discrepancy in the Pauli string coefficients between OpenFermion and CUDA-QX.

- The Pauli strings themselves match in both OpenFermion and CUDA-QX for all molecules.
- The coefficients match for H_2 but not for other molecules. For example, mismatches are observed for H_4.
- Some coefficients differ only by sign, suggesting a possible parity flip or phase shift issue.

### Steps to Reproduce
1. **Setup & Imports**  
   - Import necessary libraries: CUDA-Q, CUDA-QX, SciPy, PySCF, and NumPy.  
   - Define the molecular structure for **H_4** using Cartesian coordinates.  

2. **Compute Molecular Energies**  
   - **FCI Energy (PySCF)**: Compute exact energy using Full Configuration Interaction (FCI).  
   - **OpenFermion Energy (CUDA-Q)**: Generate Hamiltonian, compute lowest eigenvalue.  
   - **CUDA-QX Energy**: Generate Hamiltonian using CASCI, compute lowest eigenvalue.  

3. **Extract & Compare Pauli Terms**  
   - Convert Hamiltonians into Pauli string representations with coefficients.  
   - Compare OpenFermion and CUDA-QX coefficients.  
   - Print mismatches where coefficients differ beyond a tolerance (`1e-15`).  

**Purpose:** Identify discrepancies between OpenFermion and CUDA-QX Hamiltonians, specifically in Pauli string coefficients.  

### Expected Behavior
Pauli string coefficients should match between OpenFermion and CUDA-QX for all molecules.

### Actual Behavior
For molecules other than H_2, some coefficients do not match. Even when using PySCF-verified integrals, CUDA-QX produces incorrect eigenvalue energies, suggesting the issue lies in the Jordan-Wigner implementation in CUDA-QX.

kvmto:
This PR fixes #67.
## Problem
The Jordan Wigner transform did not work correctly, in fact it generated hamiltonians with wrong coefficients or terms.

## Changes
Changes so far have been applied to:

Rewritten from scratch jordan_wigner.cpp.
Added tests.
~~The original code from which the cpp version is created is https://github.com/quantumlib/OpenFermion/blob/v1.6.1/src/openfermion/transforms/opconversions/jordan_wigner.py#L128~~

## Additional notes

~~The PR is not done yet! It requires more tests and a second look to the implementation.~~
~~The code has not been linted since much more changes are needed.~~

## Findings
~~The implementation for molecule H4 now has correct energy values but higher number of terms compared to the python version.~~
~~Molecules like LiH do not achieve the correct energy values.~~

(Edited by @bmhowe23 given the latest updates from everyone)


libs/solvers/lib/operators/molecule/fermion_compilers/jordan_wigner.cpp
```diff
@@ -16,368 +27,208 @@ using namespace cudaqx;
 
 namespace cudaq::solvers {
 
-cudaq::spin_op one_body(std::size_t p, std::size_t q,
-                        std::complex<double> coeff) {
-  if (p == q)
-    return 0.5 * coeff * cudaq::spin::i(p) - 0.5 * coeff * cudaq::spin::z(p);
+cudaq::spin_op one_body(std::size_t p, std::size_t q, std::complex<double> coeff) {
```
amccaskey: I think a better approach may be found in just falling back on `spin_op` and its arithmetic overloads: 
```cpp
auto is_complex_zero = [tol](const std::complex<double> &z) {
    return std::abs(z.real()) < tol && std::abs(z.imag()) < tol;
  };

  auto adag = [](std::size_t numQubits, std::size_t j) {
    cudaq::spin_op zprod(numQubits);
    for (std::size_t k = 0; k < j; k++)
      zprod *= cudaq::spin::z(k);
    return 0.5 * zprod *
           (cudaq::spin::x(j) - std::complex<double>{0, 1} * cudaq::spin::y(j));
  };

  auto a = [](std::size_t numQubits, std::size_t j) {
    cudaq::spin_op zprod(numQubits);
    for (std::size_t k = 0; k < j; k++)
      zprod *= cudaq::spin::z(k);
    return 0.5 * zprod *
           (cudaq::spin::x(j) + std::complex<double>{0, 1} * cudaq::spin::y(j));
  };

  std::size_t nqubit = hpq.shape()[0];

  for (std::size_t i = 0; i < hpq.shape()[0]; i++)
    for (std::size_t j = 0; j < hpq.shape()[1]; j++)
      if (!is_complex_zero(hpq.at({i, j})))
        spin_hamiltonian += hpq.at({i, j}) * adag(nqubit, i) * a(nqubit, j);

  for (std::size_t i = 0; i < hpqrs.shape()[0]; i++)
    for (std::size_t j = 0; j < hpqrs.shape()[1]; j++)
      for (std::size_t k = 0; k < hpqrs.shape()[0]; k++)
        for (std::size_t l = 0; l < hpqrs.shape()[1]; l++)
          if (!is_complex_zero(hpqrs.at({i, j, k, l})))
            spin_hamiltonian += hpqrs.at({i, j, k, l}) * adag(nqubit, i) *
                                adag(nqubit, j) * a(nqubit, k) * a(nqubit, l);

  // // Remove terms with 0.0 coefficient
  std::vector<cudaq::spin_op> nonZeros;
  for (auto term : spin_hamiltonian) {
    auto coeff = term.get_coefficient();
    if (std::fabs(coeff) > tol)
      nonZeros.push_back(term);
  }
  auto op = nonZeros[0];
  for (std::size_t i = 1; i < nonZeros.size(); i++)
    op += nonZeros[i];

  return op;
```
with this I get the correct FCI energies for Yuri's example cases.

bmhowe23: Is this what you had in mind, Alex? https://github.com/NVIDIA/cudaqx/pull/82/commits/6ae55b2f5f89b7c88fde94a0acb48218d2263712

Hopefully so. I went ahead and pushed it to this branch under the hope that's what you were going for.

amccaskey: Yep that looks good

libs/solvers/python/tests/test_jordan_wigner.py
```diff
@@ -0,0 +1,69 @@
...
+    # Compute energy using CUDA-Q/OpenFermion
+    of_hamiltonian, data = cudaq.chemistry.create_molecular_hamiltonian(
+        xyz, 'sto-3g', 1, 0)
+    of_energy = np.min(np.linalg.eigvals(of_hamiltonian.to_matrix()))
+    print(f'OpenFermion energy: {of_energy.real}')
+
+    # Compute energy using CUDA-QX
+    molecule = solvers.create_molecule(xyz, 'sto-3g', 0, 0, casci=True)
+    # op = solvers.jordan_wigner(molecule.hpg, molecule.hpqrs, )
+    cudaq1_eig = np.min(np.linalg.eigvals(molecule.hamiltonian.to_matrix()))
```
bmhowe23: @wsttiger @amccaskey I assume this test needs to be updated? It looks like the code changes to jordan_wigner.cpp aren't being tested until the `op = solvers.jordan_wigner` line is uncommented?

bmhowe23: I just pushed a draft candidate test update, but I would be happy to revert/change it if others have any other suggestions.

Also - I suspect we should move this new test into the existing `libs/solvers/python/tests/test_molecule.py` test file. Any objections?

libs/solvers/python/tests/test_jordan_wigner.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Compute energy using CUDA-QX
+    molecule = solvers.create_molecule(xyz, 'sto-3g', 0, 0, casci=True)
+    op = solvers.jordan_wigner(molecule.hpq,
+                               molecule.hpqrs,
+                               core_energy=molecule.energies['nuclear_energy'])
+    assert op == molecule.hamiltonian
+    assert of_hamiltonian == molecule.hamiltonian
+
+    # FIXME - why do we need to call eigvals again if we can just assert the equality checks above?
```
bmhowe23, line 84: > FIXME - why do we need to call eigvals again if we can just assert the equality checks above?
Ahh, this is probably why:
https://github.com/bmhowe23/cuda-quantum/blob/20e5869901c69c0bbbac074cc468e1b398ce136e/runtime/cudaq/spin_op.h#L276-L278

bmhowe23 (approved):
Thanks, all!

bmhowe23:
Thanks to @Kenny-Heitritter and @james-qbraid for originally finding the parent issue for this PR!

PR #61, opened 2025-01-26
First shot memory

justinlietz:
Change the API of `qec::sample_memory_circuit` to return nRounds of data as opposed to nRounds - 1 of data. Along with returning nRounds - 1 of xor'd syndrome data, the initial unxor'd shot is also returned first.

PR #85, opened 2025-02-13
Follow-up to #61 - apply first-round data to all shots

bmhowe23:
This fixes what I think is a bug with #61. @justinlietz - please verify.

The problem was that the first round data in `syndromeTensor` was only being copied for `shot == 0` and was left uninitialized/uncopied for all other shots.


justinlietz (approved):
Good catch. This change should apply to each shot. Both changes look good to me.

PR #86, opened 2025-02-14
Follow-up to #82 - test updates

bmhowe23:


libs/solvers/python/tests/test_jordan_wigner.py
```diff
@@ -57,7 +61,8 @@ def jw_molecule_compare_hamiltonians_test(xyz):
         assert (k in of_hamiltonian_dict.keys())
 
     for k in of_hamiltonian_dict.keys():
-        np.isclose(of_hamiltonian_dict[k], cqx_op_dict[k], 1e-12)
+        assert np.isclose(abs(of_hamiltonian_dict[k]), abs(cqx_op_dict[k]),
```
kvmto: I would compare directly the coefficients, since np.isclose compares real and imaginary part separately
```suggestion
        assert np.isclose(of_hamiltonian_dict[k], cqx_op_dict[k],
```

kvmto (approved):
Looks good overall, just one small suggestion

libs/solvers/python/tests/test_jordan_wigner.py
bmhowe23: The PySCF non-repeatability is preventing us from doing that exact comparison (see https://github.com/NVIDIA/cudaqx/actions/runs/13319833635/job/37205059801#step:13:108), but I can do `assert isclose(abs(real(val1)),abs(real(val2)))` and the same thing for the imaginary components?

kvmto: I see, sorry. I thought setting to one thread would avoid the sign flipping. Then let's do that the comparison on the single components :)

bmhowe23: Setting to one thread avoided the _run-to-run_ sign flipping, but unfortunately the sign was still flipped relative to the truth data we were comparing against.

bmhowe23: Updated in 7af7504

PR #87, opened 2025-02-14
Fix undefined behavior in `cudaq::qec::to_parity_matrix`

bmhowe23:
Using `const_cast` to remove the const-ness of a parameter results in undefined behavior if the incoming parameter was truly a `const` value. Given that the `to_parity_matrix` function was called by `const` class member functions with member variables as parameters, this was indeed unsafe behavior. While I have not seen any direct badness result from this, I think it's best to pre-emptively clean this up.

### DIFF
--- a/libs/qec/lib/stabilizer_utils.cpp
+++ b/libs/qec/lib/stabilizer_utils.cpp
@@ -8,29 +8,34 @@
 #include "cudaq/qec/stabilizer_utils.h"
 
 namespace cudaq::qec {
+// Sort the stabilizers, z first, then x
+static bool spinOpComparator(const cudaq::spin_op &a, const cudaq::spin_op &b) {
+  auto astr = a.to_string(false);
+  auto bstr = b.to_string(false);
+  auto zIdxA = astr.find_first_of("Z");
+  auto zIdxB = bstr.find_first_of("Z");
+  if (zIdxA == std::string::npos) {
+    if (zIdxB != std::string::npos)
+      return false;
+
+    // No Z in either, must both contain a X
+    auto xIdxA = astr.find_first_of("X");
+    return xIdxA < bstr.find_first_of("X");
+  }
+
+  // First contains a Z
+  if (zIdxB == std::string::npos)
+    return true;
+
+  return zIdxA < zIdxB;
+}
+
+static bool isSorted(const std::vector<cudaq::spin_op> &ops) {
+  return std::is_sorted(ops.begin(), ops.end(), spinOpComparator);
+}
+
 void sortStabilizerOps(std::vector<cudaq::spin_op> &ops) {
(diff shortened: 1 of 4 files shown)


libs/qec/lib/stabilizer_utils.cpp
```diff
@@ -42,14 +47,25 @@ to_parity_matrix(const std::vector<cudaq::spin_op> &stabilizers,
   if (stabilizers.empty())
     return cudaqx::tensor<uint8_t>();
 
-  sortStabilizerOps(const_cast<std::vector<cudaq::spin_op> &>(stabilizers));
```
bmhowe23, line 45: This was the unsafe line that motivated this PR.

melody-ren: Thanks for catching this

libs/qec/lib/stabilizer_utils.cpp
```diff
@@ -8,29 +8,34 @@
...
+  }
+
+  // First contains a Z
+  if (zIdxB == std::string::npos)
+    return true;
+
+  return zIdxA < zIdxB;
+}
+
+static bool isSorted(const std::vector<cudaq::spin_op> &ops) {
```
melody-ren: Nit: might want to name this `isStabilizerSorted`. Very optional though

libs/qec/lib/stabilizer_utils.cpp
```diff
@@ -8,29 +8,34 @@
 #include "cudaq/qec/stabilizer_utils.h"
 
 namespace cudaq::qec {
+// Sort the stabilizers, z first, then x
```
melody-ren:
```suggestion
// Sort the stabilizers by the occurrence of 'Z' first, then by 'X'. 
// An earlier occurrence is considered "less". 
// Example: a = "ZZX", b = "ZXZ" → a < b 
```

melody-ren (commented):
Looks good. Minor comments

libs/qec/lib/stabilizer_utils.cpp
bmhowe23: In the example given (a = "ZZX", b = "ZXZ"), I actually think it would be the other way around. `zIdxA = zIdxB = 0`, so `zIdxA < zIdxB` would be `false`. So I would like to change the example to b = "XXZ".

libs/qec/lib/stabilizer_utils.cpp
bmhowe23: Updated in 4285f08

libs/qec/lib/stabilizer_utils.cpp
bmhowe23: Updated in 4285f08

melody-ren (approved):

PR #89, opened 2025-02-19
Use explicit measurements for memory circuit experiments

bmhowe23:
This PR makes use of a new CUDA-Q feature that speeds up our QEC workloads when using Stim as the target simulator. These changes allow all the shots to be generated in a single execution of the quantum kernel.

Ref: https://github.com/NVIDIA/cuda-quantum/pull/2567

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "76ffe23f402c230dc4b25de198142098124392c5"
+    "ref": "1e8c97d71baee9fcff57813063492e7ae1f6b559"
   }
 }
--- a/libs/core/include/cuda-qx/core/tensor.h
+++ b/libs/core/include/cuda-qx/core/tensor.h
@@ -1,5 +1,5 @@
 /****************************************************************-*- C++ -*-****
- * Copyright (c) 2024 NVIDIA Corporation & Affiliates.                         *
+ * Copyright (c) 2024 - 2025 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *
@@ -43,8 +43,9 @@ public:
   /// @brief Construct an empty tensor
   tensor()
       : pimpl(std::shared_ptr<details::tensor_impl<Scalar>>(
-            details::tensor_impl<Scalar>::get(
-                std::string("xtensor") + std::string(ScalarAsString), {})
+            details::tensor_impl<Scalar>::get(std::string("xtensor") +
+                                                  std::string(ScalarAsString),
+                                              std::vector<std::size_t>())
                 .release())) {}
 
   /// @brief Construct a tensor with the given shape
@@ -63,6 +64,16 @@ public:
             details::tensor_impl<Scalar>::get(std::string("xtensor") +
                                                   std::string(ScalarAsString),
                                               data, shape)
+                .release())) {}
+
+  /// @brief Construct a tensor with the given bitstrings
+  /// @param data Bitstrings from which to construct tensor
+  template <typename T, typename = std::enable_if_t<
+                            std::is_convertible<T, std::string>::value>>
+  tensor(const std::vector<T> &data)
+      : pimpl(std::shared_ptr<details::tensor_impl<Scalar>>(
+            details::tensor_impl<Scalar>::get(
+                std::string("xtensor") + std::string(ScalarAsString), data)
                 .release())) {}
 
   /// @brief Get the rank of the tensor
@@ -228,6 +239,10 @@ public:
   const scalar_type *data() const { return pimpl->data(); }
 
   void dump() const { pimpl->dump(); }
+
+  /// @brief Dump tensor as bits, where non-zero elements are shown as '1' and
+  /// zero-elements are shown as '.'.
+  void dump_bits() const { pimpl->dump_bits(); }
 };
 
 } // namespace cudaqx
--- a/libs/core/include/cuda-qx/core/tensor_impl.h
+++ b/libs/core/include/cuda-qx/core/tensor_impl.h
@@ -1,5 +1,5 @@
 /****************************************************************-*- C++ -*-****
- * Copyright (c) 2024 NVIDIA Corporation & Affiliates.                         *
+ * Copyright (c) 2024 - 2025 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *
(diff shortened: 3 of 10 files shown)


libs/core/include/cuda-qx/core/tensor.h
```diff
@@ -65,6 +66,16 @@ class tensor {
                                               data, shape)
                 .release())) {}
 
+  /// @brief Construct a tensor with the given bitstrings
```
bmhowe23, line 69: @amccaskey - There are many other changes we could consider pulling into `tensor` (e.g. various slice capabilities), but I don't think they are strictly required for this change, so I've kept them separate for now. I guess the main question to you is: does this change conflict with the spec in any way?

.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "76ffe23f402c230dc4b25de198142098124392c5"
+    "ref": "1e8c97d71baee9fcff57813063492e7ae1f6b559"
```
bmhowe23: Temp change to be reverted once https://github.com/NVIDIA/cuda-quantum/pull/2567 is merged.

libs/core/include/cuda-qx/core/tensor.h
amccaskey: Technically yes, it does conflict. I say go with this for now, but it will have to change when the `cudaq::basic_tensor` is upstreamed into cuda-q. There will be a `cudaq::matrix` type based on the new spec. We will need to update this to that. We can also update the spec to add this construction from strings to uint8.

libs/core/include/cuda-qx/core/tensor_impl.h
```diff
@@ -49,6 +49,38 @@ class tensor_impl : public extension_point<tensor_impl<Scalar>, const Scalar *,
...
   }
 
+  /// @brief Create a tensor implementation with the given set of bit strings
+  /// @param name The name of the tensor implementation
+  /// @param data Pointer to the tensor data as a vector of bitstrings
+  /// @return A unique pointer to the created tensor implementation
+  /// @throws std::runtime_error if the requested tensor implementation is
+  /// invalid
+  static std::unique_ptr<tensor_impl<Scalar>>
+  get(const std::string &name, const std::vector<std::string> &data) {
```
justinlietz, line 59: Should we limit this API to integral types, or even uint8_t only?

libs/core/lib/tensor_impls/xtensor_impl.cpp
```diff
@@ -257,7 +263,31 @@ class xtensor : public cudaqx::details::tensor_impl<Scalar> {
...
   const Scalar *data() const override { return m_data; }
   void dump() const override {
-    std::cerr << xt::adapt(m_data, size(), xt::no_ownership(), m_shape) << '\n';
+    std::cout << xt::adapt(m_data, size(), xt::no_ownership(), m_shape) << '\n';
+  }
+
+  void dump_bits() const override {
+    if constexpr (is_complex<Scalar>::value) {
+      throw std::runtime_error("dump_bits() unsupported for complex types");
+    } else if (rank() == 1) {
```
justinlietz, line 272: Should we limit this API to integral types, or even uint8_t only?

libs/qec/lib/experiments.cpp
```diff
@@ -122,60 +118,64 @@ sample_memory_circuit(const code &code, operation statePrep,
...
   if (statePrep == operation::prep0 || statePrep == operation::prep1) {
     // run z basis
-    for (std::size_t shot = 0; shot < numShots; shot++) {
-      platform.set_exec_ctx(&ctx);
-      memory_circuit_mz(stabRound, prep, numData, numAncx, numAncz, numRounds,
-                        xVec, zVec);
-      platform.reset_exec_ctx();
-    }
+    result = cudaq::sample(opts, memory_circuit_mz, stabRound, prep, numData,
+                           numAncx, numAncz, numRounds, xVec, zVec);
```
justinlietz: Nice! This is looking so much better. Not needed now, but we can likely flatten out additional complexity by just passing in what type of measurement to do, and flatten `memory_circuit_mz`, `memory_circuit_mx`, and `__memory_circuit_stab` into one kernel.

justinlietz (approved):
This is awesome, and validates the explicit measurements change not just for performance reasons. This looks much more like how users should be driving cudaq for these types of workloads.

Just some minor comments in the actual code. Feel to to merge as is.

libs/qec/lib/experiments.cpp
```diff
@@ -122,60 +118,64 @@ sample_memory_circuit(const code &code, operation statePrep,
   cudaqx::tensor<uint8_t> syndromeTensor({numShots * numRounds, numCols});
   cudaqx::tensor<uint8_t> dataResults({numShots, numData});
 
+  cudaq::sample_options opts;
+  opts.shots = numShots;
+  opts.noise = noise;
+  opts.explicit_measurements = true;
+
```
melody-ren, line 123: Think it looks cleaner with 
`cudaq::sample_options opts{.shots = numShots, 
.noise = noise, 
.explicit_measurements = true};`

libs/core/lib/tensor_impls/xtensor_impl.cpp
```diff
@@ -93,7 +99,7 @@ class xtensor : public cudaqx::details::tensor_impl<Scalar> {
     if (m_data)
       std::free(m_data);
 
-    m_data = m_data = new Scalar[size];
+    m_data = new Scalar[size];
```
melody-ren, line 102: `m_data` is allocated using `new[]`, so in the destructor it should be deallocated with `delete[] m_data`.
Similarly, in the `copy` function, `m_data` should be deleted using `delete[]` instead of `std::free`.

(The above is not part of this PR, but figured I'd point it out)

libs/core/lib/tensor_impls/xtensor_impl.cpp
bmhowe23: > Should we limit this API to integral types, or even uint8_t only?

Maybe it's a stretch, but I was picturing this could be useful for sparse floating point matrices, too, so I'm inclined to leave it defined for any non-complex value.

libs/core/include/cuda-qx/core/tensor_impl.h
bmhowe23: > Should we limit this API to integral types, or even uint8_t only?

While it seems a little wasteful to use types like float, double, or complex here, I don't see a great reason to exclude them here. One can represent 0-9 in all of those types perfectly fine.

libs/qec/lib/experiments.cpp
bmhowe23: That's good to know!

libs/qec/lib/experiments.cpp
bmhowe23: Updated in f14e69b

libs/core/lib/tensor_impls/xtensor_impl.cpp
bmhowe23: That's a good point, but strictly speaking, we don't actually know if users used new or malloc when getting their pointers that they are providing to us, so I don't think there is a good way to guarantee correct, clean operation here. Luckily, all supported types (in libs/core/lib/tensor_impls/xtensor_impl.cpp) are primitive and therefore don't have any constructors/destructors, so I'm inclined to leave this as is for now since free and delete essentially operate the same for those types (as far as I know).

melody-ren: Thanks for the clarification!

PR #92, opened 2025-02-20
Use known Python path if we are running in a Python context

bmhowe23:
This will help when using environments that have multiple versions of Python installed (like we do in some of our cudaqx-dev images).


melody-ren (approved):
LGTM

bmhowe23:
I also tested this in the full wheel environment here: https://github.com/NVIDIA/cudaqx/actions/runs/13450205213, so merging now.

PR #95, opened 2025-02-22
Update CUDA-Q build patch

bmhowe23:
This is needed due to a change in a file in
https://github.com/NVIDIA/cuda-quantum/pull/2629

(as discovered in https://github.com/NVIDIA/cudaqx/actions/runs/13469726675)


bmhowe23:
Successfully ran [here](https://github.com/NVIDIA/cudaqx/actions/runs/13470196011), so merging now.

PR #98, opened 2025-02-25
[docs] Ability to replace patterns in the docs.

boschmitt:
Various C++ types are made available to Python through bindings. In our Python API documentation, these types are shown with full qualified names. These names are confusing and surface a implementation details that users shouldn't need to know about.

This change adds the ability to search and replace patterns in the generated documentation text.

Note: The scalability and maintainability of this solution don't seem good. However, I have not found another way of solving the issue.


bmhowe23 (approved):
Thanks, @boschmitt

PR #99, opened 2025-02-26
Support vector types for kwargs

bmhowe23:
This updates our kwarg mapping logic to support numpy array types (mapped to std::vector). This is especially needed for decoders that are going to provide per-channel error probabilities.


melody-ren (approved):
Thanks Ben. Looks good to me. One question—what's the reasoning behind having two nearly identical files?

bmhowe23:
> Thanks Ben. Looks good to me. One question—what's the reasoning behind having two nearly identical files?

Pending the CI tests, they can be consolidated. Pinging @amccaskey to make sure he doesn't know of another reason to keep them separate.

PR #110, opened 2025-03-05
Add an example of constructing a multi-round parity-check matrix

melody-ren:


docs/sphinx/examples/qec/python/generate_parity_check_matrix.py
```diff
@@ -0,0 +1,140 @@
...
+    for i in range(nRounds):
+        H_nrounds[i * rows:(i + 1) * rows, i * cols:(i + 1) * cols] = H
+    print("H_nrounds\n", H_nrounds)
+
+    # Append columns for measurement errors to H
+    Mz = construct_measurement_error_syndrome(distance, nRounds)
+    print("Mz\n", Mz)
+    assert H_nrounds.shape[0] == Mz.shape[
+        0], "Dimensions of H_nrounds and Mz do not match"
+    H_pcm = np.concatenate((H_nrounds, Mz), axis=1)
```
bmhowe23, line 63: Do we need anything to remove duplicate columns from `H_pcm`?

melody-ren: Can you show me an example? I don't think there are duplicate columns in H_pcm?

docs/sphinx/examples/qec/python/generate_parity_check_matrix.py
```diff
@@ -0,0 +1,140 @@
...
+    # Get the parity check matrix for a single round
+    Hz = code.get_parity_z()
+    H = np.array(Hz)
+
+    # Extends H to nRounds
+    rows, cols = H.shape
+    H_nrounds = np.zeros((rows * nRounds, cols * nRounds), dtype=int)
+    for i in range(nRounds):
+        H_nrounds[i * rows:(i + 1) * rows, i * cols:(i + 1) * cols] = H
+    print("H_nrounds\n", H_nrounds)
```
bmhowe23: Is this approach the same regardless of the number of error mechanisms in the circuit-level noise?

melody-ren: (My understanding is that) this example is using the repetition code, so extending the H works for X errors, but won't work on Z errors as repetition code doesn't detect them.

bmhowe23: OK, thanks. But if we add additional noise models into the memory circuit, i.e. something like this:
https://gist.github.com/bmhowe23/4c93d48783e76abce9fc01b008b0b832
Or this: `noise.add_all_qubit_channel("mz", cudaq.BitFlipChannel(0.1))`

Will it still work? Or will we need to update this code if/when we change anything about the circuit-level noise model?

docs/sphinx/examples/qec/python/generate_parity_check_matrix.py
bmhowe23: Yes, you're right ... I don't see any dups in this matrix either. I'll mark this one as resolved.

docs/sphinx/examples/qec/python/generate_parity_check_matrix.py
```diff
@@ -0,0 +1,140 @@
...
+def construct_measurement_error_syndrome(distance, nRounds):
+    num_stabilizers = distance - 1
+    num_mea_q = num_stabilizers * nRounds
+
+    syndrome_rows = []
+
+    # In this scheme, need two rounds for each measurement syndrome
+    for i in range(nRounds - 1):
+        for j in range(num_stabilizers):
+            syndrome = np.zeros((num_mea_q,), dtype=np.int8)
```
bmhowe23: Nit, just for consistency with our other code.
```suggestion
            syndrome = np.zeros((num_mea_q,), dtype=np.uint8)
```

docs/sphinx/examples/qec/python/generate_parity_check_matrix.py
```diff
@@ -0,0 +1,140 @@
...
+    # Create distance-3 repetition code
+    code = qec.get_code("repetition", distance=distance)
+
+    # Get the parity check matrix for a single round
+    Hz = code.get_parity_z()
+    H = np.array(Hz)
+
+    # Extends H to nRounds
+    rows, cols = H.shape
+    H_nrounds = np.zeros((rows * nRounds, cols * nRounds), dtype=int)
```
bmhowe23: Nit, just for consistency with our other code.
```suggestion
    H_nrounds = np.zeros((rows * nRounds, cols * nRounds), dtype=np.uint8)
```

docs/sphinx/examples/qec/python/generate_parity_check_matrix.py
```diff
@@ -0,0 +1,140 @@
...
+
+    return code, H_pcm, Mz
+
+
+code, H_pcm, Mz = get_code_and_pcm(distance, nRounds)
+
+# Define the error probability
+p = 0.01
+noise = cudaq.NoiseModel()
+noise.add_all_qubit_channel("x", qec.TwoQubitDepolarization(p), 1)
```
bmhowe23: If it still works, it might be nice to show how to add measurement noise in this example.
```suggestion
noise.add_all_qubit_channel("x", cudaq.Depolarization2(p), 1)
noise.add_all_qubit_channel("mz", cudaq.BitFlipChannel(p))
```
Edit: would also be nice to migrate the new built-in CUDA-Q noise model: Depolarization2

docs/sphinx/examples/qec/python/generate_parity_check_matrix.py
melody-ren: Mz type of errors work since we appended the pcm to account for measurement errors. In the case of the repetition code, we can only account for mz though. The same idea should probably work for a different code

melody-ren:
Closing since this is being combined into #123

PR #112, opened 2025-03-07
Test examples in workflow

melody-ren:
Test the examples whenever the library itself (either solvers or qec) is updated, or the examples are updated.


bmhowe23 (commented):
Thanks, Melody. In general, I think this looks good. I'd like to hold off merging this one until next week.

PR #120, opened 2025-03-10
Use new CUDA-Q noise models in examples

bmhowe23:


justinlietz (approved):
Looks good. Now that we can use these updates from cudaq, we should next look to deprecate the redundant noise types in the library.

PR #123, opened 2025-03-11
Add expressive noise model example

melody-ren:
Adapted from Justin's example, using `apply_noise` instead of custom gate


docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,111 @@
+import cudaq
+import cudaq_qec as qec
+import numpy as np
+from generate_parity_check_matrix import get_code_and_pcm
```
melody-ren: @bmhowe23 What do you think of combining #110 and this one like this? In this example, we show how you can drop an error anywhere you like in the code, how to generate a PCM from it, then decode it. This example and #110 uses the same code and the same error model.

bmhowe23: In my mind, one of the goals of the examples in the docs is to allow users who _don't_ have a cloned repo to copy/paste/modify code in their own sandbox, and unless I'm missing something, I'm afraid this would prevent that.

melody-ren: ah sorry, wasn't being clear. I meant to combine both examples into one file, instead of having to import another file like I had (that was just for convenience for me)

bmhowe23: I did a BeyondCompare diff between the .py file in this PR and the one in 110, and yes, I agree that they are similar enough to combine into a single file and single PR. However, we should show usage of `cudaq.apply_noise` somewhere in the docs. I think we are missing that right now.

bmhowe23: Oops, hit send too soon. I see it in one of the files now ... stand by.

bmhowe23: Yes, it sounds like this draft PR is effectively includes all of #110 in it, so yes, I'd be ok combining them. We'll just want to beef up the words right before the example, highlighting the fact that it shows how to generate a PCM and how to use a variety of noisy simulation features.

melody-ren: Sounds good. I'll combine the two PRs into a single file and highlight both PCM generation and the noise model features in the example and the docs.

Regarding `cudaq.apply_noise`, I’ll include usage examples in our docs to demonstrate how different types of noise can be introduced using it in the context of QEC.

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -1,5 +1,11 @@
 import cudaq
+import cudaq_qec as qec
 import numpy as np
+from generate_parity_check_matrix import get_code_and_pcm
```
bmhowe23: Maybe this is still WIP, but I thought the earlier conversation indicated that you were going to combine the files together so you wouldn't have to have `from generate_parity_check_matrix import get_code_and_pcm` here ... is that right?

melody-ren: I think you might be looking at an earlier version? This was removed in 957944e45b696ce9eaa8adf6aba49743f95c12e2

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,172 @@
+import cudaq
+import cudaq_qec as qec
+import numpy as np
+
+nRounds = 3
+nShots = 500
+p = 0.1
```
bmhowe23:
```suggestion
# Physical error rate
p_per_per_round = 0.01
p_per_mz = 0.01
```

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,172 @@
...
+    data_qubits = cudaq.qvector(3)
+    ancilla_qubits = cudaq.qvector(2)
+
+    # Initialize the logical |1> state as |111>
+    x(data_qubits)
+
+    for i in range(nRounds):
+        # Random Bit Flip Errors
+        for j in range(3):
+            cudaq.apply_noise(cudaq.XError, p, data_qubits[j])
```
bmhowe23:
```suggestion
            cudaq.apply_noise(cudaq.XError, p_per_round, data_qubits[j])
```

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,172 @@
...
+        reset(ancilla_qubits[1])
+
+    # Final measurement to get the data qubits
+    mz(data_qubits)
+
+
+# Create a noise model
+noise_model = cudaq.NoiseModel()
+# Add measurement noise
+noise_model.add_all_qubit_channel("mz", cudaq.BitFlipChannel(0.01))
```
bmhowe23:
```suggestion
noise_model.add_all_qubit_channel("mz", cudaq.BitFlipChannel(p_per_mz))
```

docs/sphinx/examples_rst/qec/circuit_level_noise.rst
```diff
@@ -75,3 +75,48 @@ Here's how to use CUDA-Q QEC to perform a circuit-level noise model experiment i
...
+            cudaq.apply_noise(cudaq.DepolarizationChannel, 0.1, q[0])
+
+            # Perform gate operations
+            h(q[1])
+            x.ctrl(q[1], q[2])
+
+            # Inject a Y error into the second qubit
+            cudaq.apply_noise(cudaq.YError, 0.1, q[1])
+
+            # Apply a general Pauli noise channel to the third qubit
```
bmhowe23:
```suggestion
            # Apply a general Pauli noise channel to the third qubit, where the 3 values indicate the probability of X, Y, and Z errors.
```

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
melody-ren: Applied in edbbb992ae6ef338a24fcacc5b7734e87282a54a

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,174 @@
...
+            syndrome[pos2] = 1
+
+            syndrome_rows.append(syndrome)
+
+    return np.array(syndrome_rows).T
+
+
+# Generate the parity check matrix for n-rounds by duplicating the input parity check matrix Hz
+# and appending the measurement error syndrome matrix.
+def get_code_and_pcm(distance, nRounds, Hz):
```
justinlietz: This function name doesn't quite refer to what is done. Perhaps it could just be called, "get_circuit_level_pcm"

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Extends H to nRounds
+    rows, cols = H.shape
+    H_nrounds = np.zeros((rows * nRounds, cols * nRounds), dtype=np.uint8)
+    for i in range(nRounds):
+        H_nrounds[i * rows:(i + 1) * rows, i * cols:(i + 1) * cols] = H
+    print("H_nrounds\n", H_nrounds)
+
+    # Construct the measurement error syndrome matrix for Z errors
+    Mz = construct_measurement_error_syndrome(distance, nRounds)
```
justinlietz: Maybe call this H_mz, so it's not so close in name to "mz"

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,174 @@
...
+    print("Mz\n", Mz)
+    assert H_nrounds.shape[0] == Mz.shape[
+        0], "Dimensions of H_nrounds and Mz do not match"
+
+    # Append columns for measurement errors to H
+    H_pcm = np.concatenate((H_nrounds, Mz), axis=1)
+    print(f"H_pcm:\n{H_pcm}")
+
+    return H_pcm, Mz
```
justinlietz: Why does Mz also need to be returned?

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,174 @@
...
+H_pcm, Mz = get_code_and_pcm(3, nRounds, Hz)
+
+# Get observables
+observables = np.array([1, 0, 0, 0, 0, 0], dtype=np.uint8)
+Lz = np.array([1, 0, 0], dtype=np.uint8)
+print(f"observables:\n{observables}")
+print(f"Lz:\n{Lz}")
+# Pad the observables to be the same dimension as the decoded observable
+Lz_nrounds = np.tile(Lz, nRounds)
+Lz_nround_mz = np.pad(Lz_nrounds, (0, Mz.shape[1]), mode='constant')
```
justinlietz: I see Mz is used here which makes sense to get the padding length. We could also get this by looking at the difference between Lz_nrounds and H_pcm. Then we don't need to return Mz. There's nothing wrong with doing it this way though.

justinlietz (approved):
This looks great. A very nice showcase of what can be done with the new QEC features, and also a goalpost for where we want the next wave of features to push towards.

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
melody-ren: Changed in ed0231898a3aab1157047f9d656d31ab5ec98954

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
melody-ren: Changed in ed0231898a3aab1157047f9d656d31ab5ec98954

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
melody-ren: Fixed in ed0231898a3aab1157047f9d656d31ab5ec98954

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
melody-ren: Fixed in ed0231898a3aab1157047f9d656d31ab5ec98954

bmhowe23 (approved):
Thanks!

PR #124, opened 2025-03-13
add CITATION.cff

caldwellshane:
This follows the pattern in CUDA-Q, but no Zenodo for now. I expect this to cause Github to populate a "Cite this repository" option in the About section on the repo homepage.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,17 @@
+cff-version: 1.2.0
+title: CUDA-QX
+message: >-
+  If you use CUDA-QX in your work, please cite it as well as
+  [CUDA-Q](https://github.com/NVIDIA/cuda-quantum). Thanks!
+type: software
+authors:
+  - name: NVIDIA Corporation, CUDA-QX Development Team
+    address: 2788 San Tomas Expy
+    city: Santa Clara
+    region: California
+    post-code: '95051'
+    tel: '+1 408-486-2000'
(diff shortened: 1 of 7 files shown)


CITATION.cff
```diff
@@ -0,0 +1,18 @@
...
+    address: 2788 San Tomas Expy
+    city: Santa Clara
+    region: California
+    post-code: '95051'
+    tel: '+1 408-486-2000'
+repository-code: 'https://github.com/NVIDIA/cudaqx'
+url: 'https://nvidia.github.io/cudaqx/'
+license: Apache-2.0
+license-url: 'https://github.com/NVIDIA/cudaqx/blob/main/LICENSE'
+version: '0.2'
```
caldwellshane: happy to leave this out - it's one more place to manually update the version number

bmhowe23: > happy to leave this out - it's one more place to manually update the version number

After thinking about it a bit, yeah, I think it would be best to leave this out for now. Once we have a more mature CI that auto-populates version numbers everywhere, then maybe we could add this back in.

caldwellshane:
```suggestion
```

bmhowe23 (approved):

bmhowe23:
Normally stuff would go through CI, but nothing in CI will do anything with this file, so I will push it in without it.

PR #125, opened 2025-03-14
Add copyright header to qec examples

melody-ren:


bmhowe23 (commented):
Thanks, Melody. While it is good to add this for the files themselves, can you please update all the docs that include these files to use `:start-after:` like https://github.com/NVIDIA/cuda-quantum/blob/50066dd99025c4eae2af150c327664328b8249cc/docs/sphinx/using/backends/cloud/nvqc.rst?plain=1#L36? (I.e. we don't need the copyrights to show up in the docs renderings themselves.)

bmhowe23 (approved):
Thanks, Melody.

PR #131, opened 2025-03-17
Capture misc workflows use to create the 0.2.0 release

bmhowe23:
Building release candidates requires special scripts due to the split public/private repos. Capture them here.

### DIFF
--- a/.github/workflows/build_docker.yaml
+++ b/.github/workflows/build_docker.yaml
@@ -1,0 +1,84 @@
+name: Build Combined Docker Images
+
+on:
+  workflow_dispatch:
+    inputs:
+      base_image:
+        type: string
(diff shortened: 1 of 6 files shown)


bmhowe23 (commented):
@melody-ren Please browse at your convenience. I don't think we will make any significant changes at this time since the release is complete, but it would be good for others to be aware of these files. I anticipate that we would only need them near release time, not during day-to-day activities.

melody-ren:
> @melody-ren Please browse at your convenience. I don't think we will make any significant changes at this time since the release is complete, but it would be good for others to be aware of these files. I anticipate that we would only need them near release time, not during day-to-day activities.


Thanks Ben. Will take a look.

PR #133, opened 2025-03-20
Add Git LFS instructions 

melody-ren:
- Add the instructions for using Git LFS
- Add a workflow to block PR that contains large files and restricted file extensions

### DIFF
--- a/.github/workflows/pr_file_check.yaml
+++ b/.github/workflows/pr_file_check.yaml
@@ -1,0 +1,44 @@
+name: Check for Large Files and Restricted Extensions
+
+on:
+  pull_request:
+    branches: 
+      - main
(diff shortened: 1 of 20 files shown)


.github/workflows/pr_file_check.yaml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          set-safe-directory: true
+          fetch-depth: 1
+
+      - name: Fetch base branch
+        run: git fetch origin ${{ github.event.pull_request.base.ref }} --depth=1
+
+      - name: Check for large files
+        run: |
+          MAX_SIZE=5M  # Set max file size limit
```
melody-ren, line 32: I set this to be 5M for now. Happy to change to some other more sensible numbers

PR #134, opened 2025-03-21
Update libraries for `cudaq::spin_op` breaking changes

bmhowe23:
This PR includes the required CUDA-QX changes that must be applied once https://github.com/NVIDIA/cuda-quantum/pull/2710 (from CUDA-Q) lands.

For that PR, most of the breaking changes are in C++, but a few Python tests needed to be updated as well. Python changes will likely be coming in the near future.

Note that this change actually removes more code than it adds. This is primarily due to updates in the test code. 

Here is a summary of the changes in this PR.

1. Update CMakeLists.txt files to link against `cudaq-operator` instead of `cudaq-spin` because `cuda-spin` was removed.
2. Make use of `cudaq::spin_op_term` (a product term) where appropriate. I currently left all the public QEC API's to use `spin_op` (a sum term), but we may choose to update them to simply `spin_op_term` in the near future. That would be a CUDA-QX breaking change if we do.
3. Since the new operators no longer store degrees that had implicit `I`'s in them, we must now clarify whether we want `I` in the terms. For QEC, this means that we will sometimes use something like `cudaq::spin_op_term identity(0, get_num_data_qubits())`, but for Solvers, we have eliminated all implicit `I`'s.
4. Use `get_pauli_word()` instead of `to_string(false)`.
5. Use `evaluate_coefficient()` instead of `get_coefficient()`.
6. Use `.trim()` to remove 0-cofficient terms and `.canonicalize()` to remove `I` operations from terms. (These new helper functions help reduce boiler-plate code.)
7. For default-constructed `cudaq::spin_op` objects, clarify whether you want them to be initialized to the neutral sum term (i.e. 0 = `spin_op::empty()`) or the neutral product term (i.e. 1 = `spin_op::identity()`).
8. Change `for_each_term()` to use iterators like `for (const auto &term : hamiltonian)`.
9. As needed, update code to reflect that `get_qubit_count()` now only returns the number of _actual_ degrees used in the operator rather than the "highest degree + 1" that used to be returned. If needed, make use of the new `min_degree()` and `max_degree()` methods instead.
10. For canned test data, update to use the new serialization format (as retrieved by `get_data_representation()`).
11. For test data comparisons, be sure to call `canonicalize` on the test data and/or truth data as necessary in order to ensure correct comparisons.
12. Don't use `get_raw_data()` anymore.


libs/solvers/lib/operators/graph/clique.cpp
```diff
@@ -18,16 +18,15 @@ cudaq::spin_op get_clique_hamiltonian(const cudaqx::graph &graph,
     return cudaq::spin_op();
 
   // Initialize empty spin operator
-  cudaq::spin_op hamiltonian(nodes.size());
+  cudaq::spin_op hamiltonian;
```
bettinaheim, line 21: FYI, you may still want to keep the constructor as it was - if you let it that way, it will reserve space for all the nodes.size() terms that you add later (actually, I guess its correct size in the end is larger - looks like nodes.size() + 3 * non_edges.size() +1 at most)

libs/solvers/python/tests/test_operator_pools.py
```diff
@@ -73,7 +73,7 @@ def test_uccsd_operator_pool_correctness():
         data_counter += 1
 
     # Assert
-    expected_operators = [["XZYI", "YZXI"], ["IXZY", "IYZX"],
+    expected_operators = [["XZY", "YZX"], ["IXZY", "IYZX"],
```
bettinaheim: nit: I thought you wanted to trim all ids in the solvers? Shouldn't the operators in the pool then also be trimmed? I don't think it matters, though - just a question.

bettinaheim (commented):
Looks pretty clean to me. :)

libs/solvers/lib/operators/operator_pools/spin_complement_gsd.cpp
```diff
@@ -49,20 +49,9 @@ spin_complement_gsd::generate(const heterogeneous_map &config) const {
...
-          return;
-
-        if (std::fabs(coeff.real()) < 1e-12)
-          terms.insert({term.get_raw_data().first[0],
-                        std::complex<double>{0., coeff.imag()}});
-      });
-
-      if (!terms.empty())
-        pool.emplace_back(terms);
+      oneElectron.trim();
```
bmhowe23: I just noticed this is an unintentional behavior change (1e-12 vs 0). I'll fix that.

libs/solvers/python/tests/test_operator_pools.py
bmhowe23: The core Solvers code is indeed stripping the identities everywhere; these remaining ones are an artifact of using "to_string(False)" in the Python test code for comparisons. However, I'm glad you brought it up because I should change the above `to_string(False)` code to use `get_pauli_word(4)` and then revert these `expected_operators` changes.

Ideally, I would like to change this `expected_operators` and `expected_coefficients` code for the upcoming Python PR when `canonicalize` is available in Python.

libs/solvers/lib/operators/graph/clique.cpp
bmhowe23: Yep, this is a subtle case where the breaking change in behavior (allocating space for `nodes.size()` _product terms_ instead of allocating a single term of `nodes.size()` _qubits_) actually does no harm; however, I think it makes the PR cleaner to make the change, so I'll probably leave my change as is here.

libs/solvers/python/tests/test_operator_pools.py
bmhowe23: Updated in df6bf6e

libs/solvers/lib/operators/operator_pools/spin_complement_gsd.cpp
bmhowe23: Updated in df6bf6e

PR #136, opened 2025-04-02
Correcting syndromes per shot in circuit-level example

justinlietz:
In the current python circuit-level noise example, all syndromes are mistakenly processed each shot, rather than only the syndromes contained within a shot.

This PR reshapes the syndromes into a new array, and corrects this behavior.


melody-ren:
Wow great catch Justin. Thanks!

melody-ren (approved):

PR #137, opened 2025-04-03
Bump CUDA-Q commit and pin CMake

bmhowe23:
Contains minor changes as a result of https://github.com/NVIDIA/cuda-quantum/pull/2765 and also updates our build scripts to keep CMake < 4.0 for now because some of our dependencies do not support 4.0 at the moment.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -70,7 +70,6 @@ jobs:
       - name: Build docs
         run: |
           cmake -S . -B "build" \
-            -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
             -DCMAKE_BUILD_TYPE=Release \
             -DCMAKE_C_COMPILER=gcc-11 \
             -DCMAKE_CXX_COMPILER=g++-11 \
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -64,7 +64,7 @@ jobs:
             sphinx_copybutton sphinx_inline_tabs sphinx_gallery sphinx_rtd_theme \
             sphinx_reredirects sphinx_toolbox cupy-cuda12x cuquantum-python-cu12
 
-          python3 -m pip install cmake --user
+          python3 -m pip install "cmake<4" --user
           echo "$HOME/.local/bin:$PATH" >> $GITHUB_PATH
 
       - name: Build docs
--- a/libs/solvers/lib/operators/molecule/molecule.cpp
+++ b/libs/solvers/lib/operators/molecule/molecule.cpp
@@ -6,6 +6,7 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 #include "cudaq/solvers/operators/molecule/molecule_package_driver.h"
+#include "cudaq/utils/cudaq_utils.h"
 
 #include <fstream>
 #include <iostream>
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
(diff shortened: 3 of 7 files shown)


.github/workflows/docs.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
             sphinx_copybutton sphinx_inline_tabs sphinx_gallery sphinx_rtd_theme \
             sphinx_reredirects sphinx_toolbox cupy-cuda12x cuquantum-python-cu12
 
-          python3 -m pip install cmake --user
+          python3 -m pip install "cmake<4" --user
```
melody-ren, line 67: I added `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` on line 73 because json-src was failing on the pipeline. Probably because the cmake version was too new. 

With `cmake<4`, I'm not sure if line 73 is still needed? Haven't tried

libs/solvers/lib/operators/molecule/molecule.cpp
```diff
@@ -6,6 +6,7 @@
  * the terms of the Apache License 2.0 which accompanies this distribution.    *
  ******************************************************************************/
 #include "cudaq/solvers/operators/molecule/molecule_package_driver.h"
+#include "cudaq/utils/cudaq_utils.h"
```
melody-ren, line 9: Just curious - why is this added? Seems like the pipeline wasn't failing before without this? Same question applies to `spin_complement_gsd.cpp`

bmhowe23: This is needed due to the changes in https://github.com/NVIDIA/cuda-quantum/pull/2765. IIRC, the molecule.cpp file was using `cudaq::split` and the gsd.cpp file was using `cudaq::range`. Both of those were defined in cudaq_utils.h. They were previously smuggled in via spin_op.h, but the new version of spin_op.h doesn't actually use anything in cudaq_utils.h, so it was removed in the aforementioned PR.

.github/workflows/docs.yaml
bmhowe23: I suspect it isn't needed. I'll back it out (3809706) and let the CI test it out.

bmhowe23: Also, fwiw, CUDA-Q now includes a higher CMake version in its Docker image that I believe is compatible with our build environment, so we could (or perhaps, should) update our build scripts to use that regular CMake rather than this strangely `pip install`-ed version of CMake. That can come in a later PR.

melody-ren: But then, I thought not all deps can support a higher CMake? Well I guess we can try it out later

libs/solvers/lib/operators/molecule/molecule.cpp
melody-ren: Oh I see. Thanks for the explanation, though it's somewhat concerning that things didn't break without `cudaq_util.h`. Maybe because we only do a weekly cudaq version update?

.github/workflows/docs.yaml
bmhowe23: The cudaq one is pinned to 3.28 I believe. (It used to be older than that and our build scripts used features that needed at least 3.28.)

libs/solvers/lib/operators/molecule/molecule.cpp
bmhowe23: Right. It only broke within the last week and this PR is the first one to try to bump the cudaq commit, so it was the first to run into the issue. I actually noticed it while doing development on my local machine today, so I knew it had to be fixed before running the CI.

melody-ren: Ok that makes sense. I was setting up a new container maybe a few days ago? Straight from the image, worked. Manual rebuilt cudaq to use the latest commit, solver side would fail. Didn't look into it, but this is probably why, I think

PR #141, opened 2025-04-09
Add error checking before constructing internal structures from Python data

bmhowe23:
Our Python bindings require that matrices be supplied in row-major ordering. This PR adds error checks to verify that and adds documentation to clarify that.


justinlietz (approved):
Nice changes to document are row-major assumption, and set up checks for the user. However this is a good place to mention that these issues would be better solved by handling data strides in the cudaq::tensor class. @amccaskey we may want to consider a strides update to the cudaq::tensor spec.

PR #143, opened 2025-04-14
[docs] Update install guide to clarify usage with no GPUs available

bmhowe23:


caldwellshane:
Change looks good.

Thinking about the Macbook user reading this docs page, it might make sense also to add this sentence to the note about getting `libgfortran`: "On MacOS, you can get `gfortran` by doing `brew install gcc` or `xcode-select --install`."

Ie. at the end of this note:
    CUDA-Q Solvers will require the presence of :code:`libgfortran`, which is
    not distributed with the Python wheel, for provided classical optimizers. If
    :code:`libgfortran` is not installed, you will need to install it via your
    distribution's package manager. On Debian based systems, you can install
    this with :code:`apt-get install gfortran`.

caldwellshane (approved):

caldwellshane (approved):

PR #152, opened 2025-04-30
Update libraries for Python spin op breaking changes

bmhowe23:
Upstream changes: https://github.com/NVIDIA/cuda-quantum/pull/2817

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "f5bf9d1cb08d74716ddcd1d848c5323ba751f25c"
+    "ref": "a09726530fa5e03dd9734ab4abb2da25f02c132a"
   }
 }
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "a09726530fa5e03dd9734ab4abb2da25f02c132a"
+    "ref": "da31e1b7af1cc46806bd34278fa1e1e670f8d88c"
   }
 }
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "75f99d0b012156836e38ad933e4ab28fef7aedd4"
+    "ref": "f5bf9d1cb08d74716ddcd1d848c5323ba751f25c"
   }
 }
--- a/libs/qec/python/bindings/py_code.cpp
+++ b/libs/qec/python/bindings/py_code.cpp
@@ -75,9 +75,13 @@ public:
           "{...}. Please provide the CUDA-Q kernels for the operation "
           "encodings.");
 
-    // Get the stabilizers
-    m_stabilizers =
-        registeredCode.attr("stabilizers").cast<std::vector<cudaq::spin_op>>();
+    // Get the stabilizers. First convert to spin_op_term's and then convert to
+    // spin_op's.
+    auto stab_terms = registeredCode.attr("stabilizers")
+                          .cast<std::vector<cudaq::spin_op_term>>();
+    for (auto &term : stab_terms)
+      m_stabilizers.emplace_back(term);
+
     // Get the CUDA-Q kernels for the operation encodings
     auto opsDict = registeredCode.attr("operation_encodings").cast<py::dict>();
 
--- a/libs/qec/python/bindings/py_code.cpp
+++ b/libs/qec/python/bindings/py_code.cpp
@@ -79,8 +79,9 @@ public:
     // spin_op's.
     auto stab_terms = registeredCode.attr("stabilizers")
                           .cast<std::vector<cudaq::spin_op_term>>();
+    m_stabilizers.reserve(stab_terms.size());
     for (auto &term : stab_terms)
-      m_stabilizers.emplace_back(term);
+      m_stabilizers.emplace_back(std::move(term));
 
     // Get the CUDA-Q kernels for the operation encodings
     auto opsDict = registeredCode.attr("operation_encodings").cast<py::dict>();
--- a/libs/qec/python/bindings/type_casters.h
+++ b/libs/qec/python/bindings/type_casters.h
@@ -24,7 +24,6 @@ struct type_caster<cudaq::spin_op> {
     auto className = src.attr("__class__").attr("__name__");
     if (className.cast<std::string>() == "SpinOperator") {
       auto data = src.attr("serialize")().cast<std::vector<double>>();
-      auto numQubits = src.attr("qubit_count").cast<std::size_t>();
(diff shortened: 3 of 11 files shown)


libs/solvers/python/bindings/utils/type_casters.h
```diff
@@ -21,10 +21,23 @@ struct type_caster<cudaq::spin_op> {
...
     if (!src)
       return false;
-    auto data = src.attr("serialize")().cast<std::vector<double>>();
-    auto numQubits = src.attr("get_qubit_count")().cast<std::size_t>();
-    value = cudaq::spin_op(data);
-    return true;
+    auto className = src.attr("__class__").attr("__name__");
+    if (className.cast<std::string>() == "SpinOperator") {
+      auto data = src.attr("serialize")().cast<std::vector<double>>();
+      auto numQubits = src.attr("qubit_count").cast<std::size_t>();
```
bettinaheim: I don't think you need numQubits here anymore, here or below.

libs/solvers/python/bindings/utils/type_casters.h
```diff
@@ -21,10 +21,23 @@ struct type_caster<cudaq::spin_op> {
   bool load(handle src, bool) {
     if (!src)
       return false;
-    auto data = src.attr("serialize")().cast<std::vector<double>>();
-    auto numQubits = src.attr("get_qubit_count")().cast<std::size_t>();
-    value = cudaq::spin_op(data);
-    return true;
+    auto className = src.attr("__class__").attr("__name__");
+    if (className.cast<std::string>() == "SpinOperator") {
```
bettinaheim: I can add the serialize and to_json for SpinOperatorTerm as well. The only reason I was hesitant to do that is because it would not roundtrip and serialize/deserialize a SpinOpTerm would produce a SpinOp instead. I could do something in from_json to always deserialize into a SpinOpTerm whenever the SpinOp consists of a single term, but that may be unexpected as well. What do you think is the best thing to do here regarding serialization?

libs/qec/python/bindings/type_casters.h
```diff
@@ -21,10 +21,21 @@ struct type_caster<cudaq::spin_op> {
   bool load(handle src, bool) {
     if (!src)
       return false;
-    auto data = src.attr("serialize")().cast<std::vector<double>>();
-    auto numQubits = src.attr("get_qubit_count")().cast<std::size_t>();
-    value = cudaq::spin_op(data);
-    return true;
+    auto className = src.attr("__class__").attr("__name__");
+    if (className.cast<std::string>() == "SpinOperator") {
```
bettinaheim: Same comments here as for the solvers type_casters below.

libs/qec/python/tests/test_code.py
```diff
@@ -44,7 +44,7 @@ def test_code_stabilizers():
     assert isinstance(stabilizers, list)
     assert len(stabilizers) == 6
     assert all(isinstance(stab, cudaq.Operator) for stab in stabilizers)
-    stabStrings = [s.to_string(False) for s in stabilizers]
+    stabStrings = [term.get_pauli_word() for s in stabilizers for term in s]
```
bettinaheim, line 47: I thought a stabilizer was now a spin_op_term rather than a spin_op?

libs/qec/python/bindings/py_code.cpp
```diff
@@ -75,9 +75,14 @@ class PyCodeHandle : public qec::code {
...
           "encodings.");
 
-    // Get the stabilizers
-    m_stabilizers =
-        registeredCode.attr("stabilizers").cast<std::vector<cudaq::spin_op>>();
+    // Get the stabilizers. First convert to spin_op_term's and then convert to
+    // spin_op's.
+    auto stab_terms = registeredCode.attr("stabilizers")
+                          .cast<std::vector<cudaq::spin_op_term>>();
+    m_stabilizers.reserve(stab_terms.size());
```
bettinaheim, line 82: Why did you choose to keep stabilizers as a list of spin_ops in the python code rather than making them a list of SpinOperatorTerms in python as well (if I remember correctly they are a vector of spin_op_terms in C++)?

libs/qec/python/tests/test_code.py
bmhowe23: I kept the interface the same as before (i.e. spin_op instead of spin_op_term) to minimize the changes in #134 since #134 could be used as a reference for others. After this PR, you're correct that it would be good to change to spin_op_term. I haven't decided if we are going to do that in a graceful deprecated way, or if we will just make a hard breaking change like some of the other spin op changes.

libs/qec/python/bindings/py_code.cpp
bmhowe23: They are currently a vector of spin_op's in C++, too. E.g., https://github.com/NVIDIA/cudaqx/blob/87401419dd6c0f2a9e08973ec714ab85808fcd84/libs/qec/include/cudaq/qec/stabilizer_utils.h#L24

libs/solvers/python/bindings/utils/type_casters.h
bmhowe23: > The only reason I was hesitant to do that is because it would not roundtrip and serialize/deserialize a SpinOpTerm would produce a SpinOp instead.

I don't understand why this would happen. Doesn't deserialization occur when invoking an explicit constructor? That is - if one invokes `SpinOperatorTerm(data)`, it seems like it should always return a `SpinOperatorTerm`. When would the conversion to `SpinOp` occur?

> I could do something in from_json to always deserialize into a SpinOpTerm whenever the SpinOp consists of a single term, but that may be unexpected as well. 

Yes, that sounds unintuitive to me. I think it would be best to keep the serialize/deserialize data types consistent and explicit.

In general, I think it would be good to add serialize and to_json/from_json to the `SpinOperatorTerm` class, but the code I have above also hides this complication for most users.

libs/qec/python/tests/test_code.py
bettinaheim: That makes sense. I think it would be nice to update the libraries to use terms when terms are the correct data type.

libs/solvers/python/bindings/utils/type_casters.h
bettinaheim: I pushed an update to the CUDA-Q PR to add to_json/from_json and .serialize + matching constructor for SpinOperatorTerm as well. The data format itself will be the same regardless of whether a sum or a product was serialized, so it is up to the user to invoke the appropriate construction function.

bmhowe23: Updated in 1822469

.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "75f99d0b012156836e38ad933e4ab28fef7aedd4"
+    "ref": "a09726530fa5e03dd9734ab4abb2da25f02c132a"
```
bmhowe23: This needs to be change to a commit on `main` before merging this PR.

melody-ren (approved):

PR #153, opened 2025-04-30
Add testcases for libs/core to improve the code coverage rate

caiyunh:
[B 5250408](https://nvbugs/5250408) [CUDA-QX][Code Coverage] Add testcase for libs/core
[B 5252687](https://nvbugs/5252687) [CUDA-QX] tensor::at calling a non-const method in a const method

### DIFF
--- a/libs/core/include/cuda-qx/core/tensor.h
+++ b/libs/core/include/cuda-qx/core/tensor.h
@@ -102,7 +102,8 @@ public:
   /// @param indices The indices of the element to access
   /// @return A const reference to the element at the specified indices
   const scalar_type &at(const std::vector<size_t> &indices) const {
-    return const_cast<const details::tensor_impl<Scalar>*>(pimpl.get())->at(indices);
+    return const_cast<const details::tensor_impl<Scalar> *>(pimpl.get())
+        ->at(indices);
   }
 
   /// @brief Copy data into the tensor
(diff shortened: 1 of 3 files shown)


libs/core/include/cuda-qx/core/tensor.h
```diff
@@ -102,7 +102,8 @@ class tensor {
   /// @param indices The indices of the element to access
   /// @return A const reference to the element at the specified indices
   const scalar_type &at(const std::vector<size_t> &indices) const {
-    return pimpl->at(indices);
+    return const_cast<const details::tensor_impl<Scalar> *>(pimpl.get())
+        ->at(indices);
```
bmhowe23, line 106: Just curious - why was this change needed? Doesn't the `const` on line 104 ensure that we are calling the `const` version of `at()`? I figured the compiler would throw an error if we were accidentally trying to call the non-const version here.

bmhowe23 (commented):
Thanks for the additional test coverage, @caiyunh! I just have one question about the change to tensor.h, mainly for my own edification. All tests seem to pass both with and without that change, so this is just a curiosity.

libs/core/include/cuda-qx/core/tensor.h
bmhowe23: Never-mind, I can see that this was calling the non-const one, and I agree that calling the const one is probably better.

bmhowe23 (approved):

PR #154, opened 2025-05-02
Add `__version__` info to packages

bmhowe23:
This PR makes it so that the following commands work.

```python
>>> print(solvers.__version__)
CUDA-Q Solvers 0.0.0 (https://github.com/NVIDIA/cudaqx a36bdffa2da8fb147c3b23d170ffcf8317600fcd)
>>> print(qec.__version__)
CUDA-Q QEC 0.0.0 (https://github.com/NVIDIA/cudaqx a36bdffa2da8fb147c3b23d170ffcf8317600fcd)
```

The actual version number (0.0.0) will be replaced if built with the `CUDAQX_QEC_VERSION` and/or `CUDAQX_SOLVERS_VERSION` environment variables set.

It also adds a `decoder.get_version()` function for our decoders.

### DIFF
--- a/.github/workflows/scripts/build_wheels.sh
+++ b/.github/workflows/scripts/build_wheels.sh
@@ -115,6 +115,8 @@ fi
 export CC=gcc
 export CXX=g++
 export SETUPTOOLS_SCM_PRETEND_VERSION=$wheels_version
+export CUDAQX_QEC_VERSION=$wheels_version
+export CUDAQX_SOLVERS_VERSION=$wheels_version
 
 # ==============================================================================
 # QEC library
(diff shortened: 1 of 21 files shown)


kvmto (approved):
Looks great Ben, thanks! Everything works flawless.
Just a curiosity, why does the decoder need a `get_version()`, if it takes the one of the library?

bmhowe23:
> Looks great Ben, thanks! Everything works flawless. Just a curiosity, why does the decoder need a `get_version()`, if it takes the one of the library?

The `nv-qldpc-decoder` (which is built separately) will have its own version string because its source code is not public.

kvmto:
> > Looks great Ben, thanks! Everything works flawless. Just a curiosity, why does the decoder need a `get_version()`, if it takes the one of the library?
> 
> The `nv-qldpc-decoder` (which is built separately) will have its own version string because its source code is not public.

I suspected :) thanks

melody-ren (approved):

PR #157, opened 2025-05-07
Incorporate workflow and test updates from releases/v0.2.1 branch

bmhowe23:


bmhowe23:
This was successfully merged, but GitHub API must have glitched because it is acting like it didn't. I'm going to close it now.

PR #160, opened 2025-05-10
Optimize `convert_vec_soft_to_tensor_hard` by 50x

bmhowe23:
This small PR optimizes one of our utility functions by >50x.

In the long term, we need to consider accessor methods for `cudaqx::tensor` that don't incur the performance penalty of allocating and deallocating vectors on every `.at()` access. As best as I can tell, this is not supported very well by the underlying xtensor class.

### DIFF
--- a/libs/qec/include/cudaq/qec/decoder.h
+++ b/libs/qec/include/cudaq/qec/decoder.h
@@ -168,10 +168,9 @@ template <typename t_soft, typename t_hard,
 inline void convert_vec_soft_to_hard(const std::vector<t_soft> &in,
                                      std::vector<t_hard> &out,
                                      t_soft thresh = 0.5) {
-  out.clear();
-  out.reserve(in.size());
-  for (auto x : in)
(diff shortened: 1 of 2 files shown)


melody-ren (commented):
Looks good. Thanks, Ben. 

Curious though - (1) why is the change not applied to the 2D version of `convert_vec_hard_to_soft`? (2) What prompted this change? Based on a bug, profiling results or something else?

bmhowe23:
> Looks good. Thanks, Ben.
> 
> Curious though - (1) why is the change not applied to the 2D version of `convert_vec_hard_to_soft`? (2) What prompted this change? Based on a bug, profiling results or something else?

The key change that had a big improvement was to `convert_vec_soft_to_tensor_hard`. The change to `convert_vec_hard_to_soft` was much more minor. I noticed it had a minor improvement in the test case I was dealing (to avoid the repeated `push_back`'s, but I didn't have any data to justify making a change to the 2D version.

I made these discoveries while working on the sliding window decoder prototypes, which was performing very poorly until these changes were made.

melody-ren (approved):

PR #165, opened 2025-05-17
Update QEC interfaces to use spin_op_term

bmhowe23:
This is a follow-up to https://github.com/NVIDIA/cudaqx/pull/134, bringing our QEC interfaces in line with the new CUDA-Q interfaces for spin op terms (and no longer using spin ops).

### DIFF
--- a/libs/qec/include/cudaq/qec/code.h
+++ b/libs/qec/include/cudaq/qec/code.h
@@ -1,5 +1,5 @@
 /****************************************************************-*- C++ -*-****
- * Copyright (c) 2024 NVIDIA Corporation & Affiliates.                         *
+ * Copyright (c) 2024 - 2025 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *
@@ -117,14 +117,14 @@ protected:
   std::unordered_map<operation, encoding> operation_encodings;
 
   /// @brief Stabilizer generators for the code
-  std::vector<cudaq::spin_op> m_stabilizers;
+  std::vector<cudaq::spin_op_term> m_stabilizers;
 
   /// @brief Pauli Logical operators
-  std::vector<cudaq::spin_op> m_pauli_observables;
+  std::vector<cudaq::spin_op_term> m_pauli_observables;
 
-  std::vector<cudaq::spin_op>
+  std::vector<cudaq::spin_op_term>
   fromPauliWords(const std::vector<std::string> &words) {
-    std::vector<cudaq::spin_op> ops;
+    std::vector<cudaq::spin_op_term> ops;
     for (auto &os : words)
       ops.emplace_back(cudaq::spin_op::from_word(os));
     sortStabilizerOps(ops);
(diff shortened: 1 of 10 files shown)


libs/qec/include/cudaq/qec/code.h
```diff
@@ -117,14 +117,14 @@ class code : public extension_point<code, const heterogeneous_map &> {
   std::unordered_map<operation, encoding> operation_encodings;
 
   /// @brief Stabilizer generators for the code
-  std::vector<cudaq::spin_op> m_stabilizers;
+  std::vector<cudaq::spin_op_term> m_stabilizers;
 
   /// @brief Pauli Logical operators
-  std::vector<cudaq::spin_op> m_pauli_observables;
+  std::vector<cudaq::spin_op_term> m_pauli_observables;
```
kvmto, line 123: Observables can be sums of terms. Is the vector a substitution for the sum, or shall we have a vector of sums? Maybe @justinlietz can you shed some light here?

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -160,11 +160,11 @@ class stabilizer_grid {
...
 
-  /// @brief Get the stabilizers as a vector of cudaq::spin_ops
-  std::vector<cudaq::spin_op> get_spin_op_stabilizers() const;
+  /// @brief Get the stabilizers as a vector of cudaq::spin_op_terms
+  std::vector<cudaq::spin_op_term> get_spin_op_stabilizers() const;
 
-  /// @brief Get the observables as a vector of cudaq::spin_ops
-  std::vector<cudaq::spin_op> get_spin_op_observables() const;
+  /// @brief Get the observables as a vector of cudaq::spin_op_terms
+  std::vector<cudaq::spin_op_term> get_spin_op_observables() const;
```
kvmto, line 167: Same idea as the other comment? Maybe the comment applies to wherever we need an observable beyond Lx and Lz? Several codes might need something beyond this.

libs/qec/include/cudaq/qec/code.h
bmhowe23: Agree, it would be good for @justinlietz to weigh in here.

justinlietz: TL;DR: `spin_op_term` is OK, as the `std::vector` is a stack of pauli product operators, so we are not implying a sum of terms as would be common in the Solvers library.

You're right that in general "observables" can be sums of terms. However, in the QEC literature, the language is overloaded and "observable" typically refers specifically to logical X and logical Z operators. In the QEC library, we've allowed them to be specified by any pauli product (`spin_op_term`), so we've just referred to them generically as `pauli_observables`. Logical X and Z are not unique in QEC codes, so this is a user defined representative. It's precise goal in the library is to define which physical qubit measurements need to be `xor`'d together to give a logical qubit measurement.

Why a `std::vector`?  We can track 2 observables per logical qubit, the X and Z logical pauli operators. While we've currently made certain assumptions about a code holding only 1 logical qubit, we intend to relax this. A code block with `k` logical qubits can support `2k` observables. A last feature is convenience. Given `2k` observables, we can matrix multiply (base 2) into the measurements to compute all of the observables at once if they are all stacked.

libs/qec/include/cudaq/qec/codes/surface_code.h
justinlietz: Since these are logical qubits, you can think about what observables are accessible for a single physical qubit. We can access all of the quantum information of a physical qubit by measuring it's pauli observables `X,Y,Z`.

In an algorithm setting, when we talk about an observable like H = 0.5XX + 0.1ZZ, this is something that composed out of X's and Z's. Here we are just specifying how to access each logical qubits X or Z. Later, when we're doing algorithms on logical qubits, we will want to talk about sums like H = 0.5XX + 0.1ZZ, but hopefully it will be clear at that stage that which abstraction layer we are talking about (logical vs physical).

melody-ren (approved):

PR #171, opened 2025-05-20
Add `opt_results` field to `decoder_result` 

melody-ren:
Hey all, 
I’d like to get your thoughts on a proposed change to the `decoder_result` interface. Currently, it holds two values: `converged` and `result`. We’re considering adding a third item — a heterogeneous map — to allow the decoder to return additional outputs such as BP history and BP iterations.

There are two options:

Option 1: Add the heterogeneous map and specialize `tuple_size`. This preserves backward compatibility and avoids making the change breaking.

Option 2: Add the heterogeneous map without specializing `tuple_size`. This introduces a breaking change, as structured bindings would now expect three elements.

Let me know your thoughts — I'm also open to other ideas or suggestions.

**Update:**
The decision was to adopt option 2 and expose all three fields to the python side.

### DIFF
--- a/libs/qec/include/cudaq/qec/decoder.h
+++ b/libs/qec/include/cudaq/qec/decoder.h
@@ -13,6 +13,7 @@
 #include "cuda-qx/core/tensor.h"
 #include <future>
 #include <vector>
+#include <optional>
 
 namespace cudaq::qec {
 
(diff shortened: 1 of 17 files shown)


libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -31,10 +32,39 @@ struct decoder_result {
...
+
+  /// @brief Check if optional results exist
+  bool has_optional_results() const { return optional_results.has_value(); }
+
+  /// @brief Get optional results if they exist, otherwise return empty map
+  cudaqx::heterogeneous_map get_optional_results() const {
+    return optional_results.value_or(cudaqx::heterogeneous_map());
+  }
+
+  // Tuple-like interface for structured binding
```
melody-ren: Option 1: specialize `tuple_size`. This approach preserves backward compatibility by keeping `tuple_size<decoder_result>` as 2, which allows existing code like `auto [conv, result] = decoder->decode(syndrome)` to continue working. However, the newly added `optional_results` field won’t be available through structured bindings and must be accessed via a named accessor or directly from the struct.

E.g., you can access the `optional_results` like this `auto opt_res = decoder->decode(syndrome).optional_results` or use the `get_optional_results()` function, but this `auto [conv, result, opt_res] = decoder->decode(syndrome);` won't compile.

libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -31,10 +32,39 @@ struct decoder_result {
...
   // Manually define the equality operator
   bool operator==(const decoder_result &other) const {
-    return std::tie(converged, result) ==
-           std::tie(other.converged, other.result);
+    // First compare the non-optional fields
+    if (std::tie(converged, result) != std::tie(other.converged, other.result)) {
+      return false;
+    }
+
+    // Handle optional_results comparison - just check if both have or don't have it
```
melody-ren: Also, there doesn't seem to be an easy to way to check equality between two `heterogeneous_map` objects. Suggestions?

libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -43,6 +73,44 @@ struct decoder_result {
   }
 };
 
+
+/// @brief Decoder results
+struct decoder_result_2 {
```
melody-ren: Option 2: Keep `tuple_size` as 3. This will be a breaking change. 

`auto [conv, result] = decoder->decode(syndrome);` will need to be updated to `auto [conv, result, unused] = decoder->decode(syndrome);`.

bmhowe23: I'm interested to hear what others think, but I favor this option right now. Specializing `tuple_size`, while pretty clever, makes me a bit nervous about readability/maintainability, but perhaps that's unfounded.

libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -31,10 +32,39 @@ struct decoder_result {
   /// each index.
   std::vector<float_t> result;
 
+  /// @brief Optional additional results from the decoder stored in a heterogeneous map
+  std::optional<cudaqx::heterogeneous_map> optional_results;
+
+  /// @brief Check if optional results exist
+  bool has_optional_results() const { return optional_results.has_value(); }
```
amccaskey: if you end up adding this, I would drop this and the getter, since it's a struct you can just access it directly and use the std::optional methods directly

bmhowe23 (commented):
Thanks, Melody. Would please update libs/qec/lib/decoders/plugins/example/single_error_lut_example.cpp to populate something in the opt_result if (and only if) the corresponding flag is set in the custom args for the single error LUT constructor? That would provide a concrete example of how we anticipate this being used.

libs/qec/lib/decoders/single_error_lut.cpp
```diff
@@ -20,11 +20,60 @@ class single_error_lut : public decoder {
...
+      "syndrome_weight",   // Number of non-zero syndrome measurements (bool)
+      "decoding_time",     // Time taken to perform the decoding (bool)
+      "num_repetitions"    // Number of repetitions to perform (int > 0)
+  };
+
+  bool has_opt_results;
+  bool error_probability;
+  bool syndrome_weight;
+  bool decoding_time;
+  int num_repetitions;
```
bmhowe23:
```suggestion
  bool has_opt_results = false;
  bool error_probability = false;
  bool syndrome_weight = false;
  bool decoding_time = false;
  int num_repetitions = 0;
```

libs/qec/lib/decoders/single_error_lut.cpp
```diff
@@ -20,11 +20,60 @@ class single_error_lut : public decoder {
...
+          }
+          throw std::runtime_error(error_msg);
+        } else {
+          has_opt_results = true;
+          error_probability =
+              requested_results.get<bool>("error_probability", false);
+          syndrome_weight =
+              requested_results.get<bool>("syndrome_weight", false);
+          decoding_time = requested_results.get<bool>("decoding_time", false);
+          num_repetitions = requested_results.get<int>("num_repetitions", 0);
```
bmhowe23:
```suggestion
          error_probability =
              requested_results.get<bool>("error_probability", error_probability);
          syndrome_weight =
              requested_results.get<bool>("syndrome_weight", syndrome_weight);
          decoding_time = requested_results.get<bool>("decoding_time", decoding_time);
          num_repetitions = requested_results.get<int>("num_repetitions", num_repetitions);
```

bmhowe23 (commented):
Thanks, Melody ... one pair of minor comments and then this should be good to go.

PR #175, opened 2025-05-31
Update doc and example to use decoder_result new API

melody-ren:
Ideally, should go in after #171

### DIFF
--- a/docs/sphinx/components/qec/introduction.rst
+++ b/docs/sphinx/components/qec/introduction.rst
@@ -1012,7 +1012,7 @@ Example of running a memory experiment:
               syndrome[i] = syndromes.at({shot, i});
 
             // Decode syndrome
-            auto [converged, v_result] = decoder->decode(syndrome);
+            auto results = decoder->decode(syndrome);
             // Process correction
             // ...
           }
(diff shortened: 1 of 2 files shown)


bmhowe23 (approved):
👍

bmhowe23:
> We need to update `docs/sphinx/examples/qec/python/nv-qldpc-decoder.py` to reflect this change. The CI essentially skips that test because the nv-qldpc-decoder isn't part of our public CI tests, so we missed it.
> 
> That being said - the fact that our docs are immediately updated means that users using 0.3.0 looking at our docs would be led astray if they used it the new way, e.g. `bp_converged, dec_result, opt_result = nv_dec_gpu_and_cpu.decode(syndrome)`.
> 
> I think we should update our examples to not do things like `a, b, c = decode(...)`
> 
> Additionally, I think I've discovered that it is often noticeably slower to do `a, b, c = decode(...)` because our Python bindings are making multiple copies of the data to do that. (First they are copied into a temporary decoder_result object and then they are copied into the a, b, and c. Then the decoder_result goes out of scope. We could improve our bindings to do less copies, but usability gets tricky if objects from the parent object go out of scope. I'm inclined to make our examples stop showing `a, b, c = decode(...)`. It's better performance and less likely to encounter future breaking changes.

melody-ren:
Examples updated in #184

PR #179, opened 2025-06-06
Tensor Network Decoder

npancotti:
This pull request adds a tensor network decoder to the python code base at `libs/qec/python/cudaq_qec/plugins/decoders/`

The main entry point is the `TensorNetworkDecoder` class defined in `tensor_network_decoder.py`. The file contains the basic functionality to manipulate tensor networks and to dispatch to different backends. 

Other files in `decoders/tensor_network_utils/`

* `noise_model.py`: helper function to create tensor network representation of noise models
* `contractors.py`: the tensor network contractors backends and path finders
* `stim_interface.py`: helper functions to extract a parity_check_matrix & and error model from a `stim.DetectorErrorModel`
 

### Features

* Construct a tensor network from an arbitrary parity-check matrix (code capacity & circuit level noise)
* Use arbitrary noise models and logical obervables
    + Correlated, inhomogeneous, disordered 
* Convert a [Stim](https://github.com/quantumlib/Stim) [Detector Error Model](https://github.com/quantumlib/Stim/blob/main/doc/file_format_dem_detector_error_model.md) to a *parity-check matrix* + *logical observables* + *noise model*, and construct the corresponding tensor network(s)
* Support different CPU and GPU backends via numpy, torch, and cuQuantum (cuTensorNet)

### Initialization

```
import cudaq_qec as qec
import numpy as np

# parity check matrix
H = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
# logical observable
logicals = np.array([[0, 0, 1]])
# noise model
noise_model = 0.01 * np.random.rand(3)

decoder = qec.get_decoder(
    "tensor_network_decoder", 
    H, 
    logicals=logicals, 
    noise_model=noise_model, 
    contract_noise_model=True,
    # contractor_name="torch", # to use torch array type
    # device="cuda:0", # to load the tensors on a gpu via torch 
)
```

### Decoding

Single syndrome
```
syndrome = [0, 1, 1]
res = decoder.decode(syndrome, return_probability=True)
print(f"Results. res.result = {res.result}, converged = {res.converged}")
```

a batch of syndromes
```
syndrome_batch = np.array([[0, 1, 1], [1, 0, 1], [0, 0, 1]])
res = decoder.decode_batch(syndrome_batch, return_probability=True)
print([r.result for r in res])
```

#### New dependencies

* torch
* quimb
* opt_einsum
* stim


libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,761 @@
...
+                        self._device)
+        self.full_tn = (self.code_tn | self.logicals_tn | self.syndrome_tn |
+                        self.noise_model)
+
+        if contract:
+            for ie in self.error_inds:
+                self.full_tn.contract_ind(ie)
+
+    @classmethod
+    def from_stim_detector_error_model(
```
npancotti: We might want to pull this constructor out from the class, and just provide an example. I think the preferred instantiation is via `cudaq_qec.get_decoder(decoder_name)` anyways

npancotti:
Surprising things I noticed on the way: the order I import stim and cudaq_qec matters.
```
(.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import stim 
>>> import cudaq_qec
>>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
>>> 
```
```
(.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cudaq_qec
>>> import stim
>>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
Segmentation fault (core dumped)
(.venv) root@f75d8b092fe5:/workspaces/cudaqx# 
```

bmhowe23:
> Surprising thing I noticed on the way: the order I import stim and cudaq_qec matters.
> 
> ```
> (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import stim 
> >>> import cudaq_qec
> >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> >>> 
> ```
> 
> ```
> (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import cudaq_qec
> >>> import stim
> >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> Segmentation fault (core dumped)
> (.venv) root@f75d8b092fe5:/workspaces/cudaqx# 
> ```

Hmm, we will need to fix this. CUDA-Q includes a Stim-backed simulator. Perhaps it is re-exporting some Stim symbols that are conflicting with the true Python stim. Can you reproduce this a standalone CUDA-Q Python file? If so, I think it would be helpful to post a CUDA-Q issue here: https://github.com/NVIDIA/cuda-quantum/issues

npancotti:
> > Surprising thing I noticed on the way: the order I import stim and cudaq_qec matters.
> > ```
> > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> > Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> > Type "help", "copyright", "credits" or "license" for more information.
> > >>> import stim 
> > >>> import cudaq_qec
> > >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> > >>> 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> > Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> > Type "help", "copyright", "credits" or "license" for more information.
> > >>> import cudaq_qec
> > >>> import stim
> > >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> > Segmentation fault (core dumped)
> > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# 
> > ```
> 
> Hmm, we will need to fix this. CUDA-Q includes a Stim-backed simulator. Perhaps it is re-exporting some Stim symbols that are conflicting with the true Python stim. Can you reproduce this a standalone CUDA-Q Python file? If so, I think it would be helpful to post a CUDA-Q issue here: https://github.com/NVIDIA/cuda-quantum/issues

You mean to install CUDA-Q and Stim in the same environment and see if I get the same errror?

bmhowe23:
> > > Surprising thing I noticed on the way: the order I import stim and cudaq_qec matters.
> > > ```
> > > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> > > Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> > > Type "help", "copyright", "credits" or "license" for more information.
> > > >>> import stim 
> > > >>> import cudaq_qec
> > > >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> > > >>> 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```
> > > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> > > Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> > > Type "help", "copyright", "credits" or "license" for more information.
> > > >>> import cudaq_qec
> > > >>> import stim
> > > >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> > > Segmentation fault (core dumped)
> > > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# 
> > > ```
> > 
> > 
> > Hmm, we will need to fix this. CUDA-Q includes a Stim-backed simulator. Perhaps it is re-exporting some Stim symbols that are conflicting with the true Python stim. Can you reproduce this a standalone CUDA-Q Python file? If so, I think it would be helpful to post a CUDA-Q issue here: https://github.com/NVIDIA/cuda-quantum/issues
> 
> You mean to install CUDA-Q and Stim in the same environment and see if I get the same errror?

Yes, please

npancotti:
> > > > Surprising thing I noticed on the way: the order I import stim and cudaq_qec matters.
> > > > ```
> > > > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> > > > Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> > > > Type "help", "copyright", "credits" or "license" for more information.
> > > > >>> import stim 
> > > > >>> import cudaq_qec
> > > > >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> > > > >>> 
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > ```
> > > > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# python 
> > > > Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
> > > > Type "help", "copyright", "credits" or "license" for more information.
> > > > >>> import cudaq_qec
> > > > >>> import stim
> > > > >>> circuit = stim.Circuit.generated("surface_code:rotated_memory_z", rounds=3, distance=3)
> > > > Segmentation fault (core dumped)
> > > > (.venv) root@f75d8b092fe5:/workspaces/cudaqx# 
> > > > ```
> > > 
> > > 
> > > Hmm, we will need to fix this. CUDA-Q includes a Stim-backed simulator. Perhaps it is re-exporting some Stim symbols that are conflicting with the true Python stim. Can you reproduce this a standalone CUDA-Q Python file? If so, I think it would be helpful to post a CUDA-Q issue here: https://github.com/NVIDIA/cuda-quantum/issues
> > 
> > 
> > You mean to install CUDA-Q and Stim in the same environment and see if I get the same errror?
> 
> Yes, please

Can confirm this happens in a lean environment with cudaq and stim only. I will open an issue..

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/stim_interface.py
```diff
@@ -0,0 +1,185 @@
+# ============================================================================ #
+# Copyright (c) 2023 - Oscar Higgott                                           #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+# This file originated from:
```
bmhowe23: If this file stays in, we need to add a note to https://github.com/NVIDIA/cudaqx/blob/main/NOTICE.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -0,0 +1,120 @@
+# ============================================================================ #
+# Copyright (c) 2024 NVIDIA Corporation & Affiliates.                          #
```
bmhowe23: Nit: all these copyrights for the Nvidia files should say 2025 instead of 2024.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,760 @@
...
+
+    is represented as the tensor network:
+
+        r1          r2      < row indices (stored lazily)
+        |  \      / |
+        H   H   H   H       < Hadamard matrices
+        |   |  /    |
+        c1  c2      c3      < column indices (stored lazily)
+
+    This function can be used to the tensor network of the code and the tensor network of
```
bmhowe23: Nit: ?
```suggestion
    This function can be used to create the tensor network of the code and the tensor network of
```

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,760 @@
...
+    return arrays
+
+
+def tensor_network_from_syndrome_batch(
+    detection_events: npt.NDArray[Any],
+    syndrome_inds: List[str],
+    batch_index: str = "batch_index",
+    tags: Optional[List[str]] = None,
+) -> TensorNetwork:
+    """Build a tensor network from a batch of syndromes.
```
bmhowe23: Does this learn the noise model from the syndromes alone? How does it know how many error mechanisms there are?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,760 @@
...
+
+    The core, tensor-network graph is identical to a Tanner graph.
+    The tensor network is a bipartite graph with two types of nodes:
+    - Check nodes (c_i's) and error nodes (e_j's), delta tensors.
+    - Hadamard matrix (H) for each check-error pair.
+
+    Then, the tensor network is extended by the noise model, the logical observables,
+    and the product state / batch of syndromes.
+
+    For example,
```
bmhowe23, line 43: +1 for the nice ASCII art 😄

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,760 @@
...
+    code_tn: TensorNetwork
+    logicals_tn: TensorNetwork
+    syndrome_tn: TensorNetwork
+    noise_model: TensorNetwork
+    full_tn: TensorNetwork
+    check_inds: List[str]
+    error_inds: List[str]
+    logical_inds: List[str]
+    logical_obs_inds: List[str]
+    logicals_tags: List[str]
```
bmhowe23: Starting with Python 3.9, we tend to use `list` rather than `typing.List` in most of the CUDA-Q(X) code base. Could you please update all your code to replace `List` with `list`? Let me know if that presents any problems.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,760 @@
...
+
+        self.path_batch = _adjust_default_path_value(self.path_batch,
+                                                     is_cutensornet)
+        self.path_single = _adjust_default_path_value(self.path_single,
+                                                      is_cutensornet)
+
+    def decode(
+        self,
+        syndrome: List[bool],
+        logical_observable: Optional[str] = None,
```
bmhowe23: Is there a version of this that takes soft syndromes (probabilities)?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,760 @@
...
+        return_probability: bool = False,
+    ) -> "qec.DecoderResult":
+        """
+        Decode the syndrome by contracting the full tensor network.
+
+        Args:
+            syndrome (List[bool]): The syndrome ordered as the check indices.
+            logical_observable (Optional[str], optional): The index of the logical observable to use.
```
bmhowe23: How would this handle QLDPC-like codes where you want to return multiple logical observables? Can it only do the calculation for one of them at a time?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,760 @@
...
+    def optimize_path(
+        self,
+        output_inds: Any,
+        optimize: Any = None,
+        syndrome_batch: Optional[npt.NDArray[Any]] = None,
+    ) -> Any:
+        """Optimize the contraction path of the tensor network.
+
+        Args:
+            output_inds (tuple[str]): The output indices of the contraction.
```
bmhowe23: This one makes it seem like multiple observables can be used. Am I reading this right?

bmhowe23 (commented):
This is looking really good, Nico. We'll want to do an update to the docs to highlight this when it's done. (That can be done in a separate PR.)

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: LEt's check that is rendered correctly

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: This is independent from the errors, it just takes a list of syndromes and builds a TN representation of it. That TN will later be plugged into the full tensor net

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: Yes, it's easy to support soft syndromes in the TN language, it's even more natural. Do we want to support it in this PR?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: We agreed offline to support only one single logical operator. I also added a helper function to replace that operator if the user wants to probe more than one logical operator.
See the method `replace_logical_observable`

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
bmhowe23: Yes, that would be more consistent with our existing interfaces. E.g. https://github.com/NVIDIA/cudaqx/blob/7fc54ec61b980dcb2c878c102f222956ca45468f/libs/qec/include/cudaq/qec/decoder.h#L136 and https://github.com/NVIDIA/cudaqx/blob/7fc54ec61b980dcb2c878c102f222956ca45468f/libs/qec/python/bindings/py_decoder.cpp#L31.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
bmhowe23: Looks like this was resolved in 66e3af7e6663b1c7a6a4e2e7458e59142fadf428, so I'm marking this as resolved.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
bmhowe23: Was this one resolved? It looks like the issue is still in the latest version if I'm reading it correctly.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
bmhowe23: Did you verify that this rendered correctly? If not, do you need any help viewing the docs locally?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
bmhowe23: I'm guessing this was resolved in 43993329858f877f27a8370e9056ae8545c7b27a, so marking as resolved for now.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
bmhowe23: Just confirming based on other conversations .... it looks like I was NOT reading this correctly when I made this comment, right?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/stim_interface.py
npancotti: File removed

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: Should be resolved now

libs/qec/python/tests/test_tensor_network_decoder.py
```diff
@@ -0,0 +1,384 @@
...
+        np.testing.assert_allclose(single_results,
+                                   expected_cast,
+                                   rtol=rtol,
+                                   atol=atol)
+
+        # Compare to expected results
+        np.testing.assert_allclose(batch_results,
+                                   expected_cast,
+                                   rtol=rtol,
+                                   atol=atol)
```
kvmto, line 327: Adding at the end of the file

`if __name__ == "__main__":
    pytest.main()`
    
would help in executing the tests independently

libs/qec/python/cudaq_qec/plugins/decoders/tests/test_tensor_network_decoder.py
```diff
@@ -0,0 +1,186 @@
...
+    if not torch.cuda.is_available():
+        pytest.skip('No GPUs available, skip test')
+    t = torch.tensor([[1.0, 2.0], [3.0, 4.0]],
+                     dtype=torch.float32,
+                     device="cuda")
+    out = tensor_to_cpu(t, backend="torch", dtype="float64")
+    assert isinstance(out, torch.Tensor)
+    np.testing.assert_array_equal(out.cpu().numpy(), t.cpu().numpy())
+    assert out.dtype == torch.float64
+    assert not out.is_cuda
```
kvmto: Adding at the end of the file

`if __name__ == "__main__":
    pytest.main()`
    
would help in executing the tests independently

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/tests/test_tensor_network_utils.py
```diff
@@ -0,0 +1,135 @@
...
+        assert t.inds == error_index_pairs[i]
+        assert tags[i] in t.tags
+
+
+def test_error_pairs_noise_model_default_tags():
+    error_index_pairs = [('x', 'y')]
+    error_probabilities = [np.array([[0.6, 0.4], [0.3, 0.7]])]
+    tn = error_pairs_noise_model(error_index_pairs, error_probabilities)
+    for t in tn.tensors:
+        assert "NOISE" in t.tags
```
kvmto: Adding at the end of the file

`if __name__ == "__main__":
    pytest.main()`
    
would help in executing the tests independently

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -0,0 +1,120 @@
...
+        optimize (str, optional): Optimization strategy. Defaults to "auto".
+        slicing (tuple, optional): Slicing specification. Defaults to empty tuple.
+
+    Returns:
+        torch.Tensor: The contracted tensor.
+    """
+    return torch.einsum(subscripts, *tensors)
+
+
+torch_compiled_contractor = torch.compile(einsum_torch)
```
kvmto: I personally don't think we should leave something like this as a global variable. I would directly return it from the original function.
I am afraid in future changes this is going to be forgotten in refactoring.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -0,0 +1,120 @@
...
+        optimize (str, optional): Optimization strategy. Defaults to "auto".
+        slicing (tuple, optional): Slicing specification. Defaults to empty tuple.
+
+    Returns:
+        Any: The contracted tensor.
+    """
+    return oe.contract(subscripts, *tensors, optimize=optimize)
+
+
+torch_compiled_opt_einsum = torch.compile(contractor)
```
kvmto: as before:

I personally don't think we should leave something like this as a global variable. I would directly return it from the original function.
I am afraid in future changes this is going to be forgotten in refactoring.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def cutn_contractor(
+    subscripts: str,
+    tensors: list[Any],
```
kvmto: I think this should be a list of NDArray

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -0,0 +1,120 @@
...
+# this is used to determine the backend of the tensor arrays
+BACKENDS: list[str] = ["numpy", "torch"]
+
+# this is used to determine the contractor to use
+CONTRACTORS: Dict[str, Callable] = {
+    "numpy": contractor,
+    "torch_compiled_opt_einsum": torch_compiled_opt_einsum,
+    "torch": contractor,
+    "cutensornet": cutn_contractor,
+}
```
kvmto: I don't think this is the right place for these global variables. 
I would move both inside the decoder, since these global variables are just used to check if the rest of the flags are acceptable.
I would pass these at construction if we don't want them as private members of the function.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+    full_tn: TensorNetwork
+    check_inds: list[str]
+    error_inds: list[str]
+    logical_inds: list[str]
+    logical_obs_inds: list[str]
+    logical_tags: list[str]
+    _contractor_name: str
+    _backend: str
+    _dtype: str
+    _device: str
```
kvmto, line 95: Are we sure these have to be private member of the class? These are invoked in the tests. See other comment.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+        val: The current value.
+        is_cutensornet (bool): Whether the contractor is cutensornet.
+
+    Returns:
+        Any: The adjusted value.
+    """
+    if is_cutensornet:
+        return None if val == "auto" else val
+    else:
+        return "auto" if val is None else val
```
kvmto: Is there a specific reason why this function is outside of the class? It has a private member naming and it is invoked inside the decoder.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -0,0 +1,120 @@
...
+    return oe.contract(subscripts, *tensors, optimize=optimize)
+
+
+torch_compiled_opt_einsum = torch.compile(contractor)
+
+
+def cutn_contractor(
+    subscripts: str,
+    tensors: list[Any],
+    optimize: Optional[Any] = None,
```
kvmto: I personally would try to be more specific for future use and usability. Maybe an Optional[string | Option] I am making a guess. I think typing this  `optimize: Any` is hard but it would help the docs and maintain the code.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -0,0 +1,120 @@
...
+# this is used to determine the contractor to use
+CONTRACTORS: Dict[str, Callable] = {
+    "numpy": contractor,
+    "torch_compiled_opt_einsum": torch_compiled_opt_einsum,
+    "torch": contractor,
+    "cutensornet": cutn_contractor,
+}
+
+
+def optimize_path(optimize: Any, output_inds: tuple[str, ...],
```
kvmto, line 91: Same comment for `optimize: Any`

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+        dtype (str, optional): The data type to use. Defaults to "float64".
+        device (Optional[str], optional): The device to use. Defaults to None.
+    """
+    if device is None or device == "cpu":
+        tn.apply_to_arrays(lambda x: tensor_to_cpu(x, backend, dtype))
+    elif "cuda" in device:
+        tn.apply_to_arrays(lambda x: tensor_to_gpu(x, dtype, device))
+
+
+def set_backend(contractor_name: str, device: str) -> str:
```
kvmto: I would put this function inside the class since it is a check.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+    return do(
+        "array",
+        data,
+        like=backend,
+        dtype=to_backend_dtype(dtype, like=backend),
+        device=device,
+    )
+
+
+def set_tensor_type(
```
kvmto: I would move this inside the class and make it a private member, since it is a functionality used only by the tn decoder. Unless there is a reason why this should be outside.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+import numpy.typing as npt
+from quimb import oset
+from quimb.tensor import Tensor, TensorNetwork
+from autoray import do, to_backend_dtype
+import torch
+
+from .tensor_network_utils.contractors import BACKENDS, CONTRACTORS, optimize_path
+
+
+def tensor_network_from_parity_check(
```
kvmto: I would move all these tensor_* functions in a python file inside utils called maybe "tensor_block_builders.py" or "tensor_factories.py", together with prepare_syndrome_data_batch.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+        ], (f"Device {device} cannot be used with cuTensornet."
+            f"Please use one of the following devices: "
+            f"Available devices: {[f'cuda:{i}' for i in range(torch.cuda.device_count())]}"
+           )
+        return "torch"
+    # CPU contractor + various backends
+    else:
+        for b in BACKENDS:
+            if b in contractor_name:
+                return b
```
kvmto: What happens if the name is not present?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+        self.logical_obs = logical_obs.copy()
+        self.logical_tn = tensor_network_from_parity_check(
+            self.logical_obs,
+            col_inds=self.error_inds,
+            row_inds=self.logical_inds,
+            tags=self.logical_tags,
+        )
+
+        # Add a Hadamard tensor for each logical observable for its outer leg
+        self.logical_tn |= tensor_network_from_logical_observable(
```
kvmto: Would it be possible to use more explicit operations like "union()", this syntax is unique to quimb so I am just thinking if it could create trouble to non quimb users and maintainers. Just wondering :)

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+            logical_tags=logical_tags,
+        )
+
+        # Initialize the syndrome tensor network with no errors.
+        self.syndrome_tn = tensor_network_from_single_syndrome(
+            [0.0] * len(self.check_inds), self.check_inds)
+
+        # Construct the tensor network of code + logical + syndromes
+        # The noise model is added later
+        self.full_tn = self.code_tn | self.logical_tn | self.syndrome_tn
```
kvmto: same as below :)

libs/qec/python/tests/test_tensor_network_decoder.py
```diff
@@ -0,0 +1,384 @@
...
+    import torch
+    if torch.cuda.is_available():
+        assert decoder._contractor_name == "cutensornet"
+        assert decoder._backend == "torch"
+        assert decoder._device == "cuda:0"
+    else:
+        assert decoder._contractor_name == "numpy"
+        assert decoder._backend == "numpy"
+        assert decoder._device == "cpu"
+    assert decoder._dtype == "float32"
```
kvmto: are these fields supposed to be private?

libs/qec/python/tests/test_tensor_network_decoder.py
```diff
@@ -0,0 +1,384 @@
...
+                              noise_model=noise)
+
+    # New logical observable and indices
+    new_logical = np.array([[0, 1, 1]])
+    new_logical_inds = ["l_1"]
+    new_logical_tags = ["LOG_1"]
+
+    decoder.replace_logical_observable(logical_obs=new_logical,
+                                       logical_inds=new_logical_inds,
+                                       logical_tags=new_logical_tags)
```
kvmto, line 81: from a usability perspective I wouldn't like the generic user that just wants to decode, to let them hardcode the indices and tags for replacing the logical. I would make some checks on the new logical and make the change automatically.

libs/qec/python/tests/test_tensor_network_decoder.py
```diff
@@ -0,0 +1,384 @@
...
+    decoder = qec.get_decoder("tensor_network_decoder",
+                              H,
+                              logical_obs=logical,
+                              noise_model=noise)
+
+    # Generate a batch of random syndromes
+    batch = np.random.choice([False, True], size=(n_batch, n_checks))
+    batch = batch.astype(np.float64, copy=False)  # Ensure float64 dtype
+
+    for contractor, dtype, device in contractors:
```
kvmto: Can we add some form of logging? This is the only test that covers the full execution for all the options and it will be the first one to break as soon as we change something. It is quite bulky :)
I would like to have something telling me for which values and options it fails probably.
Some for the other tests above :)

libs/qec/python/tests/test_tensor_network_decoder.py
```diff
@@ -0,0 +1,384 @@
...
+@pytest.mark.parametrize("contractor,dtype,device,expect_gpu", [
+    ("numpy", "float64", "cpu", False),
+    ("torch", "float32", "cpu", False),
+    ("torch", "float64", "cpu", False),
+    ("torch_compiled_opt_einsum", "float64", "cpu", False),
+    ("torch_compiled_opt_einsum", "float32", "cpu", False),
+    ("cutensornet", "float32", "cuda:0", True),
+    ("cutensornet", "float64", "cuda:0", True),
+])
+def test_set_contractor_variants(contractor, dtype, device, expect_gpu):
```
bmhowe23: In the long term, I'm not sure if we want to support changing contractors after the decoder is constructed, especially if it simplifies testing or maintenance. If one wants to change how the decoder is operating, I think that - for the most part - we just want them to create a new decoder object. We can discuss a bit more, though. (The reason I bring this is up now is that this test is failing after I made some local changes, but the comment is somewhat independent of those changes.)

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
bmhowe23: I don't actually see this one being used anywhere. I'm wondering if it could be removed. (All local tests still pass if I remove it.)

melody-ren: I recall Nico mentioned this was done for performance/speed reasons during his walkthrough. Or I might just be misremembering @npancotti ? 

If I'm not hallucinating, maybe we should add a comment to document it here

bmhowe23: (Though admittedly, I did not test timing/performance with this quick change.)

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
```diff
@@ -0,0 +1,72 @@
...
+    tensors = []
+
+    if tensors_tags is None:
+        tensors_tags = ["NOISE"] * len(error_indices)
+
+    for ei, eprob, etag in zip(error_indices, error_probabilities,
+                               tensors_tags):
+        tensors.append(
+            Tensor(
+                data=np.array([1.0 - eprob, eprob]),
```
melody-ren, line 53: Might want to add a line to check if eprob is between [0.0, 1.0] in case someone passes in an array of LLRs by mistake. We can gracefully fail it here.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/tests/test_tensor_network_utils.py
```diff
@@ -0,0 +1,135 @@
+# ============================================================================ #
+# Copyright (c) 2025 NVIDIA Corporation & Affiliates.                          #
```
melody-ren: Just a reminder: let's move this to /python/tests

libs/qec/python/cudaq_qec/plugins/decoders/tests/test_tensor_network_decoder.py
```diff
@@ -0,0 +1,186 @@
+# ============================================================================ #
```
melody-ren: Just a reminder: let's move this to /python/tests

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
bmhowe23: Nit: this looks like it should perhaps be named `get_backend`, because I don't actually see it "setting" anything in here.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Returns:
+        TensorNetwork: The tensor network representing the error pairs noise model.
+    """
+    tensors = []
+
+    if tensors_tags is None:
+        tensors_tags = ["NOISE"] * len(error_index_pairs)
+
+    for ei, etensors, etag in zip(error_index_pairs, error_probabilities,
```
melody-ren, line 95: Likewise here. Not sure how much "bad input" checks we want to do here.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+
+        contraction_value = CONTRACTORS[self._contractor_name](
+            self.full_tn.get_equation(output_inds=(self.logical_obs_inds[0],)),
+            self.full_tn.arrays,
+            optimize=self.path_single,
+            slicing=self.slicing_single,
+        )
+
+        res = qec.DecoderResult()
+        res.converged = True
```
melody-ren, line 414: `converged` doesn't really apply here, though to conform with the API, we still need to set it. I think for clarity, we should add a note about this in the docstring. Maybe something like "Always true for tensor network decoder; contraction is non-iterative"? 

Longer term discussion: It's going to be another breaking change... but, there might be merit to move `converged` into `opt_results` instead

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+        res.result = [
```
melody-ren, line 415: Do we want to add an option to return hard decisions?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -0,0 +1,756 @@
...
+    ) -> "qec.DecoderResult":
+        """
+        Decode the syndrome by contracting exactly the full tensor network.
+
+        Args:
+            syndrome (list[float]): 
+                The syndrome soft decision probabilities ordered as the check indices.
+
+        Returns:
+            qec.DecoderResult: The result of the decoding.
```
melody-ren, line 388: I think we should clarify here that the returned probability is the logical error rate.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: I think we clarified this in private discussion, let me know if I can resolve the conversation

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
npancotti: I removed the torch-compiled versions in the latest commits.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
npancotti: NDArray or torch.Tensor. Modified.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
npancotti: removed

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
npancotti: I defined a new `ContractorConfig` dataclass to handle these issues all together.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: They are now stored in `ContractorConfig`

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: I moved it inside the class, not as a method though. Let me know if you are ok with the solution

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
npancotti: Added more docs

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: Function no longer needed

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
npancotti: added more context

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: done

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: done

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: removed in favor of `ContractorConfig`

libs/qec/python/tests/test_tensor_network_decoder.py
npancotti: now stored in `ContractorConfig`

libs/qec/python/tests/test_tensor_network_decoder.py
npancotti: I think they are already optional

```
    def replace_logical_observable(
            self,
            logical_obs: npt.NDArray[Any],
            logical_inds: Optional[list[str]] = None,
            logical_tags: Optional[list[str]] = None) -> None:
        """Add logical observables to the tensor network.
        Args:
            logical_obs (np.ndarray): The logical matrix.
            logical_inds (Optional[list[str]], optional): The logical indices. If None, defaults to [l_0, l_1, ...].
            logical_obs_inds (Optional[list[str]], optional): The logical observable indices. If None, defaults to [l_obs_0, l_obs_1, ...].
            logical_tags (Optional[list[str]], optional): The logical tags. If None, defaults to [LOG_0, LOG_1, ...].
        """
        assert logical_obs.shape == (1, len(self.error_inds)), (
            "logical must be a single row matrix, shape (1, n), where n is the number of errors."
            "Only single logical are supported for now.")
        if logical_inds is None:
            self.logical_inds = ["l_0"]  # Index before the Hadamard tensor
        else:
            self.logical_inds = logical_inds

        if logical_tags is None:
            self.logical_tags = ["LOG_0"]
        else:
            self.logical_tags = logical_tags
```

libs/qec/python/tests/test_tensor_network_decoder.py
npancotti: What information would be useful to get, beyond pytest logs?

libs/qec/python/tests/test_tensor_network_decoder.py
npancotti: Removed

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: Removed

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/tests/test_tensor_network_utils.py
npancotti: Done

libs/qec/python/cudaq_qec/plugins/decoders/tests/test_tensor_network_decoder.py
npancotti: done

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: I guess it's easy enough for a user to do?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: done

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: This would look something like
```
self.full_tn = TensorNetwork()
self.full_tn = self.full_tn.combine(self.code_tn, virtual=True)
self.full_tn = self.full_tn.combine(self.logical_tn, virtual=True)
self.full_tn = self.full_tn.combine(self.syndrome_tn, virtual=True)
```
Let me know if you prefer this version.

npancotti: https://quimb.readthedocs.io/en/latest/autoapi/quimb/tensor/tensor_core/index.html#quimb.tensor.tensor_core.TensorNetwork.combine

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: We could develop a concept of `converged` here if, for example, the flop count or the mem requirement is too high once we compute the path. We just don't contract and say that we did not converge

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
npancotti: adding checks all around

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
kvmto: great!

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
kvmto: great!

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
kvmto: super !

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
kvmto: yes, I think this slight verbosity would add some clarity

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
kvmto: I would use combine also here

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
melody-ren: Not sure where it's mentioned? But we can address this in a later PR focusing on docs and examples. Marking as resolved for now.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
melody-ren: Yea, was just thinking of it as a developer "quality of life" sort of thing to add. We can easily add this later if we really want to. Marking as resolved for now.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
melody-ren: I don't see where the checks are added?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
melody-ren: I like the idea but I think this might warrant a longer discussion for later. For now though, marking it as always true and explaining it in the docs should be sufficient. Marking as resolved now.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: Shall we address the rendering and the docs/example issues on a different PR? In favor to let this one in smoothly?

bmhowe23: Yes, that sounds good to me. Marking this one as resolved.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
bmhowe23: Yep, I'll mark as resolved.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: done

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
npancotti: done

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
npancotti: you're right, I put it everywhere except here.. done

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
melody-ren: Thanks Nico. Not for this PR, but it would be nice to add a few tests to confirm that these checks work as intended. Actually, I can probably just add them at a later time. Marking as resolved.

libs/qec/python/tests/test_tensor_network_decoder.py
kvmto: it would be great having something like this at least:

```    
try:
    result = risky_operation()
    assert result is not None
except Exception as e:
    logging.error(f"Test failed due to: {e}")
    pytest.fail(f"Operation failed: {e}")
```

libs/qec/python/tests/test_tensor_network_decoder.py
kvmto: I think this hasn't been addressed?

npancotti: I think it was in L565

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/tests/test_tensor_network_utils.py
npancotti: All tests were moved to `python/tests/test_tensor_network_decoder.py` I added that line at the bottom of the file.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
npancotti: This should be resolved now

bmhowe23 (approved):
Looks great ... thanks, Nico!

PR #186, opened 2025-06-08
Add move constructor to `heterogeneous_map`

melody-ren:
This PR adds a move constructor and move assignment operator to `heterogeneous_map`, enabling the compiler to move rather than copy the map when it’s returned from a function. This addition should improve performance, particularly when returning large maps such as BP history via `decoder_result`.

### DIFF
--- a/libs/core/include/cuda-qx/core/heterogeneous_map.h
+++ b/libs/core/include/cuda-qx/core/heterogeneous_map.h
@@ -55,15 +55,6 @@ public:
   /// @brief Default constructor
   heterogeneous_map() = default;
 
-  /// @brief Copy constructor
-  /// @param _other The map to copy from
-  heterogeneous_map(const heterogeneous_map &_other) { *this = _other; }
-
-  /// @brief Move constructor
-  /// @param _other The map to move from
-  heterogeneous_map(heterogeneous_map &&_other) noexcept
-      : items(std::move(_other.items)) {}
-
   /// @brief Constructor from initializer list
   /// @param list The initializer list of key-value pairs
   heterogeneous_map(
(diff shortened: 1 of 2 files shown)


libs/core/include/cuda-qx/core/heterogeneous_map.h
```diff
@@ -59,6 +59,11 @@ class heterogeneous_map {
   /// @param _other The map to copy from
   heterogeneous_map(const heterogeneous_map &_other) { *this = _other; }
 
+  /// @brief Move constructor
+  /// @param _other The map to move from
+  heterogeneous_map(heterogeneous_map &&_other) noexcept
+      : items(std::move(_other.items)) {}
```
bmhowe23: Thanks, @melody-ren. Is there any reason these can't be done with `heterogeneous_map(heterogeneous_map &&_other) noexcept = default;`? (Same for the copy constructor, actually.)

melody-ren: They totally can. I'm not sure why `copy` was done explicitly and here `move` was basically following `copy`'s example. I can change both to default if you prefer?

bmhowe23: Yes, please. In general, I prefer using `default` whenever possible as that makes it more clear that "this is regular, nothing special going on here". 😄

melody-ren: Actually, now that I think about it, since there's truly "nothing special going on" in this class and it's just a wrapper over an `unordered_map`, we should remove all explicitly defined special member functions. They are all using `=default` anyway. Agree?

bmhowe23: If you're asking about copy/move constructors and copy/move assignment operators, then yes, I agree. But are you also asking about defaulting other member functions, too? If so, which ones?

melody-ren: Oh just these four - copy/move constructors and copy/move assignment operators

PR #198, opened 2025-06-30
Create sliding window decoder

bmhowe23:
This PR creates a new `"sliding_window"` decoder. The new decoder accepts parameters for an inner decoder (and the corresponding inner decoder parameters) over which the sliding window decoder "slides" a decoding window across.


docs/sphinx/api/qec/sliding_window_api.rst
```diff
@@ -0,0 +1,119 @@
+.. class:: sliding_window
```
bmhowe23, line 1: I'm including these docs here, but they are not referenced anywhere. I will open an issue to track the inclusion of these docs into the published website when we get closer to the release.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+  std::size_t num_windows = 0;
+  std::size_t num_rounds = 0;
+  std::size_t num_syndromes_per_window = 0;
+  std::vector<std::unique_ptr<decoder>> inner_decoders;
+  std::vector<std::size_t> first_columns;
+  cudaqx::tensor<std::uint8_t> full_pcm;
+  cudaqx::tensor<std::uint8_t> full_pcm_T;
+
+  // Constants
+  static constexpr int NUM_WINDOW_PROC_TIMES = 10;
```
melody-ren: How is this number determined?

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+    straddle_end_round =
+        params.get<bool>("straddle_end_round", straddle_end_round);
+    error_rate_vec = params.get<std::vector<cudaq::qec::float_t>>(
+        "error_rate_vec", error_rate_vec);
+    inner_decoder_name =
+        params.get<std::string>("inner_decoder_name", inner_decoder_name);
+    inner_decoder_params = params.get<cudaqx::heterogeneous_map>(
+        "inner_decoder_params", inner_decoder_params);
+
+    // Perform error checking on the inputs.
```
melody-ren: Nit: I'd suggest to put some of the `params` validation logic into a validation function to keep the constructor lean (I'm fully aware that I stuffed plenty of these checks into constructors myself.... but still....)

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+                 first_column, last_column);
+      auto inner_decoder =
+          decoder::get(inner_decoder_name, H_round, inner_decoder_params_mod);
+      inner_decoders.push_back(std::move(inner_decoder));
+    }
+  }
+
+  virtual decoder_result decode(const std::vector<float_t> &syndrome) override {
+    if (syndrome.size() == this->syndrome_size) {
+      auto t0 = std::chrono::high_resolution_clock::now();
```
melody-ren, line 296: Can you please put these timing related commands into `#infdef` blocks? It'd be nice to have the option to opt out of clocking the timing performance.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+    }
+    if (inner_decoder_name.empty()) {
+      throw std::invalid_argument(
+          "sliding_window constructor: inner_decoder_name must be non-empty");
+    }
+    if (inner_decoder_params.empty()) {
+      CUDAQ_WARN("sliding_window constructor: inner_decoder_params is empty. "
+                 "Is that intentional?");
+    }
+    if (step_size == 0) {
```
melody-ren: Should probably be non-negative?

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+      throw std::invalid_argument(
+          "sliding_window constructor: error_rate_vec must be non-empty");
+    }
+
+    // Enforce that H is already sorted.
+    if (!cudaq::qec::pcm_is_sorted(H, num_syndromes_per_round)) {
+      throw std::invalid_argument("sliding_window constructor: PCM must be "
+                                  "sorted. See cudaq::qec::simplify_pcm.");
+    }
+
```
melody-ren, line 260: Need to add a validation check for `window_size`, and we expect it to be within [1, num_rounds]

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+namespace cudaq::qec {
+
+/// @brief This is a sliding window decoder that receives syndromes on a
+/// round-by-round basis, and decodes them according window-specific parameters
+/// provided in the decoder.
+class sliding_window : public decoder {
+private:
+  // Input parameters.
+  std::size_t window_size = 1;
+  std::size_t step_size = 1;
```
melody-ren, line 27: Can you please add a comment to define what `step_size` is? It seems to mean the number of rounds that the window advances by. If this is the intended meaning, then we need to add a check in the constructor such that `step_size` won't cause the windows to be skipping rounds between windows.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+        // }
+        CUDAQ_DBG("Returning decoder_result");
+        return std::move(this->rw_results);
+      }
+    }
+    CUDAQ_DBG("Returning empty decoder_result");
+    return std::vector<decoder_result>(); // empty return value
+  }
+
+  void decode_window() {
```
melody-ren, line 417: Looking at the test files and the .rst file, my guess is that this function is intended to be an internal function called by the other two decode APIs. If that's the intention, then we should make this function private and document that it's not meant to be directly called by the user.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+        CUDAQ_DBG("Returning decoder_result");
+        return std::move(this->rw_results[0]);
+      }
+    }
+    CUDAQ_DBG("Returning empty decoder_result");
+    return decoder_result(); // empty return value
+  }
+
+  virtual std::vector<decoder_result>
+  decode_batch(const std::vector<std::vector<float_t>> &syndromes) override {
```
melody-ren, line 358: Is it possible to refactor `decode` and `decode_batch`? They share a fair bit of common code. I think extracting the common logic out will improve maintainability.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+  virtual decoder_result decode(const std::vector<float_t> &syndrome) override {
+    if (syndrome.size() == this->syndrome_size) {
+      auto t0 = std::chrono::high_resolution_clock::now();
+      CUDAQ_DBG("Decoding whole block");
+      // Decode the whole thing, iterating over windows manually.
+      decoder_result result;
+      std::vector<float_t> syndrome_round(step_size * num_syndromes_per_round);
+      for (std::size_t w = 0; w < num_rounds; ++w) {
+        std::copy(syndrome.begin() + w * step_size * num_syndromes_per_round,
+                  syndrome.begin() +
```
melody-ren: This might be at the risk of slicing out of bound of `syndrome` when `step_size` > 1.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+          decoder::get(inner_decoder_name, H_round, inner_decoder_params_mod);
+      inner_decoders.push_back(std::move(inner_decoder));
+    }
+  }
+
+  virtual decoder_result decode(const std::vector<float_t> &syndrome) override {
+    if (syndrome.size() == this->syndrome_size) {
+      auto t0 = std::chrono::high_resolution_clock::now();
+      CUDAQ_DBG("Decoding whole block");
+      // Decode the whole thing, iterating over windows manually.
```
melody-ren, line 298: (I'm rather confused about this whole-block branch and the steaming mode starting at line 158). My following comments are based on the assumption that there are two modes of operation - (1) we have received the whole-block of syndrome and (2) we expect to receive one round per decode call in the streaming mode.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+                  syndrome_round.begin());
+        result = decode(syndrome_round);
+      }
+      auto t1 = std::chrono::high_resolution_clock::now();
+      std::chrono::duration<double> diff = t1 - t0;
+      CUDAQ_INFO("Whole block time: {:.3f} ms", diff.count() * 1000);
+      return result;
+    }
+    // Else we're receiving a single round.
+    if (rw_filled == 0) {
```
melody-ren, line 314: Minor: this can probably be factored out into an `init` function

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+      window_proc_times_arr[0] =
+          std::chrono::duration<double>(t1 - t0).count() * 1000;
+    }
+    if (this->rw_filled == num_syndromes_per_window) {
+      auto t0 = std::chrono::high_resolution_clock::now();
+      CUDAQ_DBG("Window is full, sliding the window");
+      std::copy(this->rolling_window[0].begin() + num_syndromes_per_round,
+                this->rolling_window[0].end(), this->rolling_window[0].begin());
+      std::copy(syndrome.begin(), syndrome.end(),
+                this->rolling_window[0].end() - num_syndromes_per_round);
```
melody-ren: Here we are advancing the window by one round per decode call, which is not always the amount of data being sent in by the whole-block decode branch when `step_size` > 1. I suggest that we keep the implementation consistent and have the whole-block branch sending in one round per call. 

Secondary point: here we are using `copy` to slide the rolling window. Might not be the most efficient way when window size gets large. Maybe consider using a circular buffer with index tracking?

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+      auto t1 = std::chrono::high_resolution_clock::now();
+      window_proc_times_arr[1] +=
+          std::chrono::duration<double>(t1 - t0).count() * 1000;
+    } else {
+      // Just copy the data to the end of the rolling window.
+      auto t0 = std::chrono::high_resolution_clock::now();
+      CUDAQ_DBG("Copying data to the end of the rolling window");
+      std::copy(syndrome.begin(), syndrome.end(),
+                this->rolling_window[0].begin() + this->rw_filled);
+      this->rw_filled += num_syndromes_per_round;
```
melody-ren: Assuming that the expected behaviour of the streaming mode is to receive one round per decode call, then line 195 to 197 is the correct logic. I do think this also means that we need to feed in one round per call in the whole-block branch to align with this logic here.

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+      std::fill(window_proc_times.begin(), window_proc_times.end(), 0.0);
+      rw_filled = 0;
+      CUDAQ_DBG("Initializing window");
+      auto t1 = std::chrono::high_resolution_clock::now();
+      window_proc_times_arr[0] =
+          std::chrono::duration<double>(t1 - t0).count() * 1000;
+    }
+    if (this->rw_filled == num_syndromes_per_window) {
+      auto t0 = std::chrono::high_resolution_clock::now();
+      CUDAQ_DBG("Window is full, sliding the window");
```
melody-ren: I got confused by this DBG message. I thought it meant that we are moving to the next window (aka, advancing by `step_size` rounds). It seems to mean we are advancing by one round. Is that what you meant here?

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+      auto t0 = std::chrono::high_resolution_clock::now();
+      CUDAQ_DBG("Copying data to the end of the rolling window");
+      std::copy(syndrome.begin(), syndrome.end(),
+                this->rolling_window[0].begin() + this->rw_filled);
+      this->rw_filled += num_syndromes_per_round;
+      auto t1 = std::chrono::high_resolution_clock::now();
+      window_proc_times_arr[2] +=
+          std::chrono::duration<double>(t1 - t0).count() * 1000;
+    }
+    if (rw_filled == num_syndromes_per_window) {
```
melody-ren: After window 0 gets filled, `rw_filled` will stay as `num_syndromes_per_window` for all subsequent decode calls (I don't think `rw_filled` is cleared until the very end?). If my understanding is correct, this means that we are effectively calling `decode_window` every **round** (instead of every window) after the first window (this is based on the assumption that in streaming mode, we get one round in each decode call). This seems to ignore `step_size`'s effect. It will probably also cause us to skip the tail end of the whole-block since we'd be over-counting `num_windows_decoded`. 

Perhaps we can use a different counter (num_rounds_since_last_decode for example?) to keep track when to call decode_window. E.g., after the first window (window 0) becomes full, decode once and sets the counter to 0. In each subsequent streaming mode call, append one round and increase the counter. Once the counter reaches `step_size`, call decode_window and clear the counter.

melody-ren:
Thanks Ben! I have some questions about how the window logic is supposed to work. My review is based on assumptions about the two modes of operation (whole-block vs streaming), and I may have misunderstood the intended behavior... Could you clarify the expected flow for step_size > 1?

docs/sphinx/api/qec/sliding_window_api.rst
```diff
@@ -0,0 +1,119 @@
...
+
+    The sliding window decoder is a wrapper around a standard decoder that
```
kvmto: Not urgent, but I would polish the text.

libs/qec/lib/decoders/sliding_window.cpp
kvmto: why a circular buffer? :)

libs/qec/lib/decoders/sliding_window.cpp
kvmto: maybe protected?

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -0,0 +1,432 @@
...
+    window_proc_times_arr[7] =
+        std::chrono::duration<double>(t7 - t6).count() * 1000;
+    CUDAQ_INFO("Window {} time: {:.3f} ms (0:{:.3f}ms 1:{:.3f}ms 2:{:.3f}ms "
+               "3:{:.3f}ms 4:{:.3f}ms 5:{:.3f}ms 6:{:.3f}ms 7:{:.3f}ms)",
+               w, window_proc_times[w], window_proc_times_arr[0],
+               window_proc_times_arr[1], window_proc_times_arr[2],
+               window_proc_times_arr[3], window_proc_times_arr[4],
+               window_proc_times_arr[5], window_proc_times_arr[6],
+               window_proc_times_arr[7]);
+  }
```
kvmto, line 537: We probably need a construct for timings to reuse anywhere we need timings.
Also the function requires some refactoring there is lots of duplicate code.

libs/qec/unittests/test_decoders.cpp
```diff
@@ -289,6 +290,187 @@ TEST(SteaneLutDecoder, checkAPI) {
...
+          EXPECT_EQ(a.at({r, c}), b.at({r, c}))
+              << "a.at({" << r << ", " << c << "}) = " << a.at({r, c})
+              << ", b.at({" << r << ", " << c << "}) = " << b.at({r, c})
+              << "\n";
+      }
+    }
+  }
+}
+
+void SlidingWindowDecoderTest(bool run_batched) {
```
kvmto: Can we split these large tests into smaller ones?
Also, could we add performance tests to track execution time?
This could help detect potential performance regressions in the future.

libs/qec/unittests/test_qec.cpp
```diff
@@ -1420,6 +1421,180 @@ TEST(PCMUtilsTester, checkShufflePCMColumns) {
   check_pcm_equality(pcm_permuted_and_sorted, pcm);
 }
 
+TEST(PCMUtilsTester, SlidingWindowDecoderTest) {
```
kvmto: Is there a reason why we can't split this one?

kvmto (commented):
Really good job! I agree with Melody's comments

libs/qec/lib/decoders/sliding_window.cpp
melody-ren: Using a circular buffer, you only need to update the tracking indices when you shift the rolling window by one round, which reduces the amount of data we need to copy (although you still need to copy in the new syndrome data). 

The other thing is that under the assumption of "streaming mode takes in one round per call," the amount to shift/update the tracking indices is a known quantity each call (it's exactly one round), which makes the implementation manageable. You can still do it if each call receives a different number of 
rounds, but then you need to be careful about updating the indices.

wsttiger: it seems that circular buffer would be a more elegant solution

docs/sphinx/api/qec/sliding_window_api.rst
bmhowe23: Do you have any concrete suggested changes? If so, it would be helpful to provide them. Otherwise, I'm not sure how to respond to this comment.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: A user already cannot use this function because it is only exposed in a .cpp file (not available in a .h file), and users only use the base class `decoder` instead. In other words, a user cannot see the `class sliding_window` object due to the `get_decoder` abstraction.

That being said - I can certainly add a comment saying that.

libs/qec/unittests/test_decoders.cpp
bmhowe23: I don't think this function can be broken down into smaller tests. This is a helper function for individual tests; i.e., this function exists to prevent duplicate code for the actual tests, and I think it is important to keep it.

More specifically, this test simply decodes syndromes using a) the global decoder, and b) the windowed decoder, and it makes sure they agree. I will add a comment saying that.

Regarding timing, I think that would best be done in a separate PR. It is indeed on the roadmap to add benchmarking tests. (Currently on the Jira board.)

libs/qec/unittests/test_qec.cpp
bmhowe23: I will remove this test. The tests in test_decoders.cpp are more complete tests anyway. (This one doesn't even truly exercise the "sliding_window" decoder.)

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: `step_size` is of type `size_t`, so it already cannot be negative.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: I think that having the timing enabled by default is a key performance parameter of a sliding window decoder, especially because we are timing things that end users cannot time. For example, in `decode_batch`, a regular user would have no visibility into how long it is taking to decode each window for the syndromes. For that reason, while we *could* wrap them in `#ifndef`s, I think that would clutter up the code (making it more difficult to read), and our CI would not actually be validating that we can build with that new flag, so I wouldn't really trust that the flags were correct anyway.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Using a circular buffer makes the code a bit more complicated to read and maintain in my opinion, but I agree that it should, in principle, yield better performance. For that reason, I went ahead and did it. a4932bc.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Comments added in 5d93610 and additional error checks added in 14ff785

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Changed to an enum, which should make it easier to reason about, here: 6f00c71

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Updated in 2bd9d3e

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Added in 14ff785

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Good catch! Fixed in 86a98c3.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Updated in 7351093

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Yes, it means we are advancing by one round. Do you have a recommended wording change here?

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Yes, that should be fixed in 86a98c3.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Good catch! Fixed in 86a98c3, and new tests added in 59181e1.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: One is typically dealing with vectors and the other is dealing with vectors of vectors, so the code is often subtly different. That being said - the point of the `decode_window` helper function is to be used by both `decode` and `decode_batched`, thereby minimizing truly duplicated code where possible.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Comments added in 5d93610.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: As mentioned above `decode_window` is a function that is used by both `decode` and `decode_batched`. I am not sure which code you are saying is duplicated. Are you saying we have duplicate code inside `decode_window`? If so, where?

libs/qec/lib/decoders/sliding_window.cpp
melody-ren: Oh that's right.

libs/qec/lib/decoders/sliding_window.cpp
melody-ren: That's a fair point.

libs/qec/lib/decoders/sliding_window.cpp
melody-ren: I see. I must admit that I didn't read decode_batch too closely so I missed these subtle differences

libs/qec/lib/decoders/sliding_window.cpp
melody-ren: How about just "Window is full, sliding the window **by one round**"?

docs/sphinx/api/qec/sliding_window_api.rst
kvmto: I can't make the suggestion for some reason.

Here, what I would write:

```
   The Sliding Window Decoder is a wrapper around a standard decoder that introduces two key differences:

1. **Sliding Window Decoding**
The decoding process is performed incrementally, one window at a time. The window size is specified by the user. This allows decoding to begin before all syndromes have been received, reducing overall latency in multi-round QEC codes.

2. **Partial Syndrome Support**
Unlike standard decoders, the :code:`decode` function (and its variants, e.g., :code:`decode_batch`) can accept partial syndromes. If partial syndromes are provided, the return vector will be empty, the decoder will not complete the processing and remain in an intermediate state, awaiting future syndromes. The return vector is only non-empty once enough data has been provided to match the original syndrome size (calculated from the Parity Check Matrix).

Sliding window decoders are of advantage in QEC codes subject to circuit-level noise
across multiple syndrome extraction rounds. These decoders permit syndrome processing
to begin before the complete syndrome measurement sequence is obtained, potentially
reducing the overall decoding latency. However, this approach introduces a trade-off:
the reduction in latency typically comes at the cost of increased logical error rates.
Therefore, the viability of sliding window decoding depends critically on the specific
code parameters, noise model, and latency requirements of the system under consideration.

Sliding window decoding imposes only a single structural constraint on the parity check
matrices: each syndrome extraction round must produce a constant number of syndrome
measurements. Notably, the decoder makes no assumptions about temporal correlations
or periodicity in the underlying noise process.

**References**: `Toward Low-latency Iterative Decoding of QLDPC Codes Under Circuit-Level Noise <https://arxiv.org/abs/2403.18901>`_

```

libs/qec/lib/decoders/sliding_window.cpp
kvmto: Ah! I just saw that the code duplication was addressed above, sorry. Just as a thought, I see lots of similarities between decode and decode batch, I think something can be done there, unless there is some design philosophy underneath it. Everything is a decode_batch at the of the day.

For the timings, can we use something like this? I think it would simplify things, and could be reused in the future.
I took some inspiration from CUDAQ.

```
class ScopedTimer {
    std::chrono::high_resolution_clock::time_point start;
    double& target;
    std::string message;
    enum class LogLevel { NONE, DEBUG, INFO };
    LogLevel log_level;
    
public:
    ScopedTimer(double& target, 
                const std::string& msg = "", 
                LogLevel level = LogLevel::NONE) 
        : target(target), message(msg), log_level(level),
          start(std::chrono::high_resolution_clock::now()) {}
    
    ~ScopedTimer() {
        auto end = std::chrono::high_resolution_clock::now();
        double elapsed = std::chrono::duration<double>(end - start).count() * 1000;
        target += elapsed;
        
        if (!message.empty()) {
            switch (log_level) {
                case LogLevel::DEBUG:
                    CUDAQ_DBG("{}: {:.3f} ms", message, elapsed);
                    break;
                case LogLevel::INFO:
                    CUDAQ_INFO("{}: {:.3f} ms", message, elapsed);
                    break;
                case LogLevel::NONE:
                default:
                    break;
            }
        }
    }
};

```

kvmto (commented):
I thought I finished the review yesterday, sorry for the delay...

melody-ren:
The updates look good to me.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: I agree that it would be good to add something like a `ScopedTimer` at some point, but I'd rather not use that for this particular case as I want all the relevant times included in a single message (easier to grep and post-process). Additionally, the forced use of additional scopes in these functions would complicate some of the current usages because it would make other variables go out of scope. I'd rather not make changes for that at the moment, and we can just save an idea like this for a future PR where it is a better fit.

docs/sphinx/api/qec/sliding_window_api.rst
bmhowe23: OK, I took most of those suggestions and put them in 0addcdb.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: Updated in d48cddc

docs/sphinx/api/qec/sliding_window_api.rst
bmhowe23: #313

PR #199, opened 2025-06-30
[docs] Update circuit-level noise example to use new DEM feature

### LINKED ISSUE #196: Update circuit level noise example once #147 is merged
Not necessary for this PR, but we should update this example to use the circuit-level DEM.

_Originally posted by @justinlietz in https://github.com/NVIDIA/cudaqx/pull/147#discussion_r2173544761_

bmhowe23:
Resolves #196

### DIFF
--- a/docs/sphinx/examples/qec/python/circuit_level_noise.py
+++ b/docs/sphinx/examples/qec/python/circuit_level_noise.py
@@ -12,26 +12,21 @@ import cudaq_qec as qec
 
 # Get a QEC code
 cudaq.set_target("stim")
-steane = qec.get_code("steane")
+distance = 5
+surface_code = qec.get_code("surface_code", distance=distance)
 
-# Get the parity check matrix of a code
-# Can get the full code, or for CSS codes
-# just the X or Z component
-H = steane.get_parity()
-print(f"H:\n{H}")
-observables = steane.get_pauli_observables_matrix()
-Lz = steane.get_observables_z()
-print(f"observables:\n{observables}")
+# Get the Z observables.
+Lz = surface_code.get_observables_z()
 print(f"Lz:\n{Lz}")
 
-nShots = 3
-nRounds = 4
+nShots = 1000
+nRounds = distance
 
 # Uncomment for repeatability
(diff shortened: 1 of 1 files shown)


docs/sphinx/examples/qec/python/circuit_level_noise.py
```diff
@@ -12,26 +12,21 @@
 
 # Get a QEC code
 cudaq.set_target("stim")
-steane = qec.get_code("steane")
+distance = 5
+surface_code = qec.get_code("surface_code", distance=distance)
```
melody-ren, line 16: Just wondering: how come this is switched to surface code?

docs/sphinx/examples/qec/python/circuit_level_noise.py
```diff
@@ -40,21 +35,23 @@
 # our expected measurement in this state is 0
 expected_value = 0
 
+# Get the detector error model for this circuit.
+dem = qec.z_dem_from_memory_circuit(surface_code, statePrep, nRounds, noise)
```
melody-ren, line 39: Might want to print dem's error matrix

melody-ren (approved):

melody-ren:
Looks good to me. Minor questions here and there. 

Not related to this PR: but maybe this example, `repetition_code_fine_grain_noise.py`, should either be retired or be updated to mention that users can now use the PCM generator instead of manually creating a PCM

docs/sphinx/examples/qec/python/circuit_level_noise.py
bmhowe23: I switched because surface code has a configurable distance (more knobs to play with if people want).

docs/sphinx/examples/qec/python/circuit_level_noise.py
bmhowe23: The DEM is 120 x 1693, so I think that is too big to print.
If I shrunk the distance down to 3, the PCM would be 24 x 231, which is also probably too big to print (and be readable).

PR #206, opened 2025-07-08
Add check for tn decoder import

melody-ren:


libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -308,7 +308,29 @@ void bindDecoder(py::module &mod) {
         cudaqx::tensor<uint8_t> tensor_H(shape);
         tensor_H.borrow(static_cast<uint8_t *>(buf.ptr), shape);
 
-        return get_decoder(name, tensor_H, hetMapFromKwargs(options));
+        try {
```
bmhowe23: I think some of this could be simplified if you just check the return results from line 280 when `name == "tensor_network_decoder"`. Basically, if `PyDecoderRegistry::contains("tensor_network_decoder")` is `false` , it probably means that dependencies are missing.

melody-ren: I agree that this can be checked at line 280. My original intent was to not skip the C++ fallback at line 311, and use the `try` block to make the "unknown decoder" error message a bit more informative. If you prefer to keep it minimal, I can trim it down and check for tn decoder on line 280. Let me know which you prefer.

bmhowe23: Parsing exception error messages for magic strings (like "Unknown decoder: ") is a bit dangerous because they might change in the future, so I'd rather avoid string matching here. If we want to make the "Unknown decoder" exception more descriptive, I think we should do that in the location where the exception is being thrown.

melody-ren: That's a good point. Will update

melody-ren:
Wheel tests passed: https://github.com/NVIDIA/cudaqx/actions/runs/16207820632

bmhowe23 (approved):
Thanks, Melody

PR #208, opened 2025-07-09
Bump CUDA-Q commit (with underlying cuQuantum update)

bmhowe23:


docker/build_env/cudaqx.dev.Dockerfile
```diff
@@ -6,7 +6,8 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-FROM ghcr.io/nvidia/cuda-quantum-devdeps:ext-cu12.0-gcc11-main
+ARG base_image=ghcr.io/nvidia/cuda-quantum-devdeps:ext-amd64-cu12.0-gcc11-main
```
bmhowe23, line 9: Note to self: I would actually like to back this change out or else I need to update build_dev.yaml.

scripts/ci/test_wheels.sh
```diff
@@ -40,7 +40,6 @@ fi
 # QEC library
 # ======================================
 
-${python} -m pip install quimb opt_einsum torch
 # Install QEC library with tensor network decoder
 wheel_file=$(ls /wheels/cudaq_qec-*-cp${python_version_no_dot}-cp${python_version_no_dot}-*.whl)
 ${python} -m pip install "${wheel_file}[tn_decoder]"
```
melody-ren: This in theory should also install the dependencies listed in the project.toml file. But when I tried, those weren't installed and hence why they were installed on line 42. Just FYI. Maybe it was just a fluke back then.

bmhowe23: Thanks for the heads up. Looks like it's working now, afaict (at least for Python >= 3.11): https://github.com/NVIDIA/cudaqx/actions/runs/16204442561/job/45752705293

scripts/ci/test_wheels.sh
```diff
@@ -40,10 +40,16 @@ fi
...
+# If Python version is 3.10, then install without tensor network decoder.
+# Otherwise, install with the tensor network decoder.
+if [ $python_version == "3.10" ]; then
+  echo "Installing QEC library without tensor network decoder"
+  ${python} -m pip install "${wheel_file}"
+else
+  echo "Installing QEC library with tensor network decoder"
+  ${python} -m pip install "${wheel_file}[tn_decoder]"
+fi
 ${python} -m pytest -v -s libs/qec/python/tests/
```
melody-ren: You might also need to do a conditional `--ignore` on the `pytest` line when tn_decoder is not installed. Example, see #206

bmhowe23: Thanks. Updated in https://github.com/NVIDIA/cudaqx/pull/208/commits/2c633e677fe69985eb1390d1e1f557609f263e9e

melody-ren (approved):

Issue #200, opened 2025-07-02
Update repetition_code_fine_grain_noise.py example to additionally demonstrate how to use new PCM generation capability


bmhowe23:
> Looks good to me. Minor questions here and there.
> 
> Not related to this PR: but maybe this example, `repetition_code_fine_grain_noise.py`, should either be retired or be updated to mention that users can now use the PCM generator instead of manually creating a PCM

Good point. I like the idea keeping the manual way but updating it to show users how to achieve the same goal (or at least very similar goal) using the new PCM generator. I'll create a new issue for that.

_Originally posted by @bmhowe23 in https://github.com/NVIDIA/cudaqx/issues/199#issuecomment-3028887484_

PR #216, opened 2025-07-16
Replace PCM generation example with one using new DEM API

kvmto:
This PR updates the example script to demonstrate how to use the new detector error model (DEM)-based parity check matrix (PCM) generation capabilities in cudaq_qec.

The file repetition_code_fine_grain_noise.py has been renamed to repetition_code_pcm.py for clarity.

The example now uses qec.dem_from_memory_circuit() and qec.z_dem_from_memory_circuit() to extract the detector error matrix and observable flips matrix directly from a memory circuit simulation.

The logic and simulation loop remain the same but are now backed by the new PCM pipeline, which improves maintainability and reflects the latest decoding workflow.

Resolves: #200


kvmto:
This is just a proposal for now, not meant to work. Is this what you needed?

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -91,28 +91,127 @@ def three_qubit_repetition_code():
...
+        return 2
+
+    def get_num_ancilla_z_qubits(self):
+        return 2
+
+    def get_num_ancilla_qubits(self):
+        return 4
+
+# Create an instance of the three-qubit repetition code
+three_qubit_repetition_code = qec.get_code("3-qubit-repetition-code")
```
bmhowe23: Can we use `qec.get_code('repetition', distance=3)` here? Ideally, I think we should be able to use that instead of lines 117-208, and if not, I'd like to better understand why not.

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -132,6 +231,14 @@ def three_qubit_repetition_code():
 Hz = [[1, 1, 0], [0, 1, 1]]  # Parity check matrix for 1 round
 H_pcm = get_circuit_level_pcm(3, nRounds, Hz)
 
+# in alternative to the above, you can also use the following fucntion.
+# This will give you the parity check matrix for the repetition code
+# after the nRounds of the repetition code
+# It is useful for more complex codes where the parity check matrix is not trivial.
+
+dem_rep = qec.z_dem_from_memory_circuit(three_qubit_repetition_code, statePrep, nRounds, noise_model)
```
bmhowe23: If this does not work (and if I understand our previous conversations correctly, then it indeed does not work), then I think that it should be fixed as part of this PR.

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
```diff
@@ -75,7 +75,7 @@ def get_circuit_level_pcm(distance, nRounds, Hz):
 
     return H_pcm
 
-
+"""
```
melody-ren: Instead of commenting out a large chunk of the original file, it might be cleaner to create a new `.py` for the repetition code example using pcm

kvmto: Yes, the comment is just to make clearer what I was trying to change. My doubt was more about the comment that Ben made above. I was wondering if we want to show the user how to make custom codes with fine grained level of control or if we want to make things easier.

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
kvmto: I think the goal of the example is to show how to customize a qec code in python and use for a memory experiment. Using the one from the library is great, but I am afraid it would not show what the user needs to find here.

melody-ren: I would suggest that we leave this example as is (which already demonstrates how to handcraft the decoder and the PCM), and start a new example in a new `.py`. Something like **repetition_code_pcm.py**? 

In the new example, you can use `get_code` and showcase how to use the pcm generator to produce the dem matrix for decoding (in contrast to the manual method introduced in the original example).

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
melody-ren: Resolved in another comment. Closing now.

kvmto:
I discovered together with Justin issues on the `sample_memory_circuit(...)` function. The syndromes do not match the data qubits generated. This has to be solved before merging the example.
@bmhowe23 @melody-ren Ideas are welcome :)

melody-ren:
> I discovered together with Justin issues on the `sample_memory_circuit(...)` function. The syndromes do not match the data qubits generated. This has to be solved before merging the example. @bmhowe23 @melody-ren Ideas are welcome :)

Sorry, I'm a bit confused here... You mean that as of now, the `sample_memory_circuit(...)` doesn't produce correct output and is yet to be fixed, ya? Is this for the repetition code only? And is this only related to when you're using the dem?

docs/sphinx/examples_rst/qec/decoders.rst
```diff
@@ -9,65 +9,40 @@ The relationship between errors and syndromes is captured mathematically by the
 stabilizer measurement, while each column represents a possible error. When we multiply an error pattern by this matrix, we get the syndrome 
 that would result from those errors.
 
-Creating a Multi-Round Parity Check Matrix
```
melody-ren, line 12: Maybe I missed something. But I thought the intent of issue #200 was to keep the handcrafted example as is, and introduce a *new* example showing how to use the pcm generator?

bmhowe23:
> I discovered together with Justin issues on the `sample_memory_circuit(...)` function. The syndromes do not match the data qubits generated. This has to be solved before merging the example.

Could you please elaborate on this a bit more? I don't understand "The syndromes do not match the data qubits generated."

docs/sphinx/examples_rst/qec/decoders.rst
kvmto: I had to start in reversed order. I opened issue https://github.com/NVIDIA/cudaqx/issues/218
which should be the handcrafted version. I am going to tackle it as soon as this gets merged.

kvmto:
> > I discovered together with Justin issues on the `sample_memory_circuit(...)` function. The syndromes do not match the data qubits generated. This has to be solved before merging the example. @bmhowe23 @melody-ren Ideas are welcome :)
> 
> Sorry, I'm a bit confused here... You mean that as of now, the `sample_memory_circuit(...)` doesn't produce correct output and is yet to be fixed, ya? Is this for the repetition code only? And is this only related to when you're using the dem?


This function generates data qubit measurements and, for each, the corresponding ancilla (syndrome) measurements across multiple rounds.
However, using the decoder currently increases the error rate, which suggests something is wrong.
To debug, we examined how syndromes relate to specific data qubit outputs.
For instance, in one case the final data qubit measurements were 0 1 1, which implies a bit-flip error between qubits 0 and 1.
But the recorded syndromes across 4 rounds were all 0 0, suggesting no error occurred.
We'd expect syndromes like 1 0, 0 0, 0 0, 1 0 in that case, reflecting detection of boundary changes.
This mismatch implies that either the syndromes are not being generated correctly or they are not aligned with the data measurements, which likely explains why decoding is ineffective.

I hope this is clear now @bmhowe23 and @melody-ren

bmhowe23:
> > > I discovered together with Justin issues on the `sample_memory_circuit(...)` function. The syndromes do not match the data qubits generated. This has to be solved before merging the example. @bmhowe23 @melody-ren Ideas are welcome :)
> > 
> > 
> > Sorry, I'm a bit confused here... You mean that as of now, the `sample_memory_circuit(...)` doesn't produce correct output and is yet to be fixed, ya? Is this for the repetition code only? And is this only related to when you're using the dem?
> 
> This function generates data qubit measurements and, for each, the corresponding ancilla (syndrome) measurements across multiple rounds. However, using the decoder currently increases the error rate, which suggests something is wrong. To debug, we examined how syndromes relate to specific data qubit outputs. For instance, in one case the final data qubit measurements were 0 1 1, which implies a bit-flip error between qubits 0 and 1. But the recorded syndromes across 4 rounds were all 0 0, suggesting no error occurred. We'd expect syndromes like 1 0, 0 0, 0 0, 1 0 in that case, reflecting detection of boundary changes. This mismatch implies that either the syndromes are not being generated correctly or they are not aligned with the data measurements, which likely explains why decoding is ineffective.
> 
> I hope this is clear now @bmhowe23 and @melody-ren

Have you tried running with CUDAQ_LOG_LEVEL=info to see [this line](https://github.com/NVIDIA/cudaqx/blob/1745f9007ee547b2934fe74cdf02b3290d29d966/libs/qec/lib/detector_error_model.cpp#L127) is being logged anywhere? If you see that line, it suggests (but does not guarantee) there may be something wrong with the QEC code (i.e. bad CNOT ordering, or other things).

docs/sphinx/examples/qec/python/repetition_code_pcm.py
```diff
@@ -0,0 +1,138 @@
...
+    # Construct multi-round syndrome vector
+    multi_round_syndromes = []
+    previous_syndrome = [0] * len(syndromes[0])
+    for j in range(nRounds):
+        syndrome_i_j = syndromes[i + j]
+        print(f"syndrome: {syndrome_i_j}")
+
+        # Compute syndrome difference (flip tracking)
+        multi_round_syndromes += list(
+            np.bitwise_xor(previous_syndrome, syndrome_i_j))
```
bmhowe23: The `sample_memory_circuit` function does the xor-ing for you (see https://github.com/NVIDIA/cudaqx/blob/1745f9007ee547b2934fe74cdf02b3290d29d966/libs/qec/lib/experiments.cpp#L172-L173), so you shouldn't have this here. That is mentioned in https://nvidia.github.io/cudaqx/examples_rst/qec/circuit_level_noise.html (when it says `During a shot, each stabilizer round’s syndrome is xor’d against the preceding syndrome, ....`). Perhaps it needs to be added into other docs sections as well.

docs/sphinx/examples/qec/python/repetition_code_pcm.py
```diff
@@ -0,0 +1,138 @@
...
+    error_prediction = np.array(result, dtype=np.uint8)
+    print(f"error_prediction: {error_prediction}")
+
+    # Apply prediction to infer logical flip
+    predicted_observable_flip = Lz_observables_flips_matrix @ error_prediction % 2
+    print(f"predicted_observable_flip: {predicted_observable_flip}")
+
+    # Check what the circuit actually measured
+    measured_obserable = logical_single_round @ data_i % 2
+    print(f"measured_obserable: {measured_obserable}")
```
bmhowe23:
```suggestion
    measured_observable = logical_single_round @ data_i % 2
    print(f"measured_observable: {measured_observable}")
```

docs/sphinx/examples/qec/python/repetition_code_pcm.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Count error without decoding
+    if (measured_obserable != expected_value):
```
bmhowe23:
```suggestion
    if (measured_observable != expected_value):
```

docs/sphinx/examples/qec/python/repetition_code_pcm.py
```diff
@@ -0,0 +1,138 @@
...
+        nLogicalErrorsWithoutDecoding += 1
+
+    # Correct prediction by combining predicted flip with actual measurement
+    predicted_observable = predicted_observable_flip ^ measured_obserable
```
bmhowe23:
```suggestion
    predicted_observable = predicted_observable_flip ^ measured_observable
```

bmhowe23 (commented):
I think one of these comments explains the reason for the issue you were seeing.

docs/sphinx/examples/qec/python/repetition_code_pcm.py
bmhowe23: Also, while looking into this, I coincidentally found an inconsistency in our docs vs implementation, going back to our original public commit. More specifically - I think https://github.com/NVIDIA/cudaqx/blob/1745f9007ee547b2934fe74cdf02b3290d29d966/docs/sphinx/components/qec/introduction.rst?plain=1#L979 is inconsistent with the `sample_memory_circuit` implementation. Our current implementation returns `(num_shots * num_rounds, syndrome_size)`, but the docs say `(num_shots, num_rounds * syndrome_size)`. I'll open a new issue for this; just noting it here for your awareness.

Edit: created #221

libs/qec/lib/experiments.cpp
```diff
@@ -357,12 +365,18 @@ cudaq::qec::detector_error_model dem_from_memory_circuit(
...
+  // printf("getting msm_obs : \n");
+  // msm_obs.dump_bits();
+
   // Uncomment print statements for debugging:
   // printf("dem.detector_error_matrix Before canonicalization:\n");
   // dem.detector_error_matrix.dump_bits();
   // printf("dem.observables_flips_matrix Before canonicalization:\n");
   // dem.observables_flips_matrix.dump_bits();
-  dem.canonicalize_for_rounds(numReturnSynPerRound);
+  // dem.canonicalize_for_rounds(numReturnSynPerRound);
```
bmhowe23: Did you mean to check in this change? (And the above debug code?) I think this could be the cause for one of the test failures.

kvmto: Yes, I was looking at the diff!

kvmto: Thanks!! :)

kvmto:
if okay I would merge it:)

docs/sphinx/examples/qec/python/repetition_code_pcm.py
```diff
@@ -0,0 +1,131 @@
...
+                                            statePrep, nShots, nRounds,
+                                            noise_model)
+
+syndromes = syndromes.reshape((nShots, nRounds, -1))
+
+# Now flatten to two dimensions again
+syndromes = syndromes.reshape((nShots, -1))
+
+# Initialize Pauli frame (Z only for repetition code)
+pauli_frame = np.array([0, 0], dtype=np.uint8)
```
bmhowe23: If this is unused, can it be removed?

docs/sphinx/examples/qec/python/repetition_code_pcm.py
```diff
@@ -0,0 +1,131 @@
...
+expected_value = 1
+
+# Counters for statistics
+nLogicalErrorsWithoutDecoding = 0
+nLogicalErrorsWDecoding = 0
+nCorrections = 0
+
+# === Loop over shots ===
+for i in range(nShots):
+    # print(f"shot: {i}")
```
bmhowe23: For these commented out print statements, could you please do it like we do it here: https://github.com/NVIDIA/cudaqx/blob/1e7aa8c14518aa6a0b8d527186c944f13c395c32/docs/sphinx/examples/qec/python/circuit_level_noise.py#L42, where we set a `verbose` flag depending on how many shots are run?

libs/qec/lib/experiments.cpp
```diff
@@ -292,21 +294,27 @@ cudaq::qec::detector_error_model dem_from_memory_circuit(
...
+    numReturnSynPerRound = numXStabs;
+  } else if (!keep_x_stabilizers && keep_z_stabilizers) {
+    numReturnSynPerRound = numZStabs;
+  }
+
   // If we are returning only x-stabilizers, we need to offset the syndrome
   // indices of mzTable by numSyndromesPerRound / 2.
-  auto offset =
-      keep_x_stabilizers && !keep_z_stabilizers ? numSyndromesPerRound / 2 : 0;
+  auto offset = keep_x_stabilizers && !keep_z_stabilizers ? numXStabs : 0;
```
bmhowe23: If we're skipping over Z stabilizers, should it be this instead?
```suggestion
  auto offset = keep_x_stabilizers && !keep_z_stabilizers ? numZStabs : 0;
```

libs/qec/lib/experiments.cpp
kvmto: Solved in f7e0ec960debacfce38de421f04550e463539c9f.

docs/sphinx/examples/qec/python/repetition_code_fine_grain_noise.py
kvmto: Solved in 2ed9dba4d4aa4e80f6a0bdaffa149b03b57a9d99.

docs/sphinx/examples/qec/python/repetition_code_pcm.py
kvmto: I will fix this part of the docs in a separate PR for clarity.

bmhowe23:
> This PR enhances the example by keeping the manual PCM creation while also showing how to achieve the same result using the new PCM generator, as discussed in the issue.

Could you please update the textual description for this PR? In particular, I don't think statement is accurate anymore. Let me know if I'm misinterpreting that, though.

docs/sphinx/examples/qec/python/repetition_code_pcm.py
kvmto: solved in 9363005a988c5285e4f56ee6eac671abd90aee0e.

docs/sphinx/examples/qec/python/repetition_code_pcm.py
kvmto: solved in 9363005a988c5285e4f56ee6eac671abd90aee0e.

bmhowe23:
> Nit here: I think this PR partially resolves #200 and is more like it's resolving #218 instead. Can we please update the PR description to reflect this?

I don't think this is resolving #218. I think resolving #218 involves creating an example with a custom `@qec.code` class, which this PR does not do. That being said - this is indeed a bit confusing. I think it is resolving #200, but contrary to what #200 says, this is eliminating the "manual way" in favor of the new way. #218 will add a new new way back.

Perhaps the title of this PR should be renamed to "Replace PCM generation example with one using new DEM API".

bmhowe23:
> > Nit here: I think this PR partially resolves #200 and is more like it's resolving #218 instead. Can we please update the PR description to reflect this?
> 
> I don't think this is resolving #218. I think resolving #218 involves creating an example with a custom `@qec.code` class, which this PR does not do. That being said - this is indeed a bit confusing. I think it is resolving #200, but contrary to what #200 says, this is eliminating the "manual way" in favor of the new way. #218 will add a new new way back.
> 
> Perhaps the title of this PR should be renamed to "Replace PCM generation example with one using new DEM API".

Kevin, I will go ahead and a) approve this PR, and b) rename this PR just so this comment isn't entirely missed. (I see you already updated the description, which is good.) If you feel strongly enough to rename it to something more clear after my rename, feel free.

PR #222, opened 2025-07-21
Add python 3.13 support

melody-ren:


bmhowe23 (commented):
Thanks, Melody. Could you please also update https://github.com/NVIDIA/cudaqx/blob/1745f9007ee547b2934fe74cdf02b3290d29d966/.github/workflows/build_dev.yaml#L87? No need to re-run the "Build wheels" job as a result of this change, but it would be good to include it in this PR.

bmhowe23 (approved):
+1

PR #224, opened 2025-07-21
Fix decoder plugin not throwing the missing dependecy err msg issue

melody-ren:
The expected behaviour for the optional decoder plugin is that if any of the python dependencies is missing and the user attempts to use the optional decoder, a runtime error message should be thrown. Adding "ImportError" to catch the missing dependency issue and therefore trigger the expected runtime error message.

### DIFF
--- a/.github/actions/get-cudaq-wheels/action.yaml
+++ b/.github/actions/get-cudaq-wheels/action.yaml
@@ -143,6 +143,17 @@ runs:
           python_version=3.12
         outputs: /cudaq-wheels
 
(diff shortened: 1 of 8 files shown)


bmhowe23 (commented):
How difficult would it be to add a negative test to verify that this error message is thrown? Given that we always expect that this error message should be thrown for Python 3.10 environments, it seems like we could add a Python 3.10-only test for that. Thoughts?

bmhowe23 (approved):
Thanks!

PR #226, opened 2025-07-22
Add tests for qec lib

caiyunh:
Add some test cases to increase the code coverage rate of qec lib to 100%.


libs/core/include/cuda-qx/core/kwargs_utils.h
```diff
@@ -67,9 +67,6 @@ inline heterogeneous_map hetMapFromKwargs(const py::kwargs &kwargs) {
       } else {
         throw std::runtime_error("Unsupported array data type in kwargs.");
       }
-    } else if (py::isinstance<py::dict>(value)) {
```
melody-ren, line 70: How come we are removing this?

melody-ren: Ah I see. Line 47 already covered this case. Case closed.

libs/qec/unittests/test_decoders.cpp
```diff
@@ -221,6 +221,75 @@ TEST(SteaneLutDecoder, checkAPI) {
...
+  // Set syndrome to 101
+  syndrome_dt[0] = 1.0;
+  syndrome_dt[2] = 1.0;
+  auto result_dt = d4->decode(syndrome_dt);
+
+  // Verify opt_results contains decoding_time
+  ASSERT_TRUE(result_dt.opt_results.has_value());
+  ASSERT_TRUE(result_dt.opt_results->contains("decoding_time"));
+  // Verify the decoding_time value is the expected 0.0
+  ASSERT_EQ(result_dt.opt_results->get<double>("decoding_time"), 0.0);
```
bmhowe23, line 292: This test is in accordance with our current implementation (good), but I would like to see us report a correct time here, even for just the "single_error_lut" decoder. I will create a new issue for this.

Edit: https://github.com/NVIDIA/cudaqx/issues/232 created.

libs/qec/unittests/test_decoders.cpp
```diff
@@ -280,3 +396,133 @@ TEST(DecoderResultTest, EqualityOperator) {
...
+
+  // Test that we can check if extensions are registered in the single-parameter
+  // registry
+  auto registered_single = cudaqx::extension_point<
+      cudaq::qec::decoder, const cudaqx::tensor<uint8_t> &>::get_registered();
+
+  // The registry should exist (even if empty), proving line 18 instantiation
+  // works This test passes if no exceptions are thrown, proving the
+  // single-parameter registry is instantiated
+  EXPECT_TRUE(true);
```
bmhowe23: I think this line is unnecessary.
```suggestion
```

caiyunh:
Thanks @bmhowe23 @melody-ren for reviewing. I have submitted the suggested changes.

PR #227, opened 2025-07-22
Fix issue 218: Add fine grained noise injection in a custom qec code

kvmto:


docs/sphinx/examples/qec/python/custom_repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,187 @@
...
+Lz_observables_flips_matrix = dem_rep.observables_flips_matrix
+
+# Sample noisy executions of the full memory circuit
+print("\n Sampling noisy memory circuit executions...")
+syndromes, data = qec.sample_memory_circuit(my_repetition_code, statePrep,
+                                            nShots, nRounds, noise_model)
+
+# Reshape syndromes to flatten rounds per shot
+syndromes = syndromes.reshape((nShots, nRounds, -1))
+syndromes = syndromes.reshape((nShots, -1))
```
bmhowe23: I think the first line is unnecessary.
```suggestion
syndromes = syndromes.reshape((nShots, -1))
```

libs/qec/python/bindings/py_code.cpp
```diff
@@ -77,13 +77,25 @@ class PyCodeHandle : public qec::code {
...
       throw std::runtime_error(
           "Invalid Python QEC Code. Must have self.operation_encodings = "
           "{...}. Please provide the CUDA-Q kernels for the operation "
           "encodings.");
 
+    if (py::hasattr(registeredCode, "pauli_observables")) {
+      auto obs_terms = registeredCode.attr("pauli_observables")
+                           .cast<std::vector<cudaq::spin_op_term>>();
+      m_pauli_observables.reserve(obs_terms.size());
+      for (auto &term : obs_terms)
```
bmhowe23: I suspect you need an rvalue reference here to make the move below work as intended. (I could be wrong, though?)
```suggestion
      for (auto &&term : obs_terms)
```

docs/sphinx/examples/qec/python/custom_repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def get_num_z_stabilizers(self):
+        return self.distance - 1
+
+
+# Instantiate the custom repetition code
+my_repetition_code = qec.get_code("custom_repetition_code", distance=distance)
+print(f"\n Created custom repetition code with distance {distance}.")
+
+# Show available codes (doesn't include Python-defined custom codes)
```
bmhowe23: Any reason to not just update `qec.get_available_codes()` to include Python? It seems confusing to add a new `qec.get_all_available_codes()` function.

docs/sphinx/examples/qec/python/custom_repetition_code_fine_grain_noise.py
```diff
@@ -0,0 +1,187 @@
...
+
+# Reshape syndromes to flatten rounds per shot
+syndromes = syndromes.reshape((nShots, nRounds, -1))
+syndromes = syndromes.reshape((nShots, -1))
+
+print(f"\n Showing first 5 of {nShots} sampled results:")
+for i in range(5):
+    print(
+        f"Shot {i+1:>2}: Logical measurement = {data[i]}, Syndromes = {syndromes[i]}"
+    )
```
bmhowe23: This should be updated to allow for nShots < 5 cases (in case the user changes it). As written, I think it would produce confusing output.

libs/qec/python/bindings/py_code.cpp
```diff
@@ -310,6 +330,17 @@ void bindCode(py::module &mod) {
   qecmod.def("get_available_codes", &get_available_codes,
              "Get a list of all available quantum error correction codes");
 
+  qecmod.def(
+      "get_all_available_codes",
```
bmhowe23: In reference to the other comment about `get_all_available_codes`, what I am essentially recommending is deleting lines 330-331 and replacing line 334 as shown here:
```suggestion
      "get_available_codes",
```

docs/sphinx/examples/qec/python/custom_repetition_code_fine_grain_noise.py
kvmto: I thought of that, but I decided to create a new one because it seemed like the original wasn’t meant to include Python codes, especially since adding them felt pretty straightforward and intuitive.  I figured having a new version might be helpful for users who want to confirm their code was successfully registered.

libs/qec/python/bindings/py_code.cpp
kvmto: solved in 93507243bba4aa67a4af3ecb4b10411192a90da2

docs/sphinx/examples/qec/python/custom_repetition_code_fine_grain_noise.py
kvmto: solved in 93507243bba4aa67a4af3ecb4b10411192a90da2

docs/sphinx/examples/qec/python/custom_repetition_code_fine_grain_noise.py
kvmto: solved in 93507243bba4aa67a4af3ecb4b10411192a90da2

bmhowe23 (approved):
LGTM. Thanks, Kevin.

PR #242, opened 2025-07-29
Move extension_point from cuda-qx to cudaq

amccaskey:
Requires https://github.com/NVIDIA/cuda-quantum/pull/3282

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "5b1000af08ad059c82af94f18782f06e0ecd12d3"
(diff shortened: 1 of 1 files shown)


.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "5b1000af08ad059c82af94f18782f06e0ecd12d3"
+    "ref": "887759d8152894b9ae4851d790724938eafdd149"
```
bmhowe23, line 4: Initial testing looks great, so nothing required for now, but just noting that we will want to revert this change (or change it to a commit that is actually on cuda-quantum `main` branch) before actually merging this CUDA-QX PR.

Issue #258, opened 2025-08-08
[DEM} canonicalize_for_rounds out-of-bounds access when iterating over column indices


kvmto:
#### Description  
In **`detector_error_model.cpp`**, the method `detector_error_model::canonicalize_for_rounds` can crash with an out-of-bounds access when processing `column_order`.  
If `column_order` contains an index greater than or equal to the size of `row_indices` or `error_rates`, the following lines access invalid memory:  

```cpp
auto &curr_row_indices = row_indices[column_index];
if (curr_row_indices.size() == 0 || error_rates[column_index] == 0)
    continue;

PR #261, opened 2025-08-13
Fix issue #258: added check and informative error message

kvmto:
This PR fixes issue #258 by adding a validation check on the dimensions of the dem and the error rates and providing a clear, informative error message when the condition is not met. This improves error handling and helps users quickly identify the cause of the problem when the construction is made by hand in a custom QEC example.


bmhowe23 (commented):
Thanks, Kevin. Could you please add a unit test that would exercise this new code? Either Python or C++ is needed, but not both.

bmhowe23 (approved):

PR #263, opened 2025-08-15
Use 'tensor-network-decoder' in pip install instructions

melody-ren:
The name`tensor_network_decoder` from the `.toml` file is normalized to be `tensor-network-decoder` in the wheel's METADATA. Newer version of `pip` is able to resolve this difference and `pip install cudaq-qec[tensor_network_decoder]` will work correctly. 

However, if you use an older version of `pip/setuptools`, you'll see an error such as "cudaq-qec 0.4.0 does not provide the extra 'tensor_network_decoder'". A safer alternative is to use the exact name as appeared in the METADATA (the normalized name with dash instead of underscore) such that even earlier versions of `pip` will find the package and install the dependencies as needed.

### DIFF
--- a/docs/sphinx/examples/qec/python/tensor_network_decoder.py
+++ b/docs/sphinx/examples/qec/python/tensor_network_decoder.py
@@ -16,17 +16,17 @@ if sys.version_info < (3, 11):
 
 # [Begin Documentation]
 """
-Example usage of tensor_network_decoder from cudaq_qec.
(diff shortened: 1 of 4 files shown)


bmhowe23 (approved):
Thanks, Melody. Next week, we may need to talk about whether it's necessary to do a follow-up release to update the instructions on the PyPI page.

PR #270, opened 2025-08-19
[docs] Update Building from Source section

bmhowe23:


docs/sphinx/quickstart/installation.rst
```diff
@@ -63,93 +63,8 @@ The container includes:
...
-        -DCMAKE_BUILD_TYPE=Release \
-        -DCMAKE_INSTALL_PREFIX=$HOME/.cudaqx
-
-4. Build and install:
-
-.. code-block:: bash
-
-    ninja install
-
-CMake Build Options
```
melody-ren, line 113: It would be nice to include these options in `Building.md` and add a note about how to include/exclude the python binding and tests. The time it takes to complete ctest can vary quite a bit depending on which libraries are built. Can address this in a follow up PR. 

Tracked by issue #271

PR #275, opened 2025-08-22
Add patch_wheel_metadata.sh script

bmhowe23:
We may - from time to time - need to update our PyPI wheels' METADATA files without going through a full new release. The script in this PR will help that process.

### DIFF
--- a/docs/sphinx/quickstart/installation.rst
+++ b/docs/sphinx/quickstart/installation.rst
@@ -27,6 +27,7 @@ install individual components:
 CUDA-QX provides optional pip-installable components:
 
 .. code-block:: bash
+
     # Install the Tensor Network Decoder from the QEC library
     pip install cudaq-qec[tensor-network-decoder]
 


melody-ren (approved):
Script works. Thanks.

Should the wheel names be changed to `cudaq-qec`/`cudaq-solvers` to match pypi though?

melody-ren:
> > Script works. Thanks.
> > Should the wheel names be changed to `cudaq-qec`/`cudaq-solvers` to match pypi though?
> 
> Do we have an example where the current names aren't working? I was thinking the only previous demonstration of failure was with optional dependency packages names.
> 
> In any case, I think the build process must be putting the underscores in the filenames. Our pyproject.toml file already has the dashes. (CUDA-Q has the same thing.)

Current names work. I just find it confusing, but maybe don't fix things that aren't broken

bmhowe23:
> Current names work. I just find it confusing, but maybe don't fix things that aren't broken

OK, good. I wouldn't even know where to change it to be honest. I think this might be a case of the Python spec dictating that because https://packaging.python.org/en/latest/specifications/binary-distribution-format/#binary-distribution-format says:
> The wheel filename is {distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl.

So the `-` is the delimiter between sections in the filename, so the build system probably changes them to `_` during the build.

PR #280, opened 2025-08-24
Add prune_cudaqx-dev_by_sha.sh helper script

bmhowe23:
This script must be manually run. We could add it to the CI at some point if we want to automate the cleanup, but since we only do this approximately once per week, manual is sufficient for now.

### DIFF
--- a/scripts/prune_cudaqx-dev_by_sha.sh
+++ b/scripts/prune_cudaqx-dev_by_sha.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # ============================================================================ #
 # Copyright (c) 2025 NVIDIA Corporation & Affiliates.                          #
 # All rights reserved.                                                         #
@@ -6,7 +7,18 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-#!/usr/bin/env bash
+# NOTE: This script is intended to be run by CUDA-QX maintainers, not regular
+# users.
+
+# This script is used to prune the old cudaqx-dev packages from the GitHub
+# Packages.  In general, the .github/workflows/build_dev.yaml creates new images
+# whenever the CUDA-Q commit is updated, so this script will need to be
+# periodically run to keep the number of packages in the GitHub Packages
+# manageable.
+
+# The script requires a TOKEN environment variable to be set by the user. The
+# token must have read:packages and delete:packages scopes.
+
 set -euo pipefail
 
 JSON_FILE="versions.json"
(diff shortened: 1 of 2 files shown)


scripts/prune_cudaqx-dev_by_sha.sh
```diff
@@ -0,0 +1,131 @@
+# ============================================================================ #
+# Copyright (c) 2025 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
```
melody-ren, line 9: Might want to add a mini README here to describe "when/what/why to use this script for"

scripts/prune_cudaqx-dev_by_sha.sh
```diff
@@ -0,0 +1,131 @@
...
+  case "$1" in
+    --force) FORCE=true; shift ;;
+    --skip-cleanup) SKIP_CLEANUP=true; shift ;;
+    -h|--help) usage; exit 0 ;;
+    *) echo "Unknown arg: $1"; usage; exit 1 ;;
+  esac
+done
+
+# Verify that the token is set
+if [[ -z "$TOKEN" ]]; then
```
melody-ren, line 57: Is this TOKEN my own (or whoever runs this script) Github login TOKEN?

scripts/prune_cudaqx-dev_by_sha.sh
```diff
@@ -0,0 +1,131 @@
+# ============================================================================ #
```
melody-ren, line 2: Entirely optional:

Might be nice if we put these scripts meant to be run locally or manually in a `local` or `manual` directory

bmhowe23: In general, the items in this folder can be run locally and/or manually. I think the items that aren't intended to be run locally or manually are in the `ci` folder.

scripts/prune_cudaqx-dev_by_sha.sh
bmhowe23: Added in 0861af6

scripts/prune_cudaqx-dev_by_sha.sh
bmhowe23: The script doesn't care who's token it is, but in general, yes, I would expect it to be the user's token. However, if the user's regular token does not have the right scopes, the script will not work. The needed scopes are listed below, and I added a note in the top-level comments added in 0861af6.

PR #281, opened 2025-08-29
Generalize `single_error_lut` to `multi_error_lut`

bmhowe23:
This retains the "single_error_lut" decoder and adds a "multi_error_lut" decoder which is essentially the same as the "single_error_lut" but with `lut_error_depth` > 1.


libs/qec/lib/decoders/single_error_lut.cpp
```diff
@@ -14,11 +14,14 @@
...
+class multi_error_lut : public decoder {
 private:
-  std::map<std::string, std::size_t> single_qubit_err_signatures;
+  std::map<std::string, std::vector<std::size_t>> error_signatures;
+
+  // Input parameters
+  int lut_error_depth = 1;
 
   // List of available result types for this decoder
   const std::vector<std::string> available_result_types = {
```
melody-ren: We should add a note about this list of results. That is - they are for demonstrating usage of `opt_results` only and the returned results aren't all meaningful (`decoding_time` being the exception).

libs/qec/lib/decoders/single_error_lut.cpp
```diff
@@ -28,16 +31,34 @@ class single_error_lut : public decoder {
...
+      }
+    }
+    // Binomial coefficient to check if lut_error_depth is too large
+    auto binom = [](int n, int k) {
+      return 1 / ((n + 1) * std::beta(n - k + 1, k + 1));
+    };
+    if (binom(block_size, lut_error_depth) > 1e9) {
+      throw std::runtime_error(
+          "lut_error_depth is too large for multi_error_lut decoder");
+    }
```
melody-ren: Line 75: 
should change the text to "multi_error_lut decoder"

libs/qec/lib/decoders/single_error_lut.cpp
```diff
@@ -150,6 +216,24 @@ class single_error_lut : public decoder {
...
+      multi_error_lut, static std::unique_ptr<decoder> create(
+                           const cudaqx::tensor<uint8_t> &H,
+                           const cudaqx::heterogeneous_map &params) {
+        return std::make_unique<multi_error_lut>(H, params);
+      })
+};
+
+CUDAQ_REGISTER_TYPE(multi_error_lut)
+
+class single_error_lut : public multi_error_lut {
```
melody-ren: Not a big deal but thought to point this out: as written, warning/error messages for `single_error_lut` will say `multi_error_lut` in the message text (assuming line 75 is fixed). Not that big of a deal in practice since we don't really use these two decoders.

libs/qec/lib/decoders/single_error_lut.cpp
```diff
@@ -78,20 +99,63 @@ class single_error_lut : public decoder {
...
+          H_e2d[c].push_back(r);
+
+    auto toggleSynForError = [&H_e2d](std::string &err_sig, std::size_t qErr) {
+      for (std::size_t r : H_e2d[qErr])
+        err_sig[r] = err_sig[r] == '1' ? '0' : '1';
+    };
 
     // For each qubit with a possible error, calculate an error signature.
-    for (std::size_t qErr = 0; qErr < block_size; qErr++) {
+    if (lut_error_depth >= 1) {
```
melody-ren: Might want to add a note about degeneracy here. Since you're inserting depth-1 (1 error) first, then d2, and d3 and `std::map::insert` doesn't overwrite, the lower weight error wins when there are multiple errors with the same signature.

bmhowe23: I reworked this the solution to make it more general in d1f54fd. I believe my new solution already addresses this comment, so I'll mark as resolved. Let me know you disagree.

libs/qec/lib/decoders/single_error_lut.cpp
bmhowe23: I'd rather just make the output values accurate rather than document that 3 out of the 4 of them are for demonstration usage only. 1a34de5

libs/qec/lib/decoders/single_error_lut.cpp
bmhowe23: Hmm, GitHub is showing me that it already showed multi_error_lut decoder here, so I'm not sure how to interpret this comment. That being said, I just changed it to "LUT" instead in the latest commit.

libs/qec/lib/decoders/single_error_lut.cpp
bmhowe23: I just changed them to the generic "LUT".

libs/qec/lib/decoders/lut.cpp
```diff
@@ -0,0 +1,263 @@
...
+     * {
+     *   "error_probability": true,    // Include error probability in results
+     *   "syndrome_weight": true,      // Include syndrome weight in results
+     *   "decoding_time": false,       // Don't include decoding time
+     * }
+     */
+    if (has_opt_results) {
+      result.opt_results =
+          cudaqx::heterogeneous_map(); // Initialize the optional map
+      // Values are for demonstration purposes only.
```
melody-ren:
```suggestion
```

PR #308, opened 2025-09-29
Prepare for CUDA 13

mitchdz:


.github/workflows/build_dev.yaml
```diff
@@ -80,7 +81,7 @@ jobs:
         TAGS="-t ghcr.io/nvidia/cudaqx-dev:${{ steps.get-cudaq-version-short.outputs.commit_date }}-${{ steps.get-cudaq-version-short.outputs.shortref }}-${{ matrix.platform }}"
         TAGS+=" -t ghcr.io/nvidia/cudaqx-dev:${{ steps.get-cudaq-version-short.outputs.shortref }}-${{ matrix.platform }}"
         TAGS+=" -t ghcr.io/nvidia/cudaqx-dev:${other_tag}-${{ matrix.platform }}"
-        docker build $TAGS -f docker/build_env/cudaqx.dev.Dockerfile . --build-arg base_image=ghcr.io/nvidia/cuda-quantum-devdeps:ext-${{ matrix.platform }}-cu12.6-gcc11-main
+        docker build $TAGS -f docker/build_env/cudaqx.dev.Dockerfile . --build-arg base_image=ghcr.io/nvidia/cuda-quantum-devdeps:ext-${{ matrix.platform }}-cu${{ matrix.cuda_version }}-gcc11-main
```
bmhowe23, line 84: This updates the base image, but if two versions of CUDA are in use, then I think it would be a race as to which image tag got pushed. The reason is because there is no CUDA tag in the output image name.

I suspect this means we need to put the CUDA version in the image names of the created images, and correspondingly, we will need to update downstream CI tests (mainly GitLab in this case) to add the CUDA version to the image name.

.github/workflows/all_libs.yaml
```diff
@@ -11,8 +11,9 @@ jobs:
       fail-fast: false
       matrix:
         platform: ['amd64', 'arm64']
+        cuda_version: ['12.6']
     runs-on: ${{ startsWith(github.repository, 'NVIDIA/cudaqx') && format('linux-{0}-cpu8', matrix.platform) || 'ubuntu-latest' }}
-    container: ghcr.io/nvidia/cuda-quantum-devdeps:ext-${{ matrix.platform }}-cu12.6-gcc11-main
+    container: ghcr.io/nvidia/cuda-quantum-devdeps:ext-${{ matrix.platform }}-cu${{ matrix.cuda_version }}-gcc11-main
```
bmhowe23, line 16: Similarly, I suspect we will need to tag the outputs of this job with CUDA version numbers.

.github/workflows/build_dev.yaml
mitchdz: Updated the tags for this one.

.github/workflows/all_libs.yaml
mitchdz: `all_libs` just builds and runs tests, it doesn't seem there's anything to tag the output of unless I'm missing something here.

bmhowe23: Ah, yes, all_libs may be fine, but all_libs_release.yaml may need updates. (It has a "Upload build artifacts" step.)

mitchdz: Sounds good, I'll look at updating all_libs_release.yaml.

.github/workflows/build_dev.yaml
```diff
@@ -77,10 +78,10 @@ jobs:
           echo "Versioned release: $ver"
           other_tag=$ver
         fi
-        TAGS="-t ghcr.io/nvidia/cudaqx-dev:${{ steps.get-cudaq-version-short.outputs.commit_date }}-${{ steps.get-cudaq-version-short.outputs.shortref }}-${{ matrix.platform }}"
-        TAGS+=" -t ghcr.io/nvidia/cudaqx-dev:${{ steps.get-cudaq-version-short.outputs.shortref }}-${{ matrix.platform }}"
+        TAGS="-t ghcr.io/nvidia/cudaqx-dev:${{ steps.get-cudaq-version-short.outputs.commit_date }}-${{ steps.get-cudaq-version-short.outputs.shortref }}-${{ matrix.platform }}-cu${{ matrix.cuda_version }}"
+        TAGS+=" -t ghcr.io/nvidia/cudaqx-dev:${{ steps.get-cudaq-version-short.outputs.shortref }}-${{ matrix.platform }}-cu${{ matrix.cuda_version }}"
         TAGS+=" -t ghcr.io/nvidia/cudaqx-dev:${other_tag}-${{ matrix.platform }}"
```
bmhowe23: I think this line need to be updated, too. Perhaps: 
```suggestion
        TAGS+=" -t ghcr.io/nvidia/cudaqx-dev:${other_tag}-${{ matrix.platform }}-cu${{ matrix.cuda_version }}"
```
?
If so, then I think https://github.com/NVIDIA/cudaqx/blob/main/Building.md needs to be updated.

mitchdz: Yes, I agree. I'll add that change to the docs.

mitchdz: Added.

bmhowe23 (approved):
Thanks, Mitch.

PR #309, opened 2025-10-01
Fixing the UHF hamiltonian generation

marwafar:


libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
```diff
@@ -50,58 +50,102 @@ def generate_molecular_spin_ham_restricted(self, h1e, h2e, ecore):
...
-                                       2 * s + 1] = 0.5 * h2e[2, p, q, r, s]
-
-                        # Mixed spin(abba, baab) <a|a><b|b>, <b|b><a|a>
-                        two_body_coeff[2 * p + 1, 2 * q, 2 * r,
-                                       2 * s + 1] = 0.5 * h2e_prime[p, q, r, s]
-                        two_body_coeff[2 * p, 2 * q + 1, 2 * r + 1,
-                                       2 * s] = 0.5 * h2e[1, p, q, r, s]
+        # This function generates the molecular spin Hamiltonian
+        # H= E_core+sum_{pq}  h_{pq} a_p^dagger a_q +
+        #  0.5 * h_{pqrs} a_p^dagger a_q^dagger a_s a_r
```
jgonthier, line 86: There is an inconsistency here with the generate_molecular_spin_ham_restricted function: 
- here h_{pqrs} is associated with a_p^dagger a_q^dagger a_s a_r
- in the restricted function h_{pqrs} is associated with a_p^dagger a_q^dagger a_r a_s

From the code at the bottom of the file generating the integrals, it seems that the documentation here is correct since the integrals are in physicist's notation  <1* 2* | 1 2> , but maybe the restricted generation should be corrected for consistency?

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
```diff
@@ -50,58 +50,102 @@ def generate_molecular_spin_ham_restricted(self, h1e, h2e, ecore):
...
+        for p in range(n_mos):
+            for q in range(n_mos):
+                for r in range(n_mos):
+                    for s in range(n_mos):
+                        # --- alpha-alpha ---
+                        val = half * h2e_alpha_alpha[p, q, r, s]
+                        two_body_coeff[self.a_idx(p),
+                                       self.a_idx(q),
+                                       self.a_idx(s),
+                                       self.a_idx(r)] += val
```
jgonthier, line 125: Just noting that the inconsistency in index ordering is also present here, the restricted function above assigns two_body_coef[p,q,r,s] = h[p,q,r,s] and here it is two_body_coeff[p,q,s,r] = h[p,q,r,s]

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
```diff
@@ -134,14 +178,21 @@ def get_spin_hamiltonian(self, xyz:str, spin:int, charge: int, basis:str, symmet
...
-        # To add (coming soon)
+        # To add
 
-        if UR and nele_cas is None:
+        if UR and integrals_natorb:
             raise RuntimeError(
-                "WARN: Unrestricted spin calculation for the full space is not supported yet on Cudaq.\
-                        Only active space is currently supported for the unrestricted spin calculations."
+                "WARN: .[pyscf] WARNING: integrals_natorb is not supported for unrestricted calculations. \
+                Set integrals_natorb to False. HF molecular orbitals are only supported for calculating integrals."
```
jgonthier, line 186: I am not sure I understand "HF molecular orbitals are only supported for calculating integrals."

Do you mean "integrals_natorb only supports RHF orbitals"? Or "UHF integrals can only be calculated with canonical orbitals"?

wsttiger (commented):
Overall, this looks good. There seems to be a typo on line 822:

822: f'solvent Hamiltonian generator missing required option - {option}'

One other issue that I see is that get_spin_hamiltonian() is ~650 lines long. It violates the Single Responsibility Principle (each method should only do one thing).

Consider refactoring into smaller methods (e.g., `_run_mean_field_calculation()`, `_run_mp2_calculation()`, `_compute_integrals_unrestricted()`, etc.).

This would improve:

-Testability
-Readability
-Maintainability
-Reusability

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
```diff
@@ -475,58 +598,123 @@ def get_spin_hamiltonian(self, xyz:str, spin:int, charge: int, basis:str, symmet
...
+                eri_ab = ao2mo.incore.general(
+                    h2e_ao, (myhf.mo_coeff[0], myhf.mo_coeff[0],
+                             myhf.mo_coeff[1], myhf.mo_coeff[1]),
+                    compact=False).reshape(nmo, nmo, nmo, nmo)
+                eri_ba = ao2mo.incore.general(
+                    h2e_ao, (myhf.mo_coeff[1], myhf.mo_coeff[1],
+                             myhf.mo_coeff[0], myhf.mo_coeff[0]),
+                    compact=False).reshape(nmo, nmo, nmo, nmo)
+
+                # Reorder integrals from `(pr|qs) to (pq|rs)`
```
jgonthier: This is a minor thing but here the reordering is from the chemist notation (1* 1 | 2* 2) to the physicist notation <1* 2* | 1 2>, so I think it would be clearer to write "from (pr|qs) to <pq|rs>" as the physicist notation is usually in brakets.

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
```diff
@@ -475,58 +598,123 @@ def get_spin_hamiltonian(self, xyz:str, spin:int, charge: int, basis:str, symmet
...
+                    (active_mo_a, active_mo_a, active_mo_b, active_mo_b),
+                    compact=False).reshape(norb_cas, norb_cas, norb_cas,
+                                           norb_cas)
+                eri_ba = ao2mo.incore.general(
+                    h2e_ao,
+                    (active_mo_b, active_mo_b, active_mo_a, active_mo_a),
+                    compact=False).reshape(norb_cas, norb_cas, norb_cas,
+                                           norb_cas)
+
+                # Reorder integrals from `(pr|qs) to (pq|rs)`
```
jgonthier: Same remark as above regarding the notation, might be clearer to write (pr|qs) to <pq|rs>

libs/solvers/python/tests/test_molecule.py
```diff
@@ -233,3 +233,52 @@ def test_as_with_casscf():
...
+                                       0,
+                                       0,
+                                       UR=True,
+                                       ccsd=True,
+                                       casci=True,
+                                       verbose=True)
+
+    print(molecule.energies)
+    assert np.isclose(-1.11, molecule.energies['hf_energy'], atol=1e-2)
+    assert np.isclose(-1.13, molecule.energies['fci_energy'], atol=1e-2)
```
jgonthier: The testing threshold of 10^{-2} seems a bit loose to me, we should be able to reproduce these energies to 10^{-4} Ha at least, isn't it?

libs/solvers/python/tests/test_molecule.py
```diff
@@ -233,3 +233,52 @@ def test_as_with_casscf():
...
+                                       ccsd=True,
+                                       casci=True,
+                                       verbose=True)
+
+    print(molecule.energies)
+    assert molecule.n_orbitals == 4
+    assert molecule.n_electrons == 4
+    assert np.isclose(molecule.energies['core_energy'], -103.31815, rtol=1e-4)
+    assert np.isclose(molecule.energies['UR-CCSD'], -108.942725, rtol=1e-4)
+    assert np.isclose(molecule.energies['UR-CASCI'], -108.94365, rtol=1e-4)
```
jgonthier: Same remark here, a relative tolerance of 1e-4 corresponds to 0.1 Ha here, which is a lot larger than chemical accuracy.

marwafar:
> Consider refactoring into smaller methods (e.g., _run_mean_field_calculation(), _run_mp2_calculation(), _compute_integrals_unrestricted(), etc.).

@wsttiger  I am unsure if we should do this. These are classical simulations and should not be our target. The goal is to calculate electron integral and generate the Hamiltonian either using HF molecular orbitals, natural orbitals, or CASSCF orbitals. I am just adding tools to enable this.

libs/solvers/python/tests/test_molecule.py
marwafar: I increased the atol to 1e-4 hartree

libs/solvers/python/tests/test_molecule.py
marwafar: This is the difference in hartree (1e-4 hartree). This is way smaller than chemical accuracy.

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
marwafar: UHF electron integral can be only calculated by using the HF molecular orbitals.

In RHF, I allowed in addition to calculating the electron integral using HF molecular orbitals, I allow calculate the electron integral using natural orbitals. I use MP2 to calculate the one particle reduce density matrix.  Then convert the molecular orbitals to natural orbitals and use them to compute electron integrals using natural orbitals.

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
marwafar: For RHF, when converting from chemist notation to physic notation, I use 
```
h2e_cas = np.asarray(h2e_cas.transpose(0, 2, 3, 1), order='C')
```
This makes it <pq|sr>

However, for UHF, when converting from chemist notation to physic notation, I use:
```
h2e_alpha_alpha = eri_aa.transpose(0, 2, 1, 3)
...
```
This makes it <pq|rs>

According to this I define what is h[p,q,r,s] in the "generate_molecular_spin_ham_restricted function" and "generate_molecular_spin_ham_ur" functions. 

I wrote the two implementation at different time and that is why they are different. We can have a new PR later and rewrite them to make all consistent. But current implementation is correct based on how I convert the chemist notation for h2e tensor

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
marwafar: Check comment above

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
marwafar: Fixed

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
marwafar: Fixed

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
jgonthier: Thank you, I saw that your implementation was correct but I did not understand what was happening with RHF. Everything is clear now! 
I agree we can make a separate PR to make everything consistent.

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
jgonthier: All good now, thanks!

libs/solvers/python/cudaq_solvers/tools/molecule/pyscf/generators/gas_phase_generator.py
jgonthier: Great, thank you!

libs/solvers/python/tests/test_molecule.py
jgonthier: I think here the tolerance is set with `rtol` and not `atol`, which means the tolerance is `rtol * 103.31815` for the core energy, so that would be only 0.1 Ha. I think we can just replace `rtol` by `atol`.

marwafar: I replaced rtol with atol

PR #320, opened 2025-10-10
Add CUDA 13 support and add metapackages

bmhowe23:
## Summary
This PR adds support for CUDA 13 (preserving full support for CUDA 12).

## Details

* Bump CUDA-Q commit to one that supports CUDA 13 (https://github.com/NVIDIA/cuda-quantum/pull/3502)
* Produce *2 sets* of wheels, bringing our total number of *regular* wheels to 4:
  * cudaq-qec-cu12
  * cudaq-qec-cu13
  * cudaq-solvers-cu12
  * cudaq-solvers-cu13
* Additionally, produce meta-packages (cudaq-qec and cudaq-solvers) which will automatically choose the appropriate subpackage. We already have cudaq-qec and cudaq-solvers wheels; these 0.5.0 meta-packages will just replace the older 0.4.0 packages at the time of release.
  * These meta-packages will contain the same optional dependencies as the child packages. That is - `pip install cudaq-qec[tensor-network-decoder]` will use the metapackage to automatically transform that into `pip install cudaq-qec-cuXY[tensor-network-decoder]`, where `XY` is the appropriate CUDA version. This also applies to `cudaq-solvers[gqe]`.
* Matrix the build and test portion of the CI to be running both 12.6 and 13.0.

## Todo (in a separate PR)

* Once internal pipelines are updated, update the "release" workflows to incorporate private artifacts from both cu12 and cu13 images.
* Consider removing the `torch` dependency, and instead update the docs to have users manually install the version that they want (pending discussion).

### DIFF
--- a/.github/workflows/build_dev.yaml
+++ b/.github/workflows/build_dev.yaml
@@ -76,7 +76,7 @@ jobs:
         is_versioned=${{ startsWith(github.ref_name, 'releases/') }}
         if ${is_main_branch}; then
           other_tag="latest"
-        elsif ${is_versioned}; then
+        elif ${is_versioned}; then
           ver=`echo ${{ github.ref_name }} | egrep -o "([0-9]{1,}\.)+[0-9]{1,}"`
           echo "Versioned release: $ver"
           other_tag=$ver
@@ -181,7 +181,7 @@ jobs:
         is_versioned=${{ startsWith(github.ref_name, 'releases/') }}
         if ${is_main_branch}; then
           other_tag="latest"
-        elsif ${is_versioned}; then
+        elif ${is_versioned}; then
           ver=`echo ${{ github.ref_name }} | egrep -o "([0-9]{1,}\.)+[0-9]{1,}"`
           echo "Versioned release: $ver"
           other_tag=$ver


.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "7781a8c856b84d5d7714a66ce6a26e4dc804830e"
+    "ref": "456db96850bb6ecf837170729575e0d3a2efa0a4"
```
bmhowe23: TODO: update this once Mitch's PR is merged to main (CUDA-Q).

.github/actions/get-cudaq-wheels/action.yaml
```diff
@@ -119,7 +119,7 @@ runs:
         context: cudaq
         file: ./cudaq/docker/release/cudaq.wheel.Dockerfile
         build-args: |
-          base_image=ghcr.io/nvidia/cuda-quantum-devdeps:manylinux-${{ inputs.platform }}-cu${{ inputs.cuda_version }}-gcc11-main
+          base_image=ghcr.io/nvidia/cuda-quantum-devdeps:manylinux-${{ inputs.platform }}-cu${{ inputs.cuda_version }}-gcc11-test-ubuntu2404
```
bmhowe23: TODO before merge: revert these "test-ubuntu" changes throughout the whole PR.

libs/qec/python/metapackages/setup.py
```diff
@@ -0,0 +1,285 @@
...
+    pass
+else:
+    # Case 2: install `package_name` source distribution
+    with open(os.path.join(setup_dir, "_version.txt"), "r") as f:
+        __version__ = f.read()
+    best_package = _infer_best_package()
+    install_requires = [
+        f"{best_package}=={__version__}",
+    ]
+    if package_name == 'cudaq-qec':
```
wsttiger: I think that you should be able to from setup.py navigate up and parse the individual toml files (cu12, cu13) to read the optional dependencies.

- import tomllib
- parse pyproject.toml.cu{version} into a dict
- read optional dependencies from dict

libs/qec/python/metapackages/setup.py
```diff
@@ -0,0 +1,285 @@
...
+    for pkg in [package_name, f'{package_name}-cu12', f'{package_name}-cu13']:
+        _log(f"Looking for existing installation of {pkg}.")
+        if _check_package_installed(pkg):
+            installed.append(pkg)
+
+    cuda_version = _get_cuda_version()
+    if cuda_version is None:
+        cudaq_qec_bdist = f'{package_name}-cu12'
+    elif cuda_version < 12000:
+        raise Exception(f'Your CUDA version ({cuda_version}) is too old.')
```
wsttiger: raise Exception(
    f'Your CUDA version ({cuda_version}) is not supported. '
    f'Minimum required: CUDA 12.0. '
)

libs/qec/python/metapackages/setup.py
```diff
@@ -0,0 +1,285 @@
...
+    # Find the existing wheel installation
+    installed = []
+    for pkg in [package_name, f'{package_name}-cu12', f'{package_name}-cu13']:
+        _log(f"Looking for existing installation of {pkg}.")
+        if _check_package_installed(pkg):
+            installed.append(pkg)
+
+    cuda_version = _get_cuda_version()
+    if cuda_version is None:
+        cudaq_qec_bdist = f'{package_name}-cu12'
```
wsttiger, line 206: Give a message that you're falling back to cu12

scripts/ci/build_metapackages.sh
```diff
@@ -0,0 +1,57 @@
...
+  # Copy the appropriate LICENSE, NOTICE, and CITATION.cff files to the metapackage directory.
+  rm -rf dist *.egg-info _version.txt $FILES_TO_COPY
+  for file in ${FILES_TO_COPY[@]}; do
+    cp $TOP_DIR/$file .
+  done
+
+  # Create a version.txt file in the metapackage directory.
+  echo $VERSION > _version.txt
+
+  CUDAQ_META_WHEEL_BUILD=1 python3 -m build . --sdist
```
wsttiger, line 61: Verify that the artififacts were created -

```
if [ ! -d "dist" ] || [ -z "$(ls -A dist 2>/dev/null)" ]; then
  echo "Error: Build succeeded but no artifacts created in dist/"
  exit 1
fi

# Verify tarball is valid
for tarball in dist/*.tar.gz; do
  if [ -f "$tarball" ]; then
    if ! tar -tzf "$tarball" > /dev/null 2>&1; then
      echo "Error: Invalid tarball created: $tarball"
      exit 1
    fi
    echo "✓ Successfully built: $(basename "$tarball")"
  fi
done
```

libs/qec/python/metapackages/setup.py
bmhowe23: setup.py runs on both the build machine (i.e. GitHub in this case) and the user's machine when they "pip install cudaq-qec". When it is running on the user's machine, they will not have access to that pyproject.toml.cu{12,13} file.

scripts/ci/build_metapackages.sh
bmhowe23: The overall script will fail and abort if `python3 -m build . --sdist` fails because the top of the script contains `set -e`, so I don't really think this check is needed, but I will add a simpler check here (just as an extra failsafe in case the build command unexpectedly returns 0 when it actually failed).

```
  # Verify the expected file was created.
  EXPECTED_FILE=dist/cudaq_${package}-${VERSION}.tar.gz
  if [ ! -f $EXPECTED_FILE ]; then
    echo "Error: Expected file $EXPECTED_FILE not found"
    exit 1
  fi

  # Verify tarball is valid
  if ! tar -tzf $EXPECTED_FILE > /dev/null 2>&1; then
    echo "Error: Invalid tarball created: $EXPECTED_FILE"
    exit 1
  fi
```

> echo "✓ Successfully built: $(basename "$tarball")"

This is pretty much the last line of the output for the `python3 -m build . --sdist`. No need to repeat it imo.

libs/qec/python/metapackages/setup.py
bmhowe23: Updated in 170f0e8

libs/qec/python/metapackages/setup.py
bmhowe23: Updated in 170f0e8

scripts/ci/build_metapackages.sh
bmhowe23: Updated in 170f0e8

libs/qec/python/metapackages/setup.py
bmhowe23: I was able to deploy the additional files into the package so that they'll exist on the user machine, and we can indeed read from there when we do. Updated in 2e682b5 and a54d397.

.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "7781a8c856b84d5d7714a66ce6a26e4dc804830e"
+    "ref": "43f539c7e45a0a47c56d43be237ffc673b23222b"
```
bmhowe23, line 4: Note: need to update docker/build_env/cudaqx.dev.Dockerfile and the GitHub action that uses it, too.

bmhowe23: Done in 553faa2b95a057a98c949c0c18b2b5f8d876fb84

PR #322, opened 2025-10-16
Added python test procedure

cketcham2333:


Building.md
```diff
@@ -52,6 +52,8 @@ ninja install
 export PYTHONPATH=${CUDAQ_INSTALL_PREFIX}:${CUDAQX_INSTALL_PREFIX}
 export PATH="${CUDAQ_INSTALL_PREFIX}/bin:${CUDAQX_INSTALL_PREFIX}/bin:${PATH}"
 ctest
+cd .. 
+python3 -m pytest -v libs/qec/python/tests --ignore libs/qec/python/tests/test_tensor_network_decoder.py
```
bmhowe23, line 61: It would be good to the comparable command for Solvers, too. Also, for any `--ignore`, could you please add a small note in the next below that just says why they are being ignored. I.e. "The `--ignore` items are for tests that support optional capabilities of the libraries. If desired, you can install the proper dependences in your dev image and then remove the `--ignore` from these commands."

PR #331, opened 2025-11-01
Misc fixes for various build/test environments

bmhowe23:
* Allow the build to skip the TensorRT-based decoder if the user explicitly disables it. (Useful for developer testing.)
* Fix a build warning from the UCCGSD test
* Also, skip the `libs/solvers/python/tests/test_uccgsd.py::test_solvers_vqe_uccgsd_lih` test in CPU-only environments (like some of our CI) because it was taking ~30 minutes just for that one test.

### DIFF
--- a/docker/build_env/cudaqx.dev.Dockerfile
+++ b/docker/build_env/cudaqx.dev.Dockerfile
@@ -19,6 +19,7 @@ LABEL org.opencontainers.image.url="https://github.com/NVIDIA/cudaqx"
 # FIXME: Remove the cmake install once private repos are updated.
 RUN apt-get update && CUDA_DASH=$(echo $cuda_version | tr '.' '-') \
   && apt-get install -y gfortran libblas-dev jq cuda-nvtx-${CUDA_DASH} \
+  && apt-get install -y git-lfs \
   && python3 -m pip install "cmake<4" --user \
   && apt-get autoremove -y --purge && apt-get clean && rm -rf /var/lib/apt/lists/*
 


Building.md
```diff
@@ -40,12 +40,14 @@ git clone https://github.com/NVIDIA/cudaqx.git
...
 
-# Configure your build (adjust as necessary)
+# Configure your build (adjust as necessary). Note TRT requires extra
+# dependencies, so the command below disables it from the build.
 cmake -G Ninja -S .. \
   -DCUDAQ_INSTALL_DIR=$CUDAQ_INSTALL_PREFIX \
   -DCMAKE_INSTALL_PREFIX=${CUDAQX_INSTALL_PREFIX} \
   -DCUDAQ_DIR=${CUDAQ_INSTALL_PREFIX}/lib/cmake/cudaq \
   -DCMAKE_BUILD_TYPE=Release
+  -DCUDAQ_QEC_BUILD_TRT_DECODER=OFF
```
melody-ren: Missing a `\` after `-DCMAKE_BUILD_TYPE=Release`. Without that `\`, will get an error of `-DCUDAQ_QEC_BUILD_TRT_DECODER=OFF: command not found`

Building.md
```diff
@@ -40,12 +40,14 @@ git clone https://github.com/NVIDIA/cudaqx.git
 cd cudaqx
 mkdir build && cd build
 
-# Configure your build (adjust as necessary)
+# Configure your build (adjust as necessary). Note TRT requires extra
+# dependencies, so the command below disables it from the build.
```
melody-ren: I'm a bit confused by this. From #307 's description, it seems like `TensorRT` is the only truly new dependency. If it's installed in the dev container (since this PR adds the installation in the dockerfile), don't we already satisfy the dependency requirement? The text reads like you need to install additional dependencies. 

I mean I agree with showing how to disable building it by demonstrating the `-DCUDAQ_QEC_BUILD_TRT_DECODER=OFF` flag. I just thought the explanation given was a bit confusing.

Building.md
```diff
@@ -55,7 +57,7 @@ export PYTHONPATH=${CUDAQ_INSTALL_PREFIX}:${CUDAQX_INSTALL_PREFIX}
 export PATH="${CUDAQ_INSTALL_PREFIX}/bin:${CUDAQX_INSTALL_PREFIX}/bin:${PATH}"
 ctest
 # Run the python tests
-# The --ignore option is to bypass tests that require additional packages not contained in 
+# The --ignore option is to bypass tests that require additional packages not contained in
 # the standard docker container
```
melody-ren, line 59: If the user disables TRT decoder by following the above cmake commands, then we need to ignore the trt's python test `test_trt_decoder.py` here too, I think?

Building.md
bmhowe23: You are correct that this is misleading in its current state. I added the "Update cudaqx-dev image to include tensorrt" commit after the Building.md change. I will just revert the change to Building.md. I will still keep the cmake option so we can disable it as needed, but since we don't describe every cmake option in the Building.md page, I don't think we need to show this one.

Building.md
bmhowe23: Agree. Since I am reverting the Building.md changes, I think that will automatically resolve this thread.

bmhowe23: Reverted in 82dde3769701648dcf82f7652e9b4d017890ebfa

Building.md
bmhowe23: Reverted in 82dde3769701648dcf82f7652e9b4d017890ebfa

Building.md
bmhowe23: Reverted in 82dde3769701648dcf82f7652e9b4d017890ebfa

libs/qec/CMakeLists.txt
```diff
@@ -61,6 +61,9 @@ option(CUDAQX_QEC_INSTALL_PYTHON
        "Install python files alongside the library."
        ${CUDAQX_INSTALL_PYTHON})
 
+# Option to control TRT decoder build (default: ON)
+option(CUDAQ_QEC_BUILD_TRT_DECODER "Build the TensorRT decoder plugin" ON)
```
melody-ren, line 65: Now that the dev image comes with TensorRT, do we really need to make the trt decoder optional? Either way works. Just curious

bmhowe23: Yes, I'd like to, at least for my own benefit.

bmhowe23: It's a good idea in general to have pluggable components be able to be enabled or disabled on a per component basis.

melody-ren (approved):

PR #332, opened 2025-11-03
Fix TRT Decoder CMake for unsupported platforms

bmhowe23:


libs/qec/lib/decoders/plugins/trt_decoder/CMakeLists.txt
```diff
@@ -56,89 +46,94 @@ find_library(TENSORRT_ONNX_LIBRARY nvonnxparser
...
-  set_target_properties(${MODULE_NAME} PROPERTIES
-    BUILD_RPATH "$ORIGIN:${TENSORRT_ROOT}/lib"
-    INSTALL_RPATH "$ORIGIN:$ORIGIN/../../lib:$ORIGIN/../../../tensorrt_libs"
-  )
+    # RPATH configuration 
+    # ==============================================================================
+
+    if (NOT SKBUILD)
+      set_target_properties(${MODULE_NAME} PROPERTIES
+        BUILD_RPATH "$ORIGIN:${TENSORRT_ROOT}/lib"
```
wsttiger: I think there should be a IF test here - if TENSORRT_ROOT is set (then presumably we're building against the TensorRT in TENSORRT_ROOT) then we can set - otherwise skip.

libs/qec/lib/decoders/plugins/trt_decoder/CMakeLists.txt
```diff
@@ -19,18 +19,7 @@ else()
     message(STATUS "Using TENSORRT_ROOT: ${TENSORRT_ROOT}")
```
wsttiger, line 17: Change line 16 to IF DEFINED and then only include line 19 in the message - The user is using TENSORRT_ROOT

bmhowe23: Updated in a21c05b. Let me know if that's what you had in mind.

libs/qec/lib/decoders/plugins/trt_decoder/CMakeLists.txt
bmhowe23: Updated in a21c05b. Let me know if that's what you had in mind.

wsttiger: LGTM

libs/qec/lib/decoders/plugins/trt_decoder/CMakeLists.txt
wsttiger: LGTM

bmhowe23: Care to formalize with an official approval for the PR then?

bmhowe23: Oops never mind. I missed that you approved the first time around.

PR #333, opened 2025-11-04
Real-time decoding support

bmhowe23:
This PR adds support using our decoders in a real-time, online environment, where the decoders are running on a low-latency connection to a real QPU. The decoder corrections (calculated in the classical computer) can be sent back to a live CUDA-Q program (running on the quantum computer) during the quantum program's execution. Additionally, one can run the decoder(s) in a simulated environment where real-time decoding occurs in the simulation context (i.e. on the user's desktop).

### DIFF
--- a/libs/qec/CMakeLists.txt
+++ b/libs/qec/CMakeLists.txt
@@ -226,3 +226,20 @@ if (QEC_EXTERNAL_DECODERS)
     add_target_libs_to_wheel(${LIB_FILE})
   endwhile()
 endif()
+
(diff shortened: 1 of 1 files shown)


bmhowe23:
These CI failures are expected in the current state. They will be addressed in a follow-up PR. It builds just fine in Docker; there are just a few tests that need to be tweaked for the public GitHub environment.

Hence, I will be merging this now.

Also, noting for posterity's sake: these changes were reviewed offline.

PR #341, opened 2025-11-06
Script and config updates in preparation for 0.5.0

bmhowe23:
This includes:

* Bumping cuda-quantum wheel dependencies from 0.12 to 0.13.
* Updating the GitHub build scripts to accept draft release artifacts that *only* hold the nv-qldpc-decoder (as opposed to duplicates of the things that are built on the public repo).
* Adds CUDA 13 to .github/workflows/all_libs_release.yaml
* Update some of our tests to not fail if they are nv-qldpc-decoder tests that are being run in a CPU-only environment (like some of our GitHub scripts do)

### DIFF
--- a/.github/workflows/all_libs.yaml
+++ b/.github/workflows/all_libs.yaml
@@ -106,7 +106,10 @@ jobs:
           LD_LIBRARY_PATH: ${{ env.MPI_PATH }}/lib:${{ env.LD_LIBRARY_PATH }}
         shell: bash
         run: |
-          pip install numpy pytest cupy-cuda${{ steps.config.outputs.cuda_major }}x cuquantum-cu${{ steps.config.outputs.cuda_major }} torch lightning ml_collections mpi4py transformers quimb opt_einsum torch nvidia-cublas cuquantum-python-cu${{ steps.config.outputs.cuda_major }}==25.09
+          # Install the correct torch first.
+          cuda_no_dot=$(echo ${{ matrix.cuda_version }} | sed 's/\.//')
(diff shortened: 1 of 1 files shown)


.github/workflows/all_libs_release.yaml
```diff
@@ -121,7 +128,18 @@ jobs:
       - name: Install python requirements
         env:
           LD_LIBRARY_PATH: ${{ env.MPI_PATH }}/lib:${{ env.LD_LIBRARY_PATH }}
-        run: pip install numpy pytest cupy-cuda${{ steps.config.outputs.cuda_major }}x cuquantum-cu${{ steps.config.outputs.cuda_major }} torch lightning ml_collections mpi4py transformers quimb opt_einsum torch
+        shell: bash
+        run: |
+          # Install the correct torch first.
+          cuda_no_dot=$(echo ${{ matrix.cuda_version }} | sed 's/\.//')
+          pip install torch==2.9.0 --index-url https://download.pytorch.org/whl/cu${cuda_no_dot}
```
melody-ren, line 135: Is this torch installation mod needed for the other workflows? E.g., `all_libs.yaml`?

bmhowe23: Yes, that would probably be a good idea. Based on https://github.com/NVIDIA/cuda-quantum/pull/3579#discussion_r2500905176, it sounds like this might need another revision though, at least to work with the rest of the CUDA-Q dependencies.

bmhowe23: I'd probably like to handle that update separately though. As written, the changes I have (plus your suggestion) are sufficient to get all of our builds built and the tests passing.

PR #343, opened 2025-11-06
Sliding window optimize

cketcham2333:


libs/qec/lib/decoder.cpp
```diff
@@ -185,7 +217,23 @@ void decoder::set_D_sparse(const std::vector<std::vector<uint32_t>> &D_sparse) {
...
+    // Check if first row is a first-round detector (single syndrome index)
+    pimpl->has_first_round_detectors =
+        (this->D_sparse.size() > 0 && this->D_sparse[0].size() == 1);
+    pimpl->current_round = 0;
+    pimpl->persistent_detector_buffer.resize(pimpl->num_syndromes_per_round);
+    pimpl->persistent_soft_detector_buffer.resize(
+        pimpl->num_syndromes_per_round);
+  } else {
+    pimpl->is_sliding_window = false;
+  }
```
bmhowe23: If this is the same as lines 197-210, can you please refactor into a function?

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -6,545 +6,488 @@
...
-          "sliding_window constructor: error_rate_vec must be non-empty");
-    }
-
-    // Enforce that H is already sorted.
-    if (!cudaq::qec::pcm_is_sorted(H, num_syndromes_per_round)) {
-      throw std::invalid_argument("sliding_window constructor: PCM must be "
-                                  "sorted. See cudaq::qec::simplify_pcm.");
-    }
+/// Helper function to validate constructor inputs.
+void sliding_window::validate_inputs() {
```
bmhowe23, line 17: These diffs are a bit hard to interpret, probably because many functions were extracted outside a class definition, which seems to be confusing GitHub. Could you please summarize the changes to this file here and/or highlight where the real changes occurred?

libs/qec/lib/decoder.cpp
cketcham2333: The code was refactored into a common function in commit #9bb39a4eb51fb16f096e3f9a03a27c285afb9704.  Resolved

libs/qec/lib/decoders/sliding_window.cpp
cketcham2333: The class definitions were extracted into sliding_window.h, while the functions remained in sliding_window.cpp.  The functions in the .cpp file had to be updated to include "sliding_window::" in the name.  I did add one new function at the end of the .cpp file to get the number of syndromes per round.

libs/qec/lib/decoder.cpp
```diff
@@ -20,6 +20,9 @@ INSTANTIATE_REGISTRY(cudaq::qec::decoder, const cudaqx::tensor<uint8_t> &)
 INSTANTIATE_REGISTRY(cudaq::qec::decoder, const cudaqx::tensor<uint8_t> &,
                      const cudaqx::heterogeneous_map &)
 
+// Include decoder implementations AFTER registry instantiation
```
bmhowe23, line 23: Just curious - does something bad happen if you include this before the registry instantiation?

cketcham2333: Yes, it doesn't compile

PR #353, opened 2025-11-11
Bug 5652917: update scripts for the failed test gqe_h2.py

caiyunh:
https://nvbugspro.nvidia.com/bug/5652917  
gqe_h2.py doesn't accept the argument --target, so we need to update the scripts.

### DIFF
--- a/scripts/validation/container/validate_container.sh
+++ b/scripts/validation/container/validate_container.sh
@@ -93,7 +93,10 @@ test_examples() {
     docker exec ${container_name} bash -c "pip install torch==2.9.0 --index-url https://download.pytorch.org/whl/cu${cuda_no_dot}"
     # Install other required packages
     docker exec ${container_name} bash -c "pip install 'lightning>=2.0.0' 'ml_collections>=0.1.0' 'mpi4py>=3.1.0' 'transformers>=4.30.0'"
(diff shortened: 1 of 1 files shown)


scripts/validation/container/validate_container.sh
```diff
@@ -93,7 +93,7 @@ test_examples() {
     docker exec ${container_name} bash -c "pip install torch==2.9.0 --index-url https://download.pytorch.org/whl/cu${cuda_no_dot}"
     # Install other required packages
     docker exec ${container_name} bash -c "pip install 'lightning>=2.0.0' 'ml_collections>=0.1.0' 'mpi4py>=3.1.0' 'transformers>=4.30.0'"
-    docker exec ${container_name} bash -c "pip install 'quimb' 'opt_einsum' 'cuquantum-python-cu${cuda_major}==25.09.1'"
+    docker exec ${container_name} bash -c "pip install 'quimb' 'opt_einsum' 'cuquantum-python-cu${cuda_major}==25.09.1' 'stim' 'beliefmatching'"
```
bmhowe23: @caiyunh can you please confirm that this works on arm64, too? We normally only install stim for amd64 builds because arm64 is a source only package. (Stim is not a dependency of our libraries, but it is a dependency of some of our tests.)

caiyunh: Yes, you are right. I changed the logic to only install `stim` and `beliefmatching` on x86_64 in a471cb42b3513b04fbc7e80f89bcb53736d1f360.

PR #365, opened 2025-11-19
Add `dem_sampling` function for circuit-level noise simulation with per-mechanism error probabilities

kvmto:
**Summary**

Adds dem_sampling() function to support sampling from detector error models (DEMs) with independent, per-error-mechanism probabilities. This generalizes sample_code_capacity() from uniform code capacity noise to realistic circuit-level noise models.


bmhowe23:
Superseded by #479

PR #369, opened 2025-11-25
Add trt cudagraph

wsttiger:
# Add CUDA Graph Optimization to TensorRT Decoder

## Overview

This PR implements CUDA graph optimization for the TensorRT decoder, providing a measurable 20% performance improvement while maintaining full numerical accuracy. The implementation uses a clean executor abstraction pattern that automatically handles models with dynamic shapes.

### 1. CUDA Graph Optimization
- Captures TensorRT inference operations into a CUDA graph on first decode call
- Replays captured graph on subsequent calls, eliminating kernel launch overhead
- ~20% reduction in inference latency (1.24x speedup)
- Follows [NVIDIA TensorRT best practices](https://docs.nvidia.com/deeplearning/tensorrt/latest/performance/best-practices.html)

### 2. Executor Abstraction (PIMPL + Variant Pattern)
- **TraditionalExecutor**: Standard TensorRT execution path
- **CudaGraphExecutor**: CUDA graph-optimized execution path
- Zero-overhead `std::variant` dispatch (no virtual calls)
- Clean separation of concerns, easy to extend

### 3. Dynamic Shape Detection
- Automatically detects models with dynamic tensor dimensions
- Detects models with multiple optimization profiles
- Falls back to traditional execution when CUDA graphs are incompatible
- Prevents runtime errors from improper CUDA graph usage

### 4. User Control
- New parameter: `use_cuda_graph` (default: `true`)
- Users can explicitly disable CUDA graphs if needed
- Clear logging of executor selection and reasoning

## Performance Results

**Benchmark Test (200 iterations):**

### DIFF
--- a/libs/qec/include/cudaq/qec/realtime/decoding_config.h
+++ b/libs/qec/include/cudaq/qec/realtime/decoding_config.h
@@ -88,6 +88,22 @@ struct single_error_lut_config {
   from_heterogeneous_map(const cudaqx::heterogeneous_map &map);
 };
 
+struct trt_decoder_config {
+  std::optional<std::string> onnx_load_path;
+  std::optional<std::string> engine_load_path;
+  std::optional<std::string> engine_save_path;
+  std::optional<std::string> precision;
+  std::optional<std::size_t> memory_workspace;
+
+  bool operator==(const trt_decoder_config &) const = default;
+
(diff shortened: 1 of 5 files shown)


bmhowe23 (commented):
Assuming the Build wheels job passes, these changes look good to me.

bmhowe23 (approved):

PR #372, opened 2025-11-29
UpCCGSD ansatz solver

rr637:
This PR introduces a full UpCCGSD operator-pool and stataprep solver implementation, constructed analogously to the current UCCGSD implementation. I also added analogous Python and C++ tests. 

The UpCCGSD ansatz is the UCC ansatz with generalized singles and paired doubles, as described in this paper: https://arxiv.org/pdf/1810.02327. 

UpCCGSD is lower-depth variant of UCCGSD in which double excitations are restricted to paired electron transfers between spatial orbitals, following the UpCC (unitary paired coupled-cluster) construction. While less expressive than full UCCGSD, it retains generalized spin-preserving singles and the most physically relevant pair-correlation channels, substantially reducing circuit depth and parameter count. This makes UpCCGSD attractive for larger systems or hardware-constrained regimes where UCCGSD is too costly.

@poojarao8

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "0cf5d85b1a86b72ef511d32e68cc2e77afd1a167"
+    "ref": "c4131f47cf236a7694cf6d5b8ff3418933817749"
   }
 }
--- a/.github/workflows/build_dev.yaml
+++ b/.github/workflows/build_dev.yaml
@@ -200,16 +200,16 @@ jobs:
         docker push -a ghcr.io/nvidia/cudaqx-dev
       shell: bash --noprofile --norc -euo pipefail {0}
 
-  cleanup:
-    name: Cleanup
-    needs: [build-cudaqx-dev, build-cudaqx-pydev]
-    runs-on: ubuntu-latest
-    if: ${{ github.repository == 'NVIDIA/cudaqx' }}
-    steps:
-    - name: Delete untagged images
-      uses: actions/delete-package-versions@v5
-      with: 
-        package-name: cudaqx-dev
-        package-type: 'container'
-        min-versions-to-keep: 0
-        delete-only-untagged-versions: 'true'
+  # cleanup:
+  #   name: Cleanup
+  #   needs: [build-cudaqx-dev, build-cudaqx-pydev]
+  #   runs-on: ubuntu-latest
+  #   if: ${{ github.repository == 'NVIDIA/cudaqx' }}
+  #   steps:
+  #   - name: Delete untagged images
+  #     uses: actions/delete-package-versions@v5
+  #     with: 
+  #       package-name: cudaqx-dev
+  #       package-type: 'container'
+  #       min-versions-to-keep: 0
+  #       delete-only-untagged-versions: 'true'
--- a/.github/workflows/build_wheels.yaml
+++ b/.github/workflows/build_wheels.yaml
@@ -341,7 +341,13 @@ jobs:
   
       - name: Test wheels
         run: |
-          for f in /wheels /metapackages /cudaq-wheels; do
+          # If using Custom CUDA-Q wheels, then /cudaq-wheels will exist.
+          # Otherwise, the directory will not exist, and that is ok.
+          WHEEL_DIRS="/wheels /metapackages"
+          if [ "${{ inputs.cudaq_wheels == 'Custom' }}" == "true" ]; then
+            WHEEL_DIRS+=" /cudaq-wheels"
+          fi
+          for f in $WHEEL_DIRS; do
             echo "ls $f"
             ls $f
           done
@@ -451,7 +457,13 @@ jobs:
   
       - name: Test wheels
         run: |
-          for f in /wheels /metapackages /cudaq-wheels; do
+          # If using Custom CUDA-Q wheels, then /cudaq-wheels will exist.
+          # Otherwise, the directory will not exist, and that is ok.
+          WHEEL_DIRS="/wheels /metapackages"
+          if [ "${{ inputs.cudaq_wheels == 'Custom' }}" == "true" ]; then
+            WHEEL_DIRS+=" /cudaq-wheels"
+          fi
+          for f in $WHEEL_DIRS; do
             echo "ls $f"
             ls $f
           done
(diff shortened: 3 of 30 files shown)


H 0-pyscf.chk
bmhowe23, line 1: I think these "H 0" files may have been accidental checkins? If so, could you please remove them from the PR?

rr637: Yes, they were accidental. Removed!

rr637:
Removed only_singles option for UpCCGSD, so only UpCCGSD and UpCCGD are available

docs/sphinx/components/solvers/introduction.rst
```diff
@@ -585,6 +596,26 @@ CUDA-QX provides several state preparations ansatz for VQE.
...
+    # Using UpCCGSD ansatz
+    geometry = [('H', (0., 0., 0.)), ('H', (0., 0., .7474))]
+    molecule = solvers.create_molecule(geometry, 'sto-3g', 0, 0, casci=True)
+
+    numQubits = molecule.n_orbitals * 2
+    numElectrons = molecule.n_electrons
+
+    # Get grouped Pauli words and coefficients from UpCCGSD pool
+    pauliWordsList, coefficientsList = solvers.stateprep.get_upccgsd_pauli_lists(
+        numQubits, only_singles=False, only_doubles=False)
```
jgonthier: The only_singles parameter was suppressed.

libs/solvers/lib/operators/operator_pools/upccgsd_operator_pool.cpp
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      std::size_t R, S;
+      if (p_beta > p_alpha) {
+        R = p_beta;
+        S = p_alpha;
+      } else {
+        R = p_alpha;
+        S = p_beta;
+      }
```
jgonthier: You define q_beta as q_alpha + 1 and p_beta as p_alpha + 1 so I don't think you need to sort anything here, we know that p_beta > p_alpha and q_beta > q_alpha.

libs/solvers/lib/stateprep/upccgsd.cpp
```diff
@@ -0,0 +1,159 @@
...
+      }
+
+      std::size_t R, S;
+      if (p_beta > p_alpha) {
+        R = p_beta;
+        S = p_alpha;
+      } else {
+        R = p_alpha;
+        S = p_beta;
+      }
```
jgonthier: Same as above, I don't think we need to sort them since we know which one is larger.

libs/solvers/lib/stateprep/upccgsd.cpp
```diff
@@ -0,0 +1,159 @@
...
+  } else {
+    // Default: add all UpCCGSD singles...
+    for (auto [p, q] : upccgsd_unique_singles(norbitals))
+      addUpCCGSDSingleExcitation(ops, p, q);
+    // ...and all UpCCGSD paired doubles
+    for (auto pair : upccgsd_unique_doubles(norbitals)) {
+      auto [pq, rs] = pair;
+      addUpCCGSDDoubleExcitation(ops, pq.first, pq.second, rs.first, rs.second);
+    }
+  }
```
jgonthier, line 109: Really minor but would it be more readable to only have 
if(!only_doubles) {
    add singles
}
instead of the if / else?

libs/solvers/python/bindings/solvers/py_solvers.cpp
```diff
@@ -202,6 +213,21 @@ void addStatePrepKernels(py::module &mod) {
                 R"(
   Generate UCCGSD operator pool (Python-style unique singles/doubles) and extract Pauli words and coefficients grouped by excitation.
 
+  Args:
+      num_qubits (int): Number of spin orbitals (qubits).
+      only_singles (bool): If True, only single excitations.
```
jgonthier: I think only_singles was removed.

marwafar: Yes. Because only_singles will be UCCGS which we have in the UCCGSD. So we suggest not to add it here to avoid duplicate.

jgonthier: Yes, sorry I was unclear. I meant to say that since only_singles was removed, it should be removed from the docstring too.

docs/sphinx/components/solvers/introduction.rst
marwafar: If only singles is not enable, you should remove it. It should not be in the option.

libs/solvers/python/bindings/solvers/py_solvers.cpp
marwafar: Agree

libs/solvers/python/tests/test_upccgsd.py
```diff
@@ -0,0 +1,96 @@
...
+                       'maxiter': 1000,
+                       'rhobeg': 0.1,
+                       'disp': True
+                   })
+    energy = res.fun
+
+    hf_energy = molecule.energies.get("HF", None)
+
+    if hf_energy is not None:
+      assert energy <= hf_energy + 1e-4
```
marwafar: Can you add one example about adapt-vqe with uPCCGSD?

marwafar: can you also use something like that in th test "assert np.isclose(energy, -1.1371, atol=1e-4)"?
Can you test example with LiH and also use "assert np.isclose(energy, -7.8638, atol=1e-3)"?

docs/sphinx/components/solvers/introduction.rst
wsttiger: removed

libs/solvers/lib/stateprep/upccgsd.cpp
wsttiger: fixed

libs/solvers/python/bindings/solvers/py_solvers.cpp
wsttiger: fixed

libs/solvers/python/tests/test_upccgsd.py
wsttiger: test was added

libs/solvers/lib/stateprep/upccgsd.cpp
wsttiger: done

libs/solvers/lib/operators/operator_pools/upccgsd_operator_pool.cpp
wsttiger: done

wsttiger (approved):
All conversations have been resolved, and the this passes the CI

PR #373, opened 2025-12-01
Implement GQE Manuscript V2 features and refactor architecture

konakaji:
# Implement GQE Manuscript V2 features (GRPO loss, Replay Buffer, Variance-based scheduler)

## Summary
This PR implements key features from the updated manuscript ([arXiv:2401.09253v2](https://arxiv.org/abs/2401.09253v2)) 
including GRPO loss function, Replay Buffer mechanism, and Variance-based temperature scheduler, 
along with a comprehensive refactoring of the GQE (Generative Quantum Eigensolver) internal 
implementation to improve modularity and add user customization options.

## Motivation
- **Implement Manuscript V2 features**: This PR implements the GRPO loss function, Replay 
  Buffer mechanism, and Variance-based temperature scheduler described in the updated manuscript 
  ([arXiv:2401.09253v2](https://arxiv.org/abs/2401.09253v2)), which were not present 
  in the original implementation
- Previous implementation mixed Fabric and Lightning frameworks, causing complexity
- Model and training logic were tightly coupled

### Core Implementation
- **Add GRPO loss function and Replay Buffer**: Implemented GRPO (Group Relative Policy 
  Optimization) loss and replay buffer mechanism as described in 
  [arXiv:2401.09253v2](https://arxiv.org/abs/2401.09253v2), set as default
- **Add Variance-based temperature scheduler**: Implemented adaptive temperature scheduling 
  based on energy variance as described in the manuscript V2
- **Separate Model and LightningModule**: Decoupled previously combined Transformer into 
  GPT2 (model) and Pipeline (LightningModule)
- **Migrate to PyTorch Lightning**: Unified training implementation to PyTorch Lightning 
  framework, removing Fabric/Lightning mixing. Updated logger from Fabric to Lightning 
  accordingly
- **Extract Scheduler**: Moved scheduler implementations to separate file for better 
  modularity

### User-Facing Features
- **Add customization options**: Users can now provide custom `trainer_kwargs` and 
  `callbacks` to customize Lightning Trainer behavior
- **Add operator pool utilities**: New `utils.py` with helper functions for generating 
  operator pools (`get_identity`, `get_gqe_pauli_pool`)

### Examples and Tests
- **Add N2 molecule example**: New `gqe_n2.py` demonstrating Pauli operator pool usage
- **Add comprehensive tests**: Tests for operator pool utilities, schedulers and updated existing tests

## Breaking Changes
⚠️ **Configuration structure changed**:
- `cfg.fabric_logger` → `cfg.lightning_logger`
- `cfg.use_fabric_logging` → `cfg.use_lightning_logging`

## Testing
- All existing tests pass (12/12 in test_gqe.py)
- New tests added for operator pool utilities
- Verified with H2 and N2 molecule examples

## References
- K. Nakaji et al., "Generative Quantum Eigensolver with Adaptive Ansatz", 
  arXiv:2401.09253v2 (2024)


libs/solvers/python/cudaq_solvers/gqe_algorithm/data.py
```diff
@@ -0,0 +1,44 @@
+from collections import deque
```
melody-ren, line 9: We'll need a licence header for this file too

libs/solvers/python/cudaq_solvers/gqe_algorithm/factory.py
```diff
@@ -0,0 +1,38 @@
+from .loss import ExpLogitMatching, GFlowLogitMatching, GRPOLoss
```
melody-ren, line 9: Please add a licence header

libs/solvers/python/cudaq_solvers/gqe_algorithm/model.py
```diff
@@ -0,0 +1,13 @@
+from transformers import GPT2LMHeadModel, GPT2Config
```
melody-ren, line 9: Please add a licence header

libs/solvers/python/cudaq_solvers/gqe_algorithm/gqe.py
```diff
@@ -6,91 +6,19 @@
 # the terms of the Apache License 2.0 which accompanies this distribution.     #
 # ============================================================================ #
 
-from .transformer import Transformer
+from .pipeline import Pipeline
```
melody-ren, line 9: Can you please confirm that `transformers` is no longer a dependency of the GQE solver? If so, then we need to update solver's `.toml` file as well. Located here: `libs/solvers/pyproject.toml.cu12` and `libs/solvers/pyproject.toml.cu13`

konakaji: Thank you. It's still used in `model.py`, so the dependency does not change.

docs/sphinx/examples/solvers/python/gqe_n2.py
```diff
@@ -0,0 +1,163 @@
...
+n_qubits = molecule.n_orbitals * 2
+n_electrons = molecule.n_electrons
+
+# Generate the operator pool using utility function
+params = [
+    0.003125, -0.003125, 0.00625, -0.00625, 0.0125, -0.0125, 0.025, -0.025,
+    0.05, -0.05, 0.1, -0.1
+]
+
+op_pool = get_gqe_pauli_pool(n_qubits, n_electrons, params)
```
marwafar, line 85: Can you provide brief explanation in the doc what that look like? Other users might be interested in adding their custom pool.

libs/solvers/python/tests/test_gqe.py
bmhowe23, line 1: One of these tests (and I don't know which one yet), are taking a very long time in our CI runs on systems that do not have GPUs. For example: https://github.com/NVIDIA/cudaqx/actions/runs/20625490093/job/59236283754. Before merging this, I think we need to find out which test that is and skip it on CPU-only systems.

libs/solvers/python/cudaq_solvers/gqe_algorithm/utils.py
```diff
@@ -0,0 +1,98 @@
...
+                else:
+                    continue
+
+                if pauli_op is None:
+                    pauli_op = gate
+                else:
+                    pauli_op = pauli_op * gate
+
+            if pauli_op is not None:
+                individual_terms.append(cudaq.SpinOperator(pauli_op))
```
marwafar, line 91: Here you are using UCCSD pool which is a list of list, each list represent either single or double excitation. In the modified gqe pool, you split this into individual terms. I am unsure if that still preserve number of particle and symmetry. If not, it is unclear what ground state you will obtain

libs/solvers/python/tests/test_gqe.py
bmhowe23: > One of these tests (and I don't know which one yet), are taking a very long time in our CI runs on systems that do not have GPUs. For example: https://github.com/NVIDIA/cudaqx/actions/runs/20625490093/job/59236283754. Before merging this, I think we need to find out which test that is and skip it on CPU-only systems.

@konakaji what do you think about disabling all of the test_gqe.py tests if a GPU is not available?

konakaji: Yes, disabling all of the tests in test_gqe.py does make sense to me.

libs/solvers/python/cudaq_solvers/gqe_algorithm/utils.py
konakaji: Thank you, Marwa, for your comments!
I think your comments concern (i) the freedom to choose the operator pool and (ii) symmetry preservation in the current pool.

Regarding the first point, users can still define their own operator pools; an example is already provided in gqe_h2.py. I think referring to it from gqe_n2.py as well addresses your concern.

konakaji: Regarding symmetry preservation, as long as we are targeting the ground-state search of neutral molecules, we can break symmetry in the operator pool, since the lowest-energy state is the one in which symmetry is preserved. In other words, the optimization should restore symmetry.

However, this involves a lot of context, so how do you feel about the following action items?
- Adding the above explanation to the doc of current get_gqe_pauli_pool
- Adding an additional pool example (get_gqe_uccsd_pool) that preserves symmetry, as you suggested

PR #383, opened 2025-12-10
Add decode batch

wsttiger:
### Summary

This PR enhances the TRT decoder by enabling `decode()` to work with models of any batch size and refactoring tests to be fully self-contained.

**Changes:**
- Enable `decode()` for batch-optimized models via automatic zero-padding
- Remove external ONNX file dependencies from tests
- Add parameterized testing for batch sizes [1, 2, 4]

### Problem & Solution

**Before:** `decode()` only worked with `batch_size=1` models, throwing errors otherwise. Tests relied on external ONNX files.

**After:** `decode()` automatically zero-pads inputs for models with `batch_size > 1`. Tests create TensorRT engines programmatically.

#### 1. Zero-Padding Support (9dc3530)

- Detects model batch size and auto-pads: `[syndrome, zeros, zeros, ...]`
- Returns only first result (the real syndrome)
- Added INFO/WARN logging for visibility

# Now works with any batch size
result = decoder.decode(syndrome)  # Auto zero-pads if batch_size > 1

#### 2. Self-Contained Tests (43b8bab)

- Build lightweight MLP engines on-the-fly using TensorRT Python API
- Parameterized fixture runs all tests 3× (batch_size = 1, 2, 4)
- New test validates zero-padding produces identical results to manual padding
- Uses pytest's `tmp_path` for automatic cleanup

**Benefits:**
- No external file dependencies
- Comprehensive batch size coverage
- Fast execution with dummy networks
- Reproducible across all environments

#### 3. Test Fixes (090a605)

- Fixed issues from refactoring
- Reduced code by 23 lines

### DIFF
--- a/libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
+++ b/libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
@@ -77,13 +77,15 @@ public:
     // Suppress the harmless warning about logger reuse that appears when
     // Python tests build engines with Python's trt.Logger and C++ decoders
     // load those engines. Both loggers remain valid; TensorRT uses whichever
-    // was registered first. This is expected behavior for mixed Python/C++ usage.
+    // was registered first. This is expected behavior for mixed Python/C++
+    // usage.
     std::string_view msg_view(msg);
     if (msg_view.find("logger passed into") != std::string_view::npos &&
-        msg_view.find("differs from one already registered") != std::string_view::npos) {
-      return;  // Suppress this specific warning
+        msg_view.find("differs from one already registered") !=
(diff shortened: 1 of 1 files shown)


bmhowe23:
@wsttiger I see a bunch of warnings in these logs: https://github.com/NVIDIA/cudaqx/actions/runs/20119546319/job/57813702181#step:12:6079

```
[TRT] [W] WARNING The logger passed into createInferBuilder differs from one already registered for an existing builder, runtime, or refitter. So the current new logger is ignored, and TensorRT will use the existing one which is returned by nvinfer1::getLogger() instead.
```

Should we be concerned?

bmhowe23 (approved):
Thanks, Scott!

PR #387, opened 2025-12-14
Add Coupled Exchange Operator (CEO) pool

jgonthier:
This PR adds the Coupled Exchange Operator (CEO) pool, and makes it available both for operator pools and state preparation.

CEO are based on qubit excitation operators, which are the equivalent of the UCCGSD excitation operators but without the Z parity strings (thereby making for simpler circuits). In addition, in CEO qubit excitation operators are coupled together to further reduce circuit implementation cost.
The CEO-ADAPT-VQE paper has all the details (https://arxiv.org/abs/2407.08696). The CEO-ADAPT-VQE method is a bit different from previous ADAPT variants in that not all operators added to the circuit are CEO operators, this is *not* implemented in the current PR.

Note that currently, the kernel to apply this pool does not make use of the optimized circuit implementation.


libs/solvers/include/cudaq/solvers/operators/ceo_excitation_utils.h
```diff
@@ -0,0 +1,92 @@
...
+
+/// @brief Generate all alpha single excitation index pairs (p, q) with p > q
+/// @details Both p and q are alpha spin orbitals (even indices in interleaved
+/// convention)
+/// @param numOrbitals Number of spatial orbitals
+/// @return Vector of (p, q) pairs representing alpha single excitations
+std::vector<std::pair<std::size_t, std::size_t>>
+generate_ceo_alpha_singles(std::size_t numOrbitals);
+
+/// @brief Generate all beta single excitation index pairs (p, q) with p > q
```
marwafar, line 25: Minor typo: the function is called "generate_ceo_alpha_singles()" but in the comment below you said Generate all beta single excitation. Check all below too.

marwafar:
LGTM

libs/solvers/include/cudaq/solvers/operators/ceo_excitation_utils.h
jgonthier: Oh yes, for each function the corresponding comment is actually above the function and not below, I think it was done the same way for the UCCSD operator pool.

marwafar: I see. I did not recognize that the comment is above the function. 
Thanks

jgonthier: No worries at all! Thanks for bringing it up.

libs/solvers/include/cudaq/solvers/operators/ceo_excitation_utils.h
```diff
@@ -0,0 +1,92 @@
...
+#include <utility>
+#include <vector>
+
+namespace cudaq::solvers {
+
+/// @brief Generate all alpha single excitation index pairs (p, q) with p > q
+/// @details Both p and q are alpha spin orbitals (even indices in interleaved
+/// convention)
+/// @param numOrbitals Number of spatial orbitals
+/// @return Vector of (p, q) pairs representing alpha single excitations
```
bmhowe23, line 21: None of the doxygen comments in this file are being rendered to the docs. Is that intentional, or is that an accidental oversight?

wsttiger: Fixed

PR #396, opened 2026-01-02
Create PyMatching decoder plugin

bmhowe23:
This PR introduces a PyMatching decoder plugin to the CUDA-Q QEC library.

Ref: https://github.com/oscarhiggott/PyMatching

## Summary

The new PyMatching decoder plugin integrates the PyMatching library into the CUDA-Q QEC framework. It allows users to decode syndromes using MWPM by simply requesting the decoder by name. The implementation handles the conversion of a parity check matrix into a PyMatching graph, supporting both standard edges (between check nodes) and boundary edges.

## Key Changes

* MWPM Decoder Plugin: Implemented a new decoder in `libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp`.
* Supports error_rate_vec as a parameter to define edge weights based on physical error rates.
* Maps the PyMatching merge_strategy options (disallow, independent, smallest_weight, keep_original, replace) to CUDA-Q QEC Plugin options.

## Tests

* C++ Unit Tests:
  * libs/qec/unittests/decoders/pymatching/test_pymatching.cpp
* Python: 
  * libs/qec/python/tests/test_decoder.py
  * libs/qec/python/tests/test_sliding_window.py

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -168,6 +168,12 @@ License at <https://github.com/libprima/prima/blob/main/LICENCE.txt>
  Copyright (c) 2020--2025, Zaikun ZHANG ( <https://www.zhangzk.net> )
 ----------------------------------------------------------------
 
+PyMatching - Apache 2.0
+<https://github.com/oscarhiggott/PyMatching>
+
+License at <https://github.com/oscarhiggott/PyMatching/blob/master/LICENSE>
+----------------------------------------------------------------
+
 Stim - Apache 2.0
 <https://github.com/quantumlib/Stim>
 


libs/qec/python/tests/test_decoder.py
```diff
@@ -339,5 +339,27 @@ def test_single_error_lut_opt_results():
...
+                                  n_syndromes_per_round=5,
+                                  weight=2,
+                                  seed=7)
+    pcm, _ = qec.simplify_pcm(pcm, np.ones(pcm.shape[1]), 10)
+    # Pick 3 random columns from the PCM and XOR them together to get the
+    # syndrome.
+    columns = np.random.choice(pcm.shape[1], 3, replace=False)
+    syndrome = np.sum(pcm[:, columns], axis=1) % 2
+    decoder = qec.get_decoder('pymatching', pcm)
+    print(syndrome)
```
wsttiger: Leftover debug print statement

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
```diff
@@ -0,0 +1,143 @@
...
+    for (size_t i = 0; i < syndrome.size(); i++)
+      if (syndrome[i] > 0.5)
+        detection_events.push_back(i);
+    pm::decode_detection_events_to_edges(mwpm, detection_events, edges);
+    // Loop over the edge pairs
+    assert(edges.size() % 2 == 0);
+    for (size_t i = 0; i < edges.size(); i += 2) {
+      auto edge = make_canonical_edge(edges.at(i), edges.at(i + 1));
+      auto col_idx = edge2col_idx.at(edge);
+      result.result[col_idx] = 1.0;
```
wsttiger, line 128: If PyMatching ever returns an edge that wasn’t registered in edge2col_idx (e.g. after graph merging or a different matching representation), at() will throw std::out_of_range. Options:
Document that the decoder assumes PyMatching’s matching uses only edges from the original graph, or
Use find() and skip or log unknown edges instead of at(), depending on how strict you want to be.

wsttiger (approved):

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
bmhowe23: I don't think that can realistically happen because `edge2col_idx` is made at the same time as the graph is created and any edge merging would strictly reduce the number of possible edges in the graph. I will just add a comment here, and if anything crazy happens in the future, I think that throwing an exception is the right thing to do.

bmhowe23: Updated in b9a5003

libs/qec/python/tests/test_decoder.py
bmhowe23: Thanks. Fixed in b9a5003.

PR #411, opened 2026-01-17
Disable support for calling kernels defined in Python from kernels defined in C++

sacpis:
Disable support for calling kernels defined in Python from kernels defined in C++.

Note: This branch (PR) builds against `features/python.redesign.0` branch in CUDA-Q.

### DIFF
--- a/libs/qec/python/bindings/py_code.cpp
+++ b/libs/qec/python/bindings/py_code.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022 - 2025 NVIDIA Corporation & Affiliates.                  *
+ * Copyright (c) 2022 - 2026 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *


.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "c4131f47cf236a7694cf6d5b8ff3418933817749"
+    "ref": "35054a5bae85129e422c589023a5b30ce32d24c2"
```
bmhowe23, line 4: Just a note - we will want to revert this change, or at least update it to a commit on main, before merging this PR.

sacpis: I think we will have to update it now itself as the jobs are failing in CI.

bmhowe23: If you are updating it to tot main, then are the rest of the changes in the PR needed? I thought tot main would work without any other changes.

sacpis: Yes, you are right. Let me put it back. Seems like `cudaq/utils/registry.h` file has been removed from `CUDA-Q` repo. Cleaning this irrelevant header file from `py_code`.

PR #420, opened 2026-02-02
Update cuda-quantum-devdeps:ext-... to cuda-quantum-devcontainer-...

bmhowe23:
This is needed to accommodate https://github.com/NVIDIA/cuda-quantum/pull/3827 (when combined with upcoming changes).


mitchdz (approved):
I made a duplicate at https://github.com/NVIDIA/cudaqx/pull/422 which is functionally identical. +1

PR #422, opened 2026-02-03
use cudaq devcontainer

mitchdz:
cudaq removed the `ghcr.io/nvidia/cuda-quantum-devdeps:ext-*` containers, and instead moved them to `ghcr.io/nvidia/cuda-quantum-devcontainer:`.


mitchdz:
Duplicate of https://github.com/NVIDIA/cudaqx/pull/420

PR #423, opened 2026-02-03
Mock decoder

cketcham2333:


libs/qec/unittests/decoders/realtime/data/syndromes_multi_err_lut.txt
bmhowe23, line 1: Just noting: this file is 30KB, which is on the upper end of what I would push into the main part of the repo. If we make updates to this file or add other ones, we'll need to use git lfs.

PR #427, opened 2026-02-05
Fpga syndrome inject

cketcham2333:


libs/qec/unittests/utils/hololink_fpga_syndrome_playback.cpp
bmhowe23, line 1: At the moment, I don't think we should distribute this with our standard CUDA-Q QEC distribution (like Python wheels, docker image, etc.), so could you please move this to `libs/qec/unittests/utils`?

cketcham2333: Resolved in commit hash dacea1c9a4e34cd87b735024bc5f7b773012bb66

bmhowe23:
Per discussion, we're closing this one for now. It will probably be replaced by another PR later.

PR #429, opened 2026-02-05
Add cuda graph launch to mock decoder and introduce autonomous decoder CRTP

cketcham2333:


libs/qec/lib/realtime/mock_decode_handler.cu
```diff
@@ -70,40 +68,60 @@ __device__ int mock_decode_rpc(void *buffer, std::uint32_t arg_len,
...
+
+    auto *response = static_cast<cudaq::nvqlink::RPCResponse *>(data_buffer);
+
+    if (g_mock_decoder != nullptr) {
+      uint8_t *measurements = static_cast<uint8_t *>(arg_buffer);
+      uint8_t *corrections = static_cast<uint8_t *>(arg_buffer);
+
+      const auto &ctx = g_mock_decoder->context();
+      g_mock_decoder->decode(measurements, corrections, ctx.num_measurements,
+                             ctx.num_observables);
```
bmhowe23, line 112: Why is "decode" being called only from threadIdx 0 and blockIdx 0? Hopefully this doesn't mean that the decoder is using only a single thread?

cketcham2333: It's a mock decoder - it doesn't need to be multi threaded.

bmhowe23: ok, that's fair. marking as resolved.

PR #434, opened 2026-02-10
Fix the gradient evaluation bugs and add test cases

kaiqiy-nv:
Fix the gradient evaluation bugs
Add test cases to expose the bugs and verify the fix


bmhowe23 (approved):
Thanks!

Issue #379, opened 2025-12-05
cudaqx::tensor fails to allocate large PCMs


kvmto:
## Summary
The `cudaq::qec::generate_random_pcm()` function fails with error `"xbuffer_storage not resizable"` when attempting to create Parity Check Matrices (PCMs) with more than ~50,000 error mechanisms. This appears to be a limitation in the underlying xtensor library used by `cudaqx::tensor`.

## Environment
- **Component**: `cudaqx::tensor` (xtensor implementation)
- **Affected Functions**: `cudaq::qec::generate_random_pcm()` in `libs/qec/lib/pcm_utils.cpp`
- **Error**: `std::runtime_error: xbuffer_storage not resizable`
- **nvq++ Version**:  (https://github.com/NVIDIA/cuda-quantum d7f1551f6c9544b6123a4daf11c391fade6ea590)

### Test Code

```cpp
#include "cudaq/qec/pcm_utils.h"
#include <iostream>
#include <random>

int main() {
  std::cout << "Testing PCM generation for different sizes...\n\n";
  
  std::vector<size_t> test_sizes = {1000, 10000, 50000, 100000};
  
  for (size_t num_mechs : test_sizes) {
    std::cout << "Attempting " << num_mechs << " mechanisms... " << std::flush;
    
    try {
      size_t n_rounds = std::max(size_t(1), num_mechs / 100);
      size_t n_errs_per_round = num_mechs / n_rounds;
      size_t n_syndromes_per_round = std::max(size_t(1), n_errs_per_round / 2);
      int weight = 3;
      
      std::mt19937_64 rng(42);
      auto pcm = cudaq::qec::generate_random_pcm(
          n_rounds, n_errs_per_round, n_syndromes_per_round, weight, std::move(rng));
      
      std::cout << " Success! Shape: " << pcm.shape()[0] << "×" << pcm.shape()[1] 
                << " (rows×cols)\n";
    } catch (const std::exception& e) {
      std::cout << " Failed! Error: " << e.what() << "\n";
    }
  }
  
  return 0;
}
```

### Execution Results
```bash
$ ./test-pcm-allocation
Testing PCM generation for different sizes...

Attempting 1000 mechanisms...  Success! Shape: 500×1000 (rows×cols)
Attempting 10000 mechanisms...  Success! Shape: 5000×10000 (rows×cols)
Attempting 50000 mechanisms...  Success! Shape: 25000×50000 (rows×cols)
[...]

PR #440, opened 2026-02-12
Allow users to construct decoders with sparse parity check matrix

bmhowe23:
Related: #379 

Note: before merging, consider breaking change effects further and prepare a separate docs update PR.

### DIFF
--- a/libs/qec/include/cudaq/qec/sparse_binary_matrix.h
+++ b/libs/qec/include/cudaq/qec/sparse_binary_matrix.h
@@ -1,0 +1,124 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#pragma once
+
+#include "cuda-qx/core/tensor.h"
+#include <cstdint>
+#include <vector>
+
+namespace cudaq::qec {
+
+/// @brief Storage layout for the sparse PCM: column-major (CSC) or row-major
+/// (CSR). All non-zero entries are assumed to be 1; values are not stored.
+enum class sparse_binary_matrix_layout { csc, csr };
+
+/// @brief Sparse parity-check matrix in either CSC or CSR form.
+/// Index types are uint32_t; non-zero values are always 1 and are not stored.
+class sparse_binary_matrix {
+public:
+  using index_type = std::uint32_t;
+
+  /// @brief Construct a sparse PCM in CSC form.
+  /// @param num_rows Number of rows.
+  /// @param num_cols Number of columns.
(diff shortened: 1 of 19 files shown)


libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
```diff
@@ -390,7 +390,7 @@ struct trt_decoder::Impl {
 // trt_decoder method implementations
 // ============================================================================
 
-trt_decoder::trt_decoder(const cudaqx::tensor<uint8_t> &H,
+trt_decoder::trt_decoder(const cudaq::qec::sparse_binary_matrix &H,
```
vedika-saravanan, line 393: will this change break downstream users? Since inputs now need to be explicitly wrapped, this could introduce friction. If backward compatibility is important, it might be worth adding a temporary overload that accepts a tensor and converts internally.

libs/qec/include/cudaq/qec/sparse_binary_matrix.h
```diff
@@ -0,0 +1,124 @@
...
+public:
+  using index_type = std::uint32_t;
+
+  /// @brief Construct a sparse PCM in CSC form.
+  /// @param num_rows Number of rows.
+  /// @param num_cols Number of columns.
+  /// @param col_ptrs Column pointer array (length num_cols + 1); column j has
+  /// indices in \p row_indices[col_ptrs[j] .. col_ptrs[j+1]-1].
+  /// @param row_indices Row indices of non-zeros (length nnz).
+  static sparse_binary_matrix from_csc(index_type num_rows, index_type num_cols,
```
vedika-saravanan, line 32: Looks like `from_csc` / `from_csr` don’t validate inputs currently. It might be worth checking things like pointer array sizes (col_ptrs, row_ptrs), index bounds, and monotonicity of the pointer arrays. If not malformed inputs could lead to subtle bugs or out-of-bounds access later.

libs/qec/lib/decoders/plugins/example/single_error_lut_example.cpp
```diff
@@ -21,22 +21,22 @@ class single_error_lut_example : public decoder {
...
     for (std::size_t qErr = 0; qErr < block_size; qErr++) {
       std::string err_sig(syndrome_size, '0');
       for (std::size_t r = 0; r < syndrome_size; r++) {
         bool syndrome = 0;
         // Toggle syndrome on every "1" entry in the row.
-        // Except if there is an error on this qubit (c == qErr).
-        for (std::size_t c = 0; c < block_size; c++)
-          syndrome ^= (c != qErr) && H.at({r, c});
+        for (std::uint32_t c : H_e2d[qErr])
+          syndrome ^= 1;
```
vedika-saravanan, line 39: Looks like the logic changed from the old code. The old version XORed all columns except qErr, while the new one only toggles non-zeros in column qErr. The new version seems more aligned with a standard LUT but just wanted to confirm.

bmhowe23:
Superceded by #550.

PR #449, opened 2026-02-23
Add optional `O` parameter to PyMatching plugin

bmhowe23:
Add optional `O` parameter to PyMatching plugin to decode to observables instead of decoding to errors.


bmhowe23:
Running extra tests since this is a late change:
* ✅ https://github.com/NVIDIA/cudaqx/actions/runs/24248662824 (wheels on GPU runners w/ nv-qldpc-decoder added in)
* ✅ https://github.com/NVIDIA/cudaqx/actions/runs/24248686887 (Docker files on GPU runners w/ nv-qldpc-decoder added in)

PR #452, opened 2026-02-26
Fix mixer forwarding | Fix MPI implementation | Add test cases

kaiqiy-nv:
Fix mixer (referenceHamiltonian) forwarding in qaoa Python binder.
Fix MPI implementation in adapt simulator.
Add test cases to validate the above modifications.

### DIFF
--- a/libs/solvers/python/tests/test_adapt.py
+++ b/libs/solvers/python/tests/test_adapt.py
@@ -90,7 +90,8 @@ def test_adapt_empty_gradients():
 
     @cudaq.kernel
     def initState(q: cudaq.qview):
-        pass
+        x(q[0])
+        x(q[1])
 
     energy, thetas, ops = solvers.adapt_vqe(initState, H, pool)
     # Should return without crashing; no operators selected
--- a/libs/solvers/python/tests/test_qaoa.py
+++ b/libs/solvers/python/tests/test_qaoa.py
@@ -255,10 +255,14 @@ def test_clique_weighted_nodes():
 
 
 def test_qaoa_custom_mixer_forwarded():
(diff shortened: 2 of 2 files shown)


bmhowe23:
> This CI failure looks unrelated to this PR. It may have been caused by a previous PR. Will need to investigate.

This should be fixed by #453

libs/solvers/python/tests/test_qaoa.py
```diff
@@ -252,3 +252,41 @@ def test_clique_weighted_nodes():
 
     expected_ham = spin.z(0) + 1.5 * spin.z(1) - 2.5
     assert ham == expected_ham
+
+
+def test_qaoa_custom_mixer_forwarded():
```
kvmto, line 257: this test from what I understand is a no-op, asserts all paths are close, would still pass if mixer is dropped. Assert they're different instead?

libs/solvers/python/tests/test_adapt.py
```diff
@@ -79,3 +80,18 @@ def initState(q: cudaq.qview):
...
+
+def test_adapt_empty_gradients():
+    # adapt_vqe must not crash when all commutators [H, O_i] are zero.
+    # H and pool are all diagonal (Z-only), so [H, O_i] = 0 for each.
+    H = spin.z(0) * spin.z(1)
+    pool = [spin.z(0), spin.z(0) * spin.z(1)]
+
+    @cudaq.kernel
+    def initState(q: cudaq.qview):
+        pass
```
kvmto: for consistency with c++ test I would have put two x gates, really minor

PR #460, opened 2026-03-05
Fix BK transformation | Fix JW parity Z-chain | Add test cases

kaiqiy-nv:
Use this PR.
Ignore the closed PR 459.

Fix BK transformation Case 4 and Case 5.
Fix JW parity Z-chain calculation.
Add test cases to validate the fixes.


libs/solvers/unittests/test_molecule.cpp
```diff
@@ -112,6 +112,43 @@ TEST(MoleculeTester, checkSimple) {
   }
 }
 
+// Verify one_particle_op for separated indices (p=0, q=3).
```
wsttiger, line 115: Change the file mode from 755 -> 644 using chmod

kaiqiy-nv: I have just changed the mode to 664 to keep it the same as all files under `libs/solvers/unittests/`.
Since it causes another commit and push, I need your `/ok` to test again. Thanks.

PR #467, opened 2026-03-17
Add error message for missing system dependencies

vedika-saravanan:
This PR improves error handling for missing system libraries. Added CI test to validate the behavior.

Note:

 When a future change introduces a new excluded system library, the PR for that change should:
  1. Add the entry to _KNOWN_SYSTEM_DEPS
  2. Add a corresponding CI test that removes the library and verifies the message


libs/core/python/_system_dep_check.py
```diff
@@ -0,0 +1,42 @@
...
+_KNOWN_SYSTEM_DEPS = {
+    "libgfortran": {
+        "package":
+            "libgfortran5",
+        "instructions":
+            ("  Debian/Ubuntu:  apt-get install -y libgfortran5\n"
+             "  RHEL/Fedora:    dnf install libgfortran\n"
+             "  Conda:          conda install -c conda-forge libgfortran-ng"),
+    },
+    "libquadmath": {
```
bmhowe23, line 19: I believe libquadmath is only required for x86. Therefore, it should not report any errors for ARM if it is not installed for ARM. Does this correctly handle that?

vedika-saravanan: Yes, it handles it correctly. The detection logic is based on string-matching the `ImportError` message. On arm64 the native extension doesn't link against `libquadmath`. So, the dynamic linker will unlikely produce that error. So, the dictionary entry is safe across all architectures

bmhowe23: Sounds good. I'll mark this conversation as resolved.

PR #473, opened 2026-03-25
Fix GQE invalid CUDA handle / AcceleratorError when moving model to GPU

vedika-saravanan:
This PR resolves cudaErrorInvalidResourceHandle / torch.AcceleratorError when running GQE on GPU (e.g. GPT2LMHeadModel(...).to("cuda")) by initializing PyTorch’s CUDA stack and performing a minimal GPU allocation before building the transformer, instead of relying on `conftest.py` preloads and bundled libcudart symlinks.

This should fix https://nvbugspro.nvidia.com/bug/5801752

### DIFF
--- a/libs/solvers/python/cudaq_solvers/gqe_algorithm/gqe.py
+++ b/libs/solvers/python/cudaq_solvers/gqe_algorithm/gqe.py
@@ -377,6 +377,9 @@ def gqe(cost, pool, config=None, **kwargs):
 
     # Don't let someone override the vocab_size
     cfg.vocab_size = len(pool)
+    if torch.cuda.is_available():
+        torch.cuda.init()
+        torch.empty(1, device="cuda")
     cudaqTarget = cudaq.get_target()
     numQPUs = cudaqTarget.num_qpus()
     model = Transformer(


libs/solvers/python/tests/conftest.py
bmhowe23, line 1: It seems like this will fix our test, which is good, but I think we need to do something to make sure any users who use GQE will not encounter this problem. [Our docs](https://nvidia.github.io/cudaqx/examples_rst/solvers/gqe.html) show how we expect users will use it. Could you please review that and let me know if any changes are needed for that usage?

.github/workflows/all_libs.yaml
```diff
@@ -123,6 +123,7 @@ jobs:
         env:
           LD_LIBRARY_PATH: ${{ env.MPI_PATH }}/lib:${{ env.LD_LIBRARY_PATH }}
           OMPI_MCA_pml: ob1
+
```
bmhowe23: Just checking: is this a stray change that is no longer needed?

libs/solvers/python/tests/conftest.py
vedika-saravanan: Sure, I will go over the documentation and let you know if any changes are needed.

.github/workflows/all_libs.yaml
vedika-saravanan: you are right, this is no longer needed I will revert it back.

libs/solvers/python/tests/conftest.py
vedika-saravanan: I have reviewed the docs and updated and pushed the changes to the PR

bmhowe23:
> Thank you, @vedika-saravanan. I am running a special CI job with GPU runners to help perform additional validation on this. If this passes, we will merge it right after.
> 
> https://github.com/NVIDIA/cudaqx/actions/runs/23874899234

@vedika-saravanan - it looks like the above job is produce at least 2 kinds of errors (probably both related).

1. Our code should work on systems that only have CPUs (albeit slower in many cases). [This link](https://github.com/NVIDIA/cudaqx/actions/runs/23874899234/job/69618758997#step:11:6910) shows a job running on a CPU runner that is unexpectedly producing an error now.
2. We are also seeing failures on our GPU runners. [This link](https://github.com/NVIDIA/cudaqx/actions/runs/23874899234/job/69618759075#step:12:2083) is one example, but there are many sub-jobs in that CI run that are producing errors.

Could you please investigate these and provide feedback about what you think the best solution is?

PR #475, opened 2026-03-30
Add graph capture functions to common decoder interface

bmhowe23:


libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -231,6 +231,28 @@ class decoder
...
+  // -- Begin realtime graph dispatch API --
+
+  /// @brief Returns true if this decoder supports graph-based realtime
+  /// dispatch via capture_decode_graph().
+  virtual bool supports_graph_dispatch() const { return false; }
+
+  /// @brief Capture a CUDA graph for realtime dispatch.
+  ///
+  /// Returns a pointer to a cudaq::qec::realtime::graph_resources struct
+  /// (caller must include realtime/graph_resources.h to interpret it).
```
bmhowe23, line 243: Todo - we should probably forward declare this struct name in this file.

bmhowe23: Created #476

PR #487, opened 2026-04-06
Update container validation script

bmhowe23:


scripts/validation/container/validate_container.sh
```diff
@@ -146,8 +146,14 @@ test_examples() {
                 if ! docker exec ${container_name} bash -c "cd /home/cudaq/cudaqx-examples/${domain}/cpp && \
                     for f in *.cpp; do \
                         echo Compiling and running \$f...; \
-                        nvq++ --enable-mlir -lcudaq-${domain} --target ${target} \$f -o test_prog && \
-                        ./test_prog || exit 1; \
+                        if grep -q '^// Compile and run:' \"\$f\"; then \
```
cketcham2333: This exact '"^// Compile and run:" match seems too narrow. code_capacity_noise.cpp uses "Compile and run with", and circuit_level_noise.cpp uses "Compile and run with:". Those files would not use their embedded compile commands here.

cketcham2333 (changes requested):

scripts/validation/container/validate_container.sh
bmhowe23: Updated in 136792cf478446737cccb3859f930859e2f2904d

cketcham2333 (approved):

PR #494, opened 2026-04-08
  GQE: PyTorch GPU compatibility check, exit/skip test on mismatch

vedika-saravanan:
Description

-  Add cuda_utils helpers; get_device() only uses CUDA when PyTorch can run kernels on the GPU.
-  gqe() prints a stderr message and exits 1 if CUDA is visible but unsupported; otherwise keeps the CUDA warm-up.
-  Wheel/container validation skips gqe_h2 when that check fails (inline python3 -c cuda_utils).
-  Docs updated for this behavior.

  
  NVBug: https://nvbugspro.nvidia.com/bug/6058557


libs/solvers/python/cudaq_solvers/gqe_algorithm/cuda_utils.py
```diff
@@ -0,0 +1,54 @@
+# ============================================================================ #
+# Copyright (c) 2025 - 2026 NVIDIA Corporation & Affiliates.                  #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under   #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
```
bmhowe23, line 7: Nit: spacing on trailing `#`'s looks like it needs alignment.

vedika-saravanan: I have fixed it. Resolving this comment

bmhowe23 (approved):

vedika-saravanan:
> > Thanks, @vedika-saravanan. The changes look good to me. I submitted https://github.com/NVIDIA/cudaqx/actions/runs/24149524755 (Build wheels) just for good measure.
> 
> The Build Wheels job is still running. I am going to merge this PR anyway. If we find any issues, we can always revert if absolutely necessary. Thanks!!

Thanks @bmhowe23. One suggestion, since we are staying on CUDA 12.6 wheel, it might be useful to have a workflow to run the tests on new architecture. This would help catch compatibility gaps earlier especially before release candidate.

PR #504, opened 2026-04-15
Fix surface code examples for `measure_result` API change

khalatepradnya:
* CUDA-Q removed the default constructor and assignment operators from `cudaq::measure_result` (commit [98ec5c0](https://github.com/NVIDIA/cuda-quantum/commit/98ec5c09d3238919cd7e537786b26058688dd25c)). 
See https://github.com/NVIDIA/cuda-quantum/pull/3800

* The `surface_code-1.cpp` and `surface_code-2.cpp` examples manually concatenated Z and X syndrome measurements using `std::vector` default construction and element-wise assignment, both now deleted operations.

* Replace separate `se_z_ft` and `se_x_ft` functions with a combined `se_zx_ft` that applies both stabilizer gate sets and calls `mz(ancz, ancx)` to produce a single combined vector directly.

* In `surface_code-1.cpp`, restructure the stationary window loop to issue the overlap re-enqueue at the end of the last round rather than the start of the next window, avoiding a default-constructed vector that the `cudaq-quake` bridge rejects.

### DIFF
--- a/libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
+++ b/libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
@@ -270,8 +270,8 @@ __qpu__ void custom_memory_circuit_stabs(
 
   // Handle the stabilizer lock-in round (numRounds == 1)
   if (numRounds == 1) {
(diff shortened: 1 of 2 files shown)


bmhowe23 (approved):
Thanks, @khalatepradnya!

PR #505, opened 2026-04-16
Add optional CUDA graph resource inspection for AI predecoder

wsttiger:
Add ai_predecoder_service::print_graph_resources() that walks the captured cuGraph and reports per-kernel grid/block dims, register usage, shared memory, and launch totals, plus a node-type summary.

Collection is opt-in via a new collect_resources parameter on capture_graph() because it uses the CUDA driver API to introspect TRT kernels, which perturbs primary-context state and breaks DOCA-based GPU-RoCE on the FPGA bridge. Only the software benchmark exposes a --print-graph-resources flag; the FPGA bridge ignores it and prints a warning.

### DIFF
--- a/libs/qec/lib/realtime/graph_resources.cu
+++ b/libs/qec/lib/realtime/graph_resources.cu
@@ -1,0 +1,197 @@
+/*******************************************************************************
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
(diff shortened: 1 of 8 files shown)


libs/qec/lib/realtime/graph_resources.cu
bmhowe23, line 1: Can this be a .cpp file since there isn't any real CUDA code in here?

PR #507, opened 2026-04-18
Fix ai_decoder_service TRT builder for quantized ONNX (FP8)

wsttiger:
The previous build_engine_from_onnx() path always created a weakly-typed network with createNetworkV2(0) and unconditionally set BuilderFlag::kFP16.  That configuration is correct for FP32/FP16 ONNX but silently miscompiles (or outright crashes) on ONNX files produced by quantization passes: FP8, NVFP4 (untested), and INT8 (untested) Q/DQ graphs require a strongly-typed network so TRT honors the dtypes declared in the ONNX instead of trying to reconcile them against an FP16 hint.

This change:

* Introduces network_typing_override {automatic, weakly_typed, strongly_typed} and an onnx_quant_info struct reporting whether an ONNX contains FP8 / FP4 / INT8 Q/DQ or explicit BF16/FP16 tensors.
* Adds inspect_onnx(), which performs a one-off throwaway TRT parse and walks IQuantizeLayer/IDequantizeLayer outputs plus primary-input dtypes to populate onnx_quant_info.
* In automatic mode, build_engine_from_onnx() inspects the ONNX, creates the network with NetworkDefinitionCreationFlag::kSTRONGLY_TYPED iff requires_strongly_typed(), and sets BuilderFlag::kFP16 only in the weakly-typed branch.  Behavior for unquantized FP32/FP16 ONNX is unchanged (still weakly-typed + FP16 hint).
* Threads network_typing_override through ai_predecoder_service's constructor and adds a --typing {auto|weak|strong} CLI flag on both test_realtime_predecoder_w_pymatching and hololink_predecoder_bridge.
* Adds an --onnx PATH override to the benchmark so non-preset models (e.g. the FP8 variant) can be tested without editing the preset table.

Separately, the driver-side element-size heuristic "(output_bytes == input_bytes + 1) ? uint8 : int32" is replaced by actual TRT tensor volumes via new ai_decoder_service accessors get_input_num_elements() and get_output_num_elements().  The old heuristic happened to give the right answer for the in-tree models but silently halves the detector count for any FP16-IO model.

trt_dtype_size() is replaced by trt_dtype_bits() + trt_tensor_bytes() so sub-byte dtypes (INT4, FP4) and BF16/FP8 are sized correctly, and binding logs report elem_bits instead of elem_size.

Verified end-to-end on GB200 with the d13_r104 predecoder at 500 req/s (2000us rate) over 120s with real syndromes from
/scratch/Ising-Decoding/test_data/9apr2026/10k_shots/p0.003:

                            FP16          FP8
  Density reduction         98.3%        99.0%
  Residual density          0.092%       0.051%
  Pipeline LER              0.00180      0.00130
  Latency p50 / p99 (us)    208 / 370    196 / 345
  PyMatching decode avg     77.3 us      54.2 us

Auto-detect selects strongly-typed for the _fp8.onnx (log line: "Auto-detected ONNX quant signals: fp8=1 ... -> strongly-typed") and stays on the fast weakly-typed + FP16 path for the unquantized ONNX.


wsttiger:
I made this a draft PR because int8 and nvfp4 are untested

libs/qec/unittests/test_ai_decoder_quantized_onnx.cpp
bmhowe23, line 1: Since this file does not have any CUDA kernels (nor should it), can you please rename it to just be .cpp?

wsttiger: Renamed

bmhowe23:
@wsttiger can you please investigate if the CI is running your new tests? I don't think it is.

https://github.com/NVIDIA/cudaqx/actions/runs/26045597566/job/76568838794#step:14:282 (skipped here)
And I don't see it here:
https://github.com/NVIDIA/cudaqx/actions/runs/26045597566/job/76571648950#step:14:7

bmhowe23:
> @wsttiger can you please investigate if the CI is running your new tests? I don't think it is.
> 
> https://github.com/NVIDIA/cudaqx/actions/runs/26045597566/job/76568838794#step:14:282 (skipped here) And I don't see it here: https://github.com/NVIDIA/cudaqx/actions/runs/26045597566/job/76571648950#step:14:7

Pinging on this again, @wsttiger

wsttiger:
Thanks for the ping. I moved `test_ai_decoder_quantized_onnx` into the QEC GPU C++ test selection in CI, so the smoke tests now run from the GPU tests action instead of only being built/skipped in the CPU build/test jobs.

One wrinkle: the CI GPU runners are A100s, and TensorRT reports that FP8 Q/DQ networks require FP8-capable hardware. So the INT8 smoke test runs in CI, while the FP8 runtime/numerics portion now skips on non-FP8 hardware. The FP8 ONNX inspection still runs, and the full FP8 runtime test will run on supported hardware such as GB200.

bmhowe23 (approved):

PR #479, opened 2026-04-01
Add dem_sampling CPU/GPU across C++ and Python

kvmto:
## Summary
- Add `dem_sampling` (C++) with CPU and cuStabilizer-backed GPU paths, and
  expose it through nanobind and Python as `cudaq_qec.dem_sampling` with
  `backend="auto" | "cpu" | "gpu"` and NumPy / PyTorch tensor support
  (incl. CUDA device pointers on the GPU path).
- Tests: new C++ unit tests (`DemSamplingCPU`, `DemSamplingGPU`) and
  `libs/qec/python/tests/test_dem_sampling.py` covering CPU/GPU and
  NumPy/PyTorch inputs.

## Build / packaging
- `FindcuStabilizer.cmake` auto-discovers the library via `CUSTABILIZER_ROOT`,
  `CUQUANTUM_ROOT`, or the active Python environment's `cuquantum-python-cuXX`
  wheel. If missing, CMake errors with a pip install hint.
- `cuquantum-python-cuXX>=26.03.0` is promoted to a core runtime dep of
  `cudaq-qec` (it ships cuStabilizer); `torch` stays optional under
  `tensor_network_decoder` / `all`.
- Built wheels RPATH into the sibling `cuquantum` wheel, so
  `libcustabilizer.so.0` is found without `LD_LIBRARY_PATH`; CI drops the
  previous `CUSTABILIZER_ROOT` / `LD_LIBRARY_PATH` plumbing.

## Test plan
- [ ] C++ `DemSamplingCPU` / `DemSamplingGPU` tests in CI.
- [ ] `test_dem_sampling.py` in CI with CUDA-enabled torch.
- [ ] Wheel builds green on cu12 and cu13.

PR #510, opened 2026-04-20
[enhancement] QEC gpu python tests, follow-up #479

kvmto:
## Summary
Dependent on the merge of PR #479 .

- Add Python GPU test coverage to the QEC GPU CI job (`pr-build-gpu` in `lib_qec.yaml`)
- The GPU job previously only ran C++ tests via `ctest`. Python tests with GPU-dependent code paths (cuQuantum tensor network contractor, DEM sampling GPU backend, TRT decoder inference) were skipped on CPU runners and never executed on GPU runners.
- Install torch (CUDA wheel) and Python test dependencies on the GPU runner
- Add a `pytest` step targeting `test_tensor_network_decoder.py`, `test_dem_sampling.py`, and `test_trt_decoder.py`
- Expand the `ctest` regex to also include `TRTDecoder` C++ tests

## Test plan

- [ ] `pr-build-gpu` job runs successfully on amd64 and arm64
- [ ] "Run GPU Python tests" step shows previously-skipped cuQuantum/CUDA tests now running
- [ ] "Run GPU C++ tests" step includes TRTDecoder tests (amd64)
- [ ] No regressions in existing CPU test jobs (`pr-build`)

### DIFF
--- a/.github/workflows/lib_qec.yaml
+++ b/.github/workflows/lib_qec.yaml
@@ -161,6 +161,11 @@ jobs:
       actions: read
       contents: read
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          set-safe-directory: true
+
       - name: Download build artifacts
         uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


kvmto:
I think it works, the only thing that could be added to the extra gpu tests is to print why some of them were skipped. It could be because the hardware of the runners is not compatible with TRT. Am I overthinking this? @bmhowe23 @ivanbasov

bmhowe23:
> I think it works, the only thing that could be added to the extra gpu tests is to print why some of them were skipped. It could be because the hardware of the runners is not compatible with TRT. Am I overthinking this? @bmhowe23 @ivanbasov

I'm not 100% sure which ones you're referring to, but TRT is unsupported on arm64 + CUDA12, I believe, so that's probably why.

bmhowe23 (approved):

PR #511, opened 2026-04-20
[Docs] Dem Sampling Docs, follow-up #479

kvmto:
This is a follow-up for #479 

### Summary
Adds comprehensive documentation for the `dem_sampling` feature across the Sphinx doc tree: API references (Python and C++), a dedicated example page with narrative explanation, and a new section in the QEC introduction page.

### Changes
- **`docs/sphinx/examples_rst/qec/dem_sampling.rst`** — New dedicated page covering the DEM sampling workflow, GPU acceleration pipeline (sparse Bernoulli sampling, GF(2) sparse-dense multiply, bitpacking), input type handling (NumPy, PyTorch CUDA, PyTorch CPU), and backend selection logic.
- **`docs/sphinx/examples/qec/python/dem_sampling.py`** — Runnable Python example: basic usage, syndrome verification, reproducibility check, explicit CPU backend.
- **`docs/sphinx/examples/qec/cpp/dem_sampling.cpp`** — Runnable C++ example using `dem_sampler::cpu::sample_dem` with syndrome verification.
- **`docs/sphinx/components/qec/introduction.rst`** — Added "DEM Sampling" subsection under "Numerical Experiments" with inline Python/C++ snippets and a link to the full example page.
- **`docs/sphinx/api/qec/python_api.rst`** — Added `autofunction:: cudaq_qec.dem_sampling` section.
- **`docs/sphinx/api/qec/cpp_api.rst`** — Added `doxygenfunction` entries for all 5 C++ API functions (`cpu::sample_dem` x2, `gpu::sample_dem`, legacy `dem_sampling` x2).
- **`docs/sphinx/examples_rst/qec/examples.rst`** — Added `dem_sampling.rst` to the examples toctree.

### Testing
- All 4 code snippets (Python standalone, Python inline, C++ standalone, bash compile command) compiled/ran successfully.
- Sphinx build (`ninja docs`) succeeds with no new warnings.
- Rendered HTML verified for correct heading hierarchy, content, and cross-links.


docs/sphinx/api/qec/python_api.rst
```diff
@@ -98,6 +98,13 @@ Common
 
 .. autofunction:: cudaq_qec.sample_code_capacity
 
+.. _dem_sampling_python_api:
+
+DEM Sampling
+============
+
+.. autofunction:: cudaq_qec.dem_sampling
```
eliotheinrich, line 120: Might not be able to change at this point, but it is a little confusing that the Python API and C++ API have different names for the same function (`dem_sampling` vs `sample_dem`). `sample_dem` is a better verb name, imo.

docs/sphinx/examples_rst/qec/dem_sampling.rst
```diff
@@ -0,0 +1,142 @@
...
+A **detector error model** (DEM) describes the probabilistic relationship
+between independent error mechanisms and the detectors (syndrome bits) that
+observe them. Given a binary check matrix :math:`H` and a vector of per-mechanism
+error probabilities, DEM sampling generates random error vectors and the
+corresponding syndromes via
+
+.. math::
+
+   \text{errors}_{ij} \sim \text{Bernoulli}(p_j), \qquad
+   \text{syndromes} = \text{errors} \cdot H^T \pmod{2}.
```
eliotheinrich, line 16: Could you define these objects and their shapes here, or follow the convention in the `Numerical Experiments - Conventions` section above?

docs/sphinx/api/core/cpp_api.rst
```diff
@@ -13,6 +13,12 @@ Namespaces
     :desc-only:
 .. doxygennamespace:: cudaq::qec::decoding::config
     :desc-only:
+.. doxygennamespace:: cudaq::qec::dem_sampler
+    :desc-only:
+.. doxygennamespace:: cudaq::qec::dem_sampler::cpu
+    :desc-only:
+.. doxygennamespace:: cudaq::qec::dem_sampler::gpu
+    :desc-only:
```
eliotheinrich, line 21: These are currently the only namespaces in the namespaces API section that include briefs; not sure if this is a deliberate convention broken here, or something that should be corrected for the other namespaces as well.

In either case, the briefs seem to describe the implementation functions. They should probably live on the actual function definitions rather than the encapsulating namespace.

docs/sphinx/api/qec/cpp_api.rst
```diff
@@ -44,6 +44,25 @@ Detector Error Model
...
+.. doxygenfunction:: cudaq::qec::dem_sampler::cpu::sample_dem(const cudaqx::tensor<uint8_t> &, std::size_t, const std::vector<double> &, unsigned)
+
+GPU sampling (device pointers, requires cuStabilizer):
+
+.. doxygenfunction:: cudaq::qec::dem_sampler::gpu::sample_dem
+
+Convenience wrappers (delegate to ``cpu::sample_dem``):
+
+.. doxygenfunction:: cudaq::qec::dem_sampling(const cudaqx::tensor<uint8_t> &, std::size_t, const std::vector<double> &)
+.. doxygenfunction:: cudaq::qec::dem_sampling(const cudaqx::tensor<uint8_t> &, std::size_t, const std::vector<double> &, unsigned)
```
eliotheinrich: The comments on these functions describes them as 'legacy API'; should we advertise them here without labelling them as such?

docs/sphinx/examples_rst/qec/dem_sampling.rst
```diff
@@ -0,0 +1,142 @@
...
+   required beyond passing standard ``uint8`` and ``float64`` arrays.
+
+2. **PyTorch CUDA tensors** — The GPU path reads input device pointers directly
+   via ``data_ptr()`` and writes outputs into ``torch.empty`` tensors on the
+   same device, avoiding any host-device copies. This is the fastest path when
+   inputs are already on the GPU. PyTorch is an optional dependency; install
+   with ``pip install torch``.
+
+3. **PyTorch CPU tensors** — With ``backend="gpu"`` the tensors are
+   automatically moved to CUDA (via ``.to(device)``) before sampling. With
```
eliotheinrich, line 104: With `backend="gpu"` but no GPU available and a PyTorch CPU tensor, this path throws with the error message at https://github.com/NVIDIA/cudaqx/pull/511/changes#diff-b06c93489a678a30bcb3d9a6ee5bcb52e3029ff962655b410b35a29385e3ebf8R82-R83, which seems contradictory (CPU tensors are supported, the real issue is that a GPU wasn't available); I believe it should probably hit the descriptive error message at https://github.com/NVIDIA/cudaqx/pull/511/changes#diff-b06c93489a678a30bcb3d9a6ee5bcb52e3029ff962655b410b35a29385e3ebf8R567-R570 first.

libs/qec/python/bindings/py_dem_sampling.cpp
```diff
@@ -621,7 +621,9 @@ void bindDemSampling(nb::module_ &mod) {
         When PyTorch CUDA tensors are provided on the GPU path, outputs are
         returned as PyTorch CUDA tensors. Otherwise, outputs are NumPy arrays.
 
-        PyTorch CPU tensors are not supported. Convert to NumPy first.
+        PyTorch CPU tensors are moved to CUDA automatically under
+        backend="gpu"; under backend="auto" or "cpu" they are rejected —
+        convert to NumPy first.
```
eliotheinrich: Same comment as above when a GPU is unavailable

libs/qec/python/cudaq_qec/dem_sampling.py
```diff
@@ -59,7 +60,8 @@ def dem_sampling(
     CUDA tensors are provided, outputs are returned as CUDA tensors.
     Otherwise, outputs are NumPy arrays.
 
-    PyTorch CPU tensors are not supported. Convert to NumPy first.
+    PyTorch CPU tensors are moved to CUDA under ``backend="gpu"`` but rejected
+    under ``"auto"``/``"cpu"``; convert to NumPy first for the CPU path.
```
eliotheinrich: Same comment as above when a GPU is unavailable

libs/qec/python/bindings/py_dem_sampling.cpp
eliotheinrich: This calls `to(device)`, so it should be wrapped in a try-catch and handled by setting `failure_reason` and returning

eliotheinrich (commented):
Nice docs! Added a few comments for consistency, and flagged a minor bug in the error handling paths.

Is there any reason for the name mismatch between the Python API and C++ API? It feels a little strange to have the same function have two names.

docs/sphinx/api/qec/python_api.rst
kvmto: There is historical reasons for that. If this is a problem we can make a followup in the future where we change this.

docs/sphinx/api/core/cpp_api.rst
kvmto: Good point. I moved the implementation descriptions off the `dem_sampler::cpu` and `dem_sampler::gpu` namespaces and onto the functions themselves; the GPU pipeline steps now live on `gpu::sample_dem`, so the namespaces no longer carry function-level briefs. Rebuilt the docs to confirm it renders.

docs/sphinx/api/qec/cpp_api.rst
kvmto: Agreed. I relabeled these as "Legacy convenience wrappers" and pointed to the preferred `dem_sampler::cpu::sample_dem` overloads above.

docs/sphinx/examples_rst/qec/dem_sampling.rst
kvmto: Done. I added the shapes inline: `H` is [n_checks x n_mechanisms], `errors` is [num_shots x n_mechanisms], and `syndromes` is [num_shots x n_checks]. I also noted that the `errors * H^T` form is the batched, row-vector version of the `S = H * E` convention used elsewhere in the QEC docs.

docs/sphinx/examples_rst/qec/dem_sampling.rst
kvmto: You are right, this was misleading. I fixed the underlying handling: under `backend="gpu"` a CPU tensor is now treated as valid input (it is moved to CUDA), and when no GPU is available it surfaces the descriptive "GPU backend requested but unavailable" error instead of the "CPU tensors not supported" one. Verified with the GPU both present and hidden. Updated the prose to match.

libs/qec/python/bindings/py_dem_sampling.cpp
kvmto: Done. The `.to(device)` moves are now wrapped in a try/catch that sets `failure_reason` and returns false, so a failed move becomes a clean fallback rather than an uncaught exception.

libs/qec/python/bindings/py_dem_sampling.cpp
kvmto: Fixed. Under `backend="gpu"` with no GPU you now get the GPU-unavailable error rather than the CPU-tensor message. The docstring is updated to say CPU tensors are only rejected under `auto`/`cpu`.

libs/qec/python/cudaq_qec/dem_sampling.py
kvmto: Same fix here. The wrapper docstring now notes that CPU tensors are moved to CUDA under `backend="gpu"` (raising a GPU-unavailable error if there is no GPU) and only rejected under `auto`/`cpu`.

eliotheinrich (approved):
LGTM!

docs/sphinx/api/qec/python_api.rst
melody-ren: Let's make an issue to track this deferred action

docs/sphinx/api/qec/cpp_api.rst
```diff
@@ -44,6 +44,26 @@ Detector Error Model
 .. doxygenfunction:: cudaq::qec::decoder_context_from_memory_circuit(const code &, operation, std::size_t, cudaq::noise_model &, bool)
 .. doxygenfunction:: cudaq::qec::dem_from_stim_text(const std::string &, bool)
 
+.. _dem_sampling_cpp_api:
+
+DEM Sampling
```
melody-ren:
```suggestion
Detector Error Model (DEM) Sampling
```

docs/sphinx/api/qec/python_api.rst
```diff
@@ -98,6 +98,13 @@ Common
 
 .. autofunction:: cudaq_qec.sample_code_capacity
 
+.. _dem_sampling_python_api:
+
+DEM Sampling
```
melody-ren:
```suggestion
Detector Error Model (DEM) Sampling
```

docs/sphinx/components/qec/introduction.rst
```diff
@@ -1418,3 +1418,51 @@ Additional Noise Models
       noise.add_all_qubit_channel(
           "x", cudaq::depolarization2(/*probability*/ 0.01),
           /*numControls*/ 1);
+
+DEM Sampling
```
melody-ren: This entire section, and possibly the short paragraph on detector error model inside "CUDAQ QEC by Example" should move to be under the header of "Decoder Framework" since it's part of the tool chain of our decoders. At any rate, it should not stay here, after numerical experiments.

libs/qec/python/bindings/py_dem_sampling.cpp
```diff
@@ -54,6 +54,24 @@ nb::ndarray<nb::numpy, double> asNumpyFloat64(nb::object obj) {
   return arr;
 }
 
+/// True if either input is a torch CPU tensor (torch importable + a CPU
+/// tensor).
```
melody-ren, line 58: How come we are combining source code change in an doc update? Does the doc need these source file updates and if yes, why is that?

docs/sphinx/api/qec/python_api.rst
kvmto: Opened https://github.com/NVIDIA/cudaqx/issues/718 to track unifying the `dem_sampling` (Python) and `sample_dem` (C++) naming.

docs/sphinx/api/qec/cpp_api.rst
kvmto: Applied. I also extended the heading underline to match the longer title; the accepted suggestion left it at the old 12-char width, which the new warnings-as-errors build (#715) flags as "Title underline too short".

docs/sphinx/api/qec/python_api.rst
kvmto: Same underline fix applied here so the warnings-as-errors build passes.

docs/sphinx/components/qec/introduction.rst
kvmto: Done. Moved the "DEM Sampling" section under "Decoder Framework" (just before "Pre-built QEC Decoders"), out of "Numerical Experiments". Left the runnable example page under "QEC by Example" since that is where the example belongs.

libs/qec/python/bindings/py_dem_sampling.cpp
kvmto: These came out of the review on this PR: the gpu-unavailable error path flagged above returned a misleading "CPU tensors not supported" message, and the docs described behavior the code did not quite deliver. The fix landed here so the docs and code agree, and it was approved with that included. Happy to split it into a followup PR if you would prefer to keep this one docs-only.

docs/sphinx/api/qec/cpp_api.rst
melody-ren: oh right. Thanks. I forgot about the underline formatting stuff. Always trips me up

libs/qec/python/bindings/py_dem_sampling.cpp
melody-ren: I see. Generally I'd prefer to keep docs update PRs to docs only. That said, given this change is small and I just verified that it works as described, we can merge as is.  We should add test coverage for this guard and I created an issue to track this #734

PR #512, opened 2026-04-20
Pybind -> Nanobind and bump cuQuantum dependency

bmhowe23:
This is a required change to be compatible with upstream CUDA-Q changes (https://github.com/NVIDIA/cuda-quantum/pull/4311)

In addition, bump cuquantum-python dependencies to 26.3.1 (https://github.com/NVIDIA/cuda-quantum/pull/4342).

### DIFF
--- a/libs/qec/python/bindings/py_decoder.cpp
+++ b/libs/qec/python/bindings/py_decoder.cpp
@@ -16,27 +16,34 @@
 #include <filesystem>
 #include <limits>
 #include <link.h>
-#include <pybind11/complex.h>
-#include <pybind11/functional.h>
-#include <pybind11/numpy.h>
-#include <pybind11/operators.h>
-#include <pybind11/stl.h>
+#include <nanobind/nanobind.h>
+#include <nanobind/ndarray.h>
+#include <nanobind/stl/complex.h>
+#include <nanobind/stl/function.h>
+#include <nanobind/stl/string.h>
+#include <nanobind/stl/optional.h>
+#include <nanobind/stl/unique_ptr.h>
+#include <nanobind/stl/variant.h>
+#include <nanobind/trampoline.h>
+#include <nanobind/stl/vector.h>
 
 #include "cuda-qx/core/kwargs_utils.h"
 #include "cuda-qx/core/library_utils.h"
 #include "type_casters.h"
 
-namespace py = pybind11;
+namespace nb = nanobind;
 using namespace cudaqx;
 
 namespace cudaq::qec {
 
 class PyDecoder : public decoder {
 public:
-  PyDecoder(const py::array_t<uint8_t> &H) : decoder(toTensor(H)) {}
+  NB_TRAMPOLINE(decoder, 1);
+
+  PyDecoder(const nb::ndarray<nb::numpy, uint8_t> &H) : decoder(toTensor(H)) {}
 
   decoder_result decode(const std::vector<float_t> &syndrome) override {
-    PYBIND11_OVERRIDE_PURE(decoder_result, decoder, decode, syndrome);
+    NB_OVERRIDE_PURE(decode, syndrome);
   }
 };
 
@@ -45,21 +52,23 @@ class PyDecoderRegistry {
 private:
   static std::unordered_map<
       std::string,
-      std::function<py::object(const py::array_t<uint8_t> &, py::kwargs)>>
+      std::function<nb::object(const nb::ndarray<nb::numpy, uint8_t> &,
+                               nb::kwargs)>>
       registry;
 
 public:
   static void register_decoder(
       const std::string &name,
-      std::function<py::object(const py::array_t<uint8_t> &, py::kwargs)>
+      std::function<nb::object(const nb::ndarray<nb::numpy, uint8_t> &,
+                               nb::kwargs)>
           factory) {
     cudaq::info("Registering Pythonic Decoder with name {}", name);
     registry[name] = factory;
   }
 
-  static py::object get_decoder(const std::string &name,
-                                const py::array_t<uint8_t> &H,
-                                py::kwargs options) {
+  static nb::object get_decoder(const std::string &name,
+                                const nb::ndarray<nb::numpy, uint8_t> &H,
+                                nb::kwargs options) {
     auto it = registry.find(name);
     if (it == registry.end()) {
       throw std::runtime_error("Unknown decoder: " + name);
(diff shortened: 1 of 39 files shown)


bmhowe23:
@vedika-saravanan could you please take a look at the CI failures? It is failing on a new check that you recently added. I'm sure that some of the changes in this PR are the root cause of the failure, but I'm not sure where the best place to start looking is. I suspect it might be related to the `import cudaq` that was added to `solvers/__init__.py`, but it's not clear why importing cudaq would fail in this way for this test.

bmhowe23:
> @vedika-saravanan could you please take a look at the CI failures? It is failing on a new check that you recently added. I'm sure that some of the changes in this PR are the root cause of the failure, but I'm not sure where the best place to start looking is. I suspect it might be related to the `import cudaq` that was added to `solvers/__init__.py`, but it's not clear why importing cudaq would fail in this way for this test.

I tried an alternative way of doing it in 96ceb941b11a3d690235bb4ad22dbaef12816e5c. I'll let the CI churn on that tonight, but I suspect it will fail in the same way.

sacpis:
@bmhowe23 Is this good to go?

libs/qec/python/bindings/type_casters.h
```diff
@@ -191,12 +236,13 @@ auto copyCUDAQXTensorToPyArray(const cudaqx::tensor<T> &tensor) {
...
   std::memcpy(data_copy, tensor.data(), total_size * sizeof(T));
 
-  // Create a NumPy array using the buffer protocol
-  return py::array_t<T>(
-      {rows, cols},                  // Shape of the array
-      {cols * sizeof(T), sizeof(T)}, // Strides for row-major layout
-      data_copy,                     // Pointer to the data
-      py::capsule(data_copy, [](void *p) { delete[] static_cast<T *>(p); }));
+  size_t arr_shape[] = {rows, cols};
+  int64_t arr_strides[] = {(int64_t)(cols * sizeof(T)), (int64_t)sizeof(T)};
```
sacpis: Should we apply the same fix here as [9ef1089](https://github.com/NVIDIA/cudaqx/pull/512/changes/9ef10894bc9966bdcbd031f4f7bdd85950bf4504)?

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -213,27 +228,32 @@ void bindDecoder(py::module &mod) {
...
+          [](const detector_error_model &self) -> nb::object {
             const auto &t = self.observables_flips_matrix;
-            // Question: do you need py::cast(&self) here?
-            return py::array_t<uint8_t>(
-                t.shape(), {t.shape()[1] * sizeof(uint8_t), sizeof(uint8_t)},
-                t.data());
+            auto rows = t.shape()[0];
+            auto cols = t.shape()[1];
+            size_t shape[2] = {rows, cols};
+            int64_t strides[2] = {(int64_t)(cols * sizeof(uint8_t)),
```
sacpis: Let's remove the strides array and pass nb::none() or we can just omit it.

libs/qec/python/bindings/py_decoding.cpp
```diff
@@ -8,36 +8,38 @@
...
+
+namespace nb = nanobind;
 
 namespace cudaq::qec::decoding {
 
 // Convert a Python iterable of bool/int/measure_result into
 // std::vector<cudaq::measure_result>
 static std::vector<cudaq::measure_result>
-to_measure_result_vector(const py::handle &iterable) {
+to_measure_result_vector(const nb::handle &iterable) {
```
sacpis: Seems like no one is calling it. Safe to remove.

sacpis (approved):
Overall LGTM. Thanks @bmhowe23. Left a few comments.

libs/qec/python/bindings/type_casters.h
sacpis: I mean we can apply the fix similar to `tensor_to_numpy` in `py_solvers.cpp`.

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -73,135 +83,140 @@ class PyDecoderRegistry {
...
+                                  (int64_t)sizeof(uint8_t)};
+            auto arr = nb::ndarray<nb::numpy, uint8_t>(
+                const_cast<uint8_t *>(t.data()), 2, shape, nb::none(), strides);
```
sacpis: Let's remove the strides array and pass `nb::none()` or we can just omit it.

libs/qec/python/bindings/type_casters.h
sacpis: Fixed in [3a8824f](https://github.com/NVIDIA/cudaqx/pull/512/commits/3a8824f5392fa5042c941fb160ed24758c6021da).

libs/qec/python/bindings/py_decoder.cpp
sacpis: Fixed in [3a8824f](https://github.com/NVIDIA/cudaqx/pull/512/commits/3a8824f5392fa5042c941fb160ed24758c6021da).

libs/qec/python/bindings/py_decoding.cpp
sacpis: Fixed in [3a8824f](https://github.com/NVIDIA/cudaqx/pull/512/commits/3a8824f5392fa5042c941fb160ed24758c6021da).

libs/qec/python/bindings/py_decoder.cpp
sacpis: Fixed in [3a8824f](https://github.com/NVIDIA/cudaqx/pull/512/commits/3a8824f5392fa5042c941fb160ed24758c6021da).

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -310,33 +328,31 @@ void bindDecoder(py::module &mod) {
...
-        py::buffer_info buf = H.request();
-
-        if (buf.ndim != 2) {
+        if (H.ndim() != 2) {
           throw std::runtime_error(
               "Parity check matrix must be 2-dimensional.");
         }
 
-        if (buf.itemsize != sizeof(uint8_t)) {
+        if (sizeof(uint8_t) != sizeof(uint8_t)) {
```
bmhowe23: This looks like a bug that needs to be fixed.

bmhowe23: Fixed in 828e22a

PR #518, opened 2026-04-24
Reduce size of cudaqx-dev image and make TensorRT dependency checking automatic

bmhowe23:
`CUDAQ_QEC_BUILD_TRT_DECODER` was a boolean option defaulting to ON, which meant a default build on a machine without TensorRT emitted a series of CMake errors unless the user explicitly disabled it.

Convert it to a tri-state CACHE STRING with values `AUTO / ON / OFF`
(default `AUTO`):

  - `AUTO`: probe for TensorRT; build the plugin if found, otherwise skip it silently with a single STATUS message explaining how to enable it.
  - `ON`  : probe for TensorRT; FATAL_ERROR if it cannot be found.
  - `OFF` : skip the plugin and do not probe for TensorRT.

Note: this should reduce the size of our cudaqx-dev image by >5GB.


wsttiger (approved):
All good!

PR #522, opened 2026-04-27
(qec): Tensor Network dependency cleanup

vedika-saravanan:
Trims imports and transitive deps along the tensor-network decoder path (e.g. drop CuPy, lazy 
  `opt_einsum`/cuQuantum, typing/import hygiene, dead code removal, small bootstrap fixes). No 
  intentional behavior change on the decoding path.

Followup planned PRs:

 • PR 1: Optimize and clean up Nico’s noise-learning change: API, tests, and baseline performance.
 • PR 2: Documentation updates.


.github/workflows/lib_qec.yaml
```diff
@@ -104,7 +104,7 @@ jobs:
           # Install the correct torch first.
           cuda_no_dot=$(echo ${{ matrix.cuda_version }} | sed 's/\.//')
           pip install torch==2.9.0 --index-url https://download.pytorch.org/whl/cu${cuda_no_dot}
-          pip install numpy pytest onnxscript cupy-cuda${{ steps.config.outputs.cuda_major }}x cuquantum-cu${{ steps.config.outputs.cuda_major }} quimb opt_einsum nvidia-cublas "cuquantum-python-cu${{ steps.config.outputs.cuda_major }}>=26.3.0"
```
kvmto, line 107: there is a cupy dependency still at line 214 :)
is that supposed to be there?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -5,65 +5,60 @@
...
+    return oe.contract(subscripts, *tensors, optimize=optimize)
 
 
-def contractor(subscripts: str,
-               tensors: list[Any],
-               optimize: str = "auto",
-               slicing: tuple = tuple(),
-               device_id: int = 0) -> Any:
+def oe_torch_contractor(subscripts: str,
+                        tensors: list[torch.Tensor],
```
kvmto: this could break if __future__. annotations is not there, works because stringified, shall we import proper torch typing?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
```diff
@@ -5,16 +5,27 @@
...
+
+from typing import Any
+
 import numpy as np
+import numpy.typing as npt
+import torch
 from quimb import oset
 from quimb.tensor import TensorNetwork, Tensor
 
+from ..tensor_network_decoder import TensorNetworkDecoder
```
kvmto: this seems to create a circular import, I would try to test it.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -5,65 +5,60 @@
...
-        slicing (tuple, optional): Not supported in this implementation.
-            Defaults to empty tuple.
-        device_id (int, optional): Device ID for the contraction. Defaults to 0.
 
     Returns:
-        torch.Tensor: The contracted tensor.
+        Any: The contracted tensor.
     """
-    return torch.einsum(subscripts, *tensors)
+    import opt_einsum as oe
```
kvmto: can we have this outside?
It is a hard dependency and we use it in the whole file.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -104,6 +137,7 @@ def optimize_path(optimize: Any, output_inds: tuple[str, ...],
     Returns:
         tuple[Any, Any]: The contraction path and optimizer info.
     """
+    from cuquantum import tensornet as cutn
```
kvmto: There are 2 cuQuantum lazy imports in contractors.py. cutn_contractor's semmes more cosmetic. optimize_path's is broken since it always runs, regardless of which contractor the user picked. Please push it inside a "if is_cutn" branch and duck-type the type check so the isinstance doesn't itself trigger the import.

```def optimize_path(optimize, output_inds, tn):
    is_cutn = optimize is None or (
        type(optimize).__module__.startswith("cuquantum")
        and type(optimize).__name__ == "OptimizerOptions"
    )
    if is_cutn:
        from cuquantum import tensornet as cutn
        return cutn.contract_path(
            tn.get_equation(output_inds=output_inds),
            *tn.arrays,
            optimize=optimize,
        )
    ci = tn.contraction_info(output_inds=output_inds, optimize=optimize)
    return ci.path, ci
```

kvmto: welcome to hear you opinion as well here.
I just want things to break :)
(from experience)

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -88,6 +84,43 @@ def cutn_contractor(subscripts: str,
     )
 
 
+_oe_expr_cache: dict[tuple, Any] = {}
```
kvmto: wondering if this can explode in terms of memory. I would make a small estimate

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -133,13 +167,19 @@ class ContractorConfig:
...
+        ("oe_torch_compiled", "torch", "cuda"),
         ("cutensornet", "numpy", "cuda"),
         ("cutensornet", "torch", "cuda"),
     )
     _allowed_backends: ClassVar[list[str]] = ["numpy", "torch"]
     _contractors: ClassVar[dict[str, Callable]] = {
         "numpy": contractor,
         "torch": contractor,
+        "oe_torch": oe_torch_contractor,
+        "oe_torch_compiled": oe_torch_compiled_contractor,
```
kvmto: Are these reachable?

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
```diff
@@ -100,3 +111,305 @@ def error_pairs_noise_model(
...
+                device=dev,
+            )
+        else:
+            self._observable_flips = value.bool().to(dev)
+        self.obs_idx_true = torch.where(self._observable_flips)[0]
+        self.obs_idx_false = torch.where(~self._observable_flips)[0]
+
+    # -- noise model rebuild ---------------------------------------------------
+
+    def _rebuild_noise_from_params(self) -> None:
```
kvmto: _rebuild_noise_from_params uses positional iteration (enumerate(self.noise_model.tensors)) and assumes that tensor i holds open index error_inds[i]. 

That holds for the factory-built TN today, but it's an unwritten invariant: any user passing a hand-built TensorNetwork, any future change in quimb's internal tensor ordering, or any reordering of error_inds after construction will silently misalign _noise_probs with the tensors they're supposed to drive, gradients flow to the wrong probabilities and the loss still goes down. 

NMOptimizerCompiled already does the right thing by keying off t.inds[0]. please apply the same pattern in the parent class (lookup via noise_model.ind_map[ei] or a precomputed error_index --> tensor dict in __init__).


tldr: we need tests :)

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/noise_models.py
```diff
@@ -100,3 +111,305 @@ def error_pairs_noise_model(
             tags=oset([etag]),
         ))
     return TensorNetwork(tensors)
+
+
+class NMOptimizer(TensorNetworkDecoder):
```
kvmto: Please check the logic of this class, we need to write proper tests everywhere

kvmto (commented):
Thanks Vedika, looks pretty good! :)
Very appreciated!
I have some small concerns here and there and questions regarding the new big class `NMOptimizer`. I would double check the logic.
let's have a discussion :)
I am not sure if this helps or was the purpose of the PR, but merging this could introduce a faulty new feature

.github/workflows/lib_qec.yaml
vedika-saravanan: fixed in https://github.com/NVIDIA/cudaqx/pull/522/changes/b7100ab7ba4c6f3ceeb309f26fe1ef6dddb531e1

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
vedika-saravanan: fixed in https://github.com/NVIDIA/cudaqx/pull/522/changes/b7100ab7ba4c6f3ceeb309f26fe1ef6dddb531e1

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
vedika-saravanan: fixed in https://github.com/NVIDIA/cudaqx/pull/522/changes/b7100ab7ba4c6f3ceeb309f26fe1ef6dddb531e1

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
vedika-saravanan: I agree with you Kevin. fixed in https://github.com/NVIDIA/cudaqx/pull/522/changes/b7100ab7ba4c6f3ceeb309f26fe1ef6dddb531e1

vedika-saravanan:
> Thanks Vedika, looks pretty good! :) Very appreciated! I have some small concerns here and there and questions regarding the new big class `NMOptimizer`. I would double check the logic. let's have a discussion :) I am not sure if this helps or was the purpose of the PR, but merging this could introduce a faulty new feature

Thanks, @kvmto, for the detailed review! To clarify the intent of this PR, the goal here is to import Nico's existing noise learning implementation and do a small dependency/import cleanup, so that follow-up PRs can focus purely on functional changes and are easier to review. I have addressed comments 1, 2, 4, and 5 in commit https://github.com/NVIDIA/cudaqx/commit/b7100ab7ba4c6f3ceeb309f26fe1ef6dddb531e1.

The logic of NMOptimizer, circular import concern and the positional iteration concern in _rebuild_noise_from_params are on my radar. I have a follow-up PR roadmap that covers logic review, the DEM bridge, and proper tests. Keeping this PR scoped to cleanup makes it easier to diff what's actually new in each subsequent PR.

Happy to discuss the roadmap if that helps give more context!

kvmto:
> > Thanks Vedika, looks pretty good! :) Very appreciated! I have some small concerns here and there and questions regarding the new big class `NMOptimizer`. I would double check the logic. let's have a discussion :) I am not sure if this helps or was the purpose of the PR, but merging this could introduce a faulty new feature
> 
> Thanks, @kvmto, for the detailed review! To clarify the intent of this PR, the goal here is to import Nico's existing noise learning implementation and do a small dependency/import cleanup, so that follow-up PRs can focus purely on functional changes and are easier to review. I have addressed comments 1, 2, 4, and 5 in commit [b7100ab](https://github.com/NVIDIA/cudaqx/commit/b7100ab7ba4c6f3ceeb309f26fe1ef6dddb531e1).
> 
> The logic of NMOptimizer, circular import concern and the positional iteration concern in _rebuild_noise_from_params are on my radar. I have a follow-up PR roadmap that covers logic review, the DEM bridge, and proper tests. Keeping this PR scoped to cleanup makes it easier to diff what's actually new in each subsequent PR.
> 
> Happy to discuss the roadmap if that helps give more context!

Hi @vedika-saravanan thanks a lot for the quick addressing!
How do you plan to address the last comments? :)
PR to this PR?

bmhowe23:
It would be a good idea to run this PR through the "Build wheels" job. In fact, we could hook that up to the regular per-PR if that would make things easier.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_utils/contractors.py
```diff
@@ -5,65 +5,60 @@
...
+from torch import Tensor
 from quimb.tensor import TensorNetwork
 
 
-def einsum_torch(subscripts: str,
-                 tensors: list[torch.Tensor],
-                 optimize: str = "auto",
-                 slicing: tuple = tuple(),
-                 device_id: int = 0) -> Any:
+def contractor(subscripts: str,
```
bmhowe23, line 18: Just confirming based on offline discussion: we're removing `einsum_torch` because it is not used, and GitHub is making the diffs appear more confusing than necessary.

vedika-saravanan:
> It would be a good idea to run this PR through the "Build wheels" job. In fact, we could hook that up to the regular per-PR if that would make things easier.

I have started build wheel test: https://github.com/NVIDIA/cudaqx/actions/runs/25079912031
I will create a separate PR to add it to regular PR check.

vedika-saravanan:
> > > Thanks Vedika, looks pretty good! :) Very appreciated! I have some small concerns here and there and questions regarding the new big class `NMOptimizer`. I would double check the logic. let's have a discussion :) I am not sure if this helps or was the purpose of the PR, but merging this could introduce a faulty new feature
> > 
> > 
> > Thanks, @kvmto, for the detailed review! To clarify the intent of this PR, the goal here is to import Nico's existing noise learning implementation and do a small dependency/import cleanup, so that follow-up PRs can focus purely on functional changes and are easier to review. I have addressed comments 1, 2, 4, and 5 in commit [b7100ab](https://github.com/NVIDIA/cudaqx/commit/b7100ab7ba4c6f3ceeb309f26fe1ef6dddb531e1).
> > The logic of NMOptimizer, circular import concern and the positional iteration concern in _rebuild_noise_from_params are on my radar. I have a follow-up PR roadmap that covers logic review, the DEM bridge, and proper tests. Keeping this PR scoped to cleanup makes it easier to diff what's actually new in each subsequent PR.
> > Happy to discuss the roadmap if that helps give more context!
> 
> Hi @vedika-saravanan thanks a lot for the quick addressing! How do you plan to address the last comments? :) PR to this PR?

Hi @kvmto, I have removed the new class objects, this PR only contains dependency cleanup. Once this gets approved and merged to main, I will work on checking and optimizing the logic for Noise Learning.

kvmto:
LGTM on the reworked scope, much smaller and easier to reason about. :)
The cuQuantum import in `optimize_path` is now correctly gated by the duck-typed module check, the workflow leftovers are cleaned up consistently, and pulling `NMOptimizer` out into a follow-up is the right call. A separate PR can help.

A few small things below, one clarifying question and three line-level nits. Nothing here blocks merge.

1. was the plugin loader actually working on `main`?

The rename in `cudaq_qec/__init__.py` looks like it might be a real bug fix, not just a refactor:

```python
# pre-PR
from .plugins import decoders, codes
...
for finder, name, ispkg in iter_namespace(plugins.decoders):  # `plugins` not bound locally
    ...
```
`from .plugins import decoders, codes` only binds `decoders` and `codes` to the local namespace , `plugins` is never bound. And `cudaq_qec/plugins/` is a PEP 420 implicit namespace package (no `__init__.py`), so there's no implicit re-export either. That for-loop should have been raising `NameError` at import time on `main`.

Could you double-check with a quick sanity test against a stock wheel from `main` like:

```bash
python -c "import cudaq_qec; print(cudaq_qec.qecrt.get_available_decoders())"
```
to confirm the auto-loader was actually doing its job before this PR? 
If it wasn't, this fix is also turning on a previously-dead code path, and I'd want to be sure that doesn't have side effects (previously-unloaded plugins now registering, decoder-name collisions, etc.). 


2. Duplicate import in `test_tensor_network_decoder.py:23-25` :)

3. `import torch` at the top of `test_tensor_network_decoder.py:13`
Pre-PR, this file never imported torch at module top (the CuPy check was lazy inside `is_nvidia_gpu_available`). Now if anyone runs the qec test suite without the `[tensor_network_decoder]` extras, the file fails to collect instead of skipping cleanly. 

4. Stale docstring in `tensor_network_decoder.py:481`
`cutn` is no longer imported in this file (the lazy `from cuquantum.tensornet import OptimizerOptions` was removed). Suggest `cuquantum.tensornet.OptimizerOptions` for consistency with the second line.


Everything else like positional iteration in `_rebuild_noise_from_params`, the `NMOptimizer` circular-import, per-step path search in the trainer's forward pass, lack of tests for the trainer belongs on the follow-up :)
Good job!

kvmto (commented):
quick check into the big comment

vedika-saravanan:
@kvmto , thanks for the detailed feedback, 

> 1. was the plugin loader actually working on main?

On main, the loader path is broken, plugins aren’t bound, so it would hit a NameError before loading anything. I verified this with a quick script: main fails namespace iteration, while this branch correctly enumerates decoders/codes. So, there wasn’t a working autoloader before, this change enables it. In a sanity test, only expected in-tree plugins were registered, with no unexpected side effects or collisions. Optionally, we could add a lightweight stdlib-only smoke check in CI for regression coverage (happy to follow up if useful).

> 2. Duplicate import in test_tensor_network_decoder.py:23-25 :)

Fixed in https://github.com/NVIDIA/cudaqx/pull/522/commits/21b210039c0c0a39eab78cb29f434417247ce347

> 3. import torch at the top of test_tensor_network_decoder.py:13

Replaced top-level import torch with pytest.importorskip("torch", ...) so missing optional deps skip this file cleanly instead of failing collection, when torch is present it’s equivalent to import torch. Fixed in https://github.com/NVIDIA/cudaqx/pull/522/commits/21b210039c0c0a39eab78cb29f434417247ce347

> 4. Stale docstring in tensor_network_decoder.py:481

Fixed in https://github.com/NVIDIA/cudaqx/pull/522/commits/21b210039c0c0a39eab78cb29f434417247ce347

bmhowe23:
> On main, the loader path is broken, plugins aren’t bound, so it would hit a NameError before loading anything.

Could you please elaborate on this? It doesn't seem to be broken for me. I added the following code to confirm, and it is working as expected for me (on `main`). That is - I did not see any "failed to import" messages once I had the proper dependencies installed.

```python
for finder, name, ispkg in iter_namespace(plugins.decoders):
    try:
        importlib.import_module(name)
        print(f'{name} imported successfully')
    except (ModuleNotFoundError, ImportError) as e:
        print(f'failed to import {name}, {e}')
        pass
```

vedika-saravanan:
> > On main, the loader path is broken, plugins aren’t bound, so it would hit a NameError before loading anything.
> 
> Could you please elaborate on this? It doesn't seem to be broken for me. I added the following code to confirm, and it is working as expected for me (on `main`). That is - I did not see any "failed to import" messages once I had the proper dependencies installed.
> 
> ```python
> for finder, name, ispkg in iter_namespace(plugins.decoders):
>     try:
>         importlib.import_module(name)
>         print(f'{name} imported successfully')
>     except (ModuleNotFoundError, ImportError) as e:
>         print(f'failed to import {name}, {e}')
>         pass
> ```

To followup on this, I think we're both correct here at different levels.

Runtime you're right. import cudaq_qec on main doesn't NameError; the loader runs. from .plugins import decoders, codes cause cudaq_qec.plugins to be attached to cudaq_qec as a side effect ([docs](https://docs.python.org/3/reference/import.html#submodules)), and since __init__.py's globals are cudaq_qec.__dict__, plugins resolve at runtime.

Source-level, the line still reads like a bug, which is why @kvmto flagged it the same way and why Ruff (F821) and Pyright report plugins as undefined here. So, the rename to iter_namespace(decoders) / iter_namespace(codes) is best read same runtime behavior, but using the names the from-import actually binds.

bmhowe23:
> Source-level, the line still reads like a bug, which is why @kvmto flagged it the same way ...

Kevin's comment
> this fix is also turning on a previously-dead code path, and I'd want to be sure that doesn't have side effects (previously-unloaded plugins now registering, decoder-name collisions, etc.).

OK, I think based on your comments and my experiment, we don't have to worry about side effects here because this it was actually working on `main` before, and this just makes it more clear. Thanks!

bmhowe23 (approved):

PR #523, opened 2026-04-29
[CUDRR1FA-393] Remove usage of CUDA-Q library mode

tlshannon:
Remove set(CUDAQ_LIBRARY_MODE ON).  Also need to remove find_package(CUDAQ) because this calls enable_language(CUDAQ), which associates all .cpp/.cxx files with the CUDAQ compiler instead of CXX.  Import specific CUDAQ targets instead.


boschmitt (approved):
LGTM

bmhowe23 (approved):

PR #524, opened 2026-04-30
[CUDRR1FA-374] Extend `trt_decoder` with global decoder chaining

bmhowe23:
Add a "predecoder" execution mode to the TensorRT decoder so it can be chained with a second decoder (e.g. PyMatching) and return logical-frame observables directly. The TRT model is assumed to emit a single output that concatenates [pre_L (num_observables entries), residual_dets (rest)].

New constructor parameters:
- "batch_size": required when the ONNX model has a dynamic batch dim. Used to size the optimization profile and pre-allocate I/O buffers.
- "global_decoder" + "global_decoder_params": optional decoder name and params for a follow-up decoder run on the residual_dets portion of the TRT output. Created with the same H passed to trt_decoder.
- "O": observables matrix (num_observables x block_size). Enables decode()/decode_batch() to return the predicted logical frame. Number of observables is inferred from O.shape()[0].

```
Decode behavior matrix:
- no global_decoder, no O   -> raw TRT output (unchanged).
- no global_decoder, O      -> return the pre_L prefix only.
- global_decoder, no O      -> entire output -> global_decoder.result.
- global_decoder, O         -> residual -> global_decoder; return
                               pre_L XOR global_decoder.logical_frame.
```

Other changes:
- Dynamic batch support: setInputShape per call when the model's batch dim is -1; ONNX builder now installs a min/opt/max optimization profile when "batch_size" is provided.
- Split decode_batch into a typed decode_batch_impl<float|uint8_t> for cleaner dtype dispatch (engine I/O dtypes float32 / uint8 unchanged).


test_surface_code_trt.py
```diff
@@ -0,0 +1,91 @@
+# ============================================================================ #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# This is a draft test script that should be improved and run by the CI (where possible).
```
bmhowe23: This test needs to be moved to a valid spot, fetch real data from somewhere, and made more thorough.

wsttiger: Replaced by test_trt_decoder_composite.cpp

wsttiger (approved):
We've discussed this exhaustively

PR #531, opened 2026-05-06
Update libraries for upstream CUDA-Q changes (LLVM22)

bmhowe23:
## Summary

This PR updates the CUDA-Q Libraries to build cleanly with the LLVM 22 toolchain and aligns the CI, container, and wheel build paths with the GCC 12-based CUDA-Q environment. It also reduces drift from upstream CUDA-Q build logic.

## What changed

- Updated CMake configuration to default to LLVM `22.1` and use the full major/minor version when locating LLVM.
- Switched CI jobs, wheel builds, helper scripts, docker images, cache keys, and workflow toolchain settings from `gcc-11` to `gcc-12`.
- Scoped GCC12-specific flags such as `--param=evrp-mode=legacy` and `-Wno-attributes` so they only apply to the relevant C/C++ compilation paths.
- Bumped the pinned CUDA-Q / CUDA-Q Realtime refs and updated the `get-cudaq-build` action to delegate to upstream `cudaq/scripts/build_cudaq.sh`, while preserving CUDA-QX-specific defaults.
- Added parallel `cmake --build -j` usage in CI helper scripts and pinned `cmake<4` where needed for compatibility.
- Matched CUDA-Q's nanobind configuration by using the CUDA-Q nanobind domain and pinning the version expected by the current upstream integration. Note: CUDA-Q rolledback nanobind from 2.12.0 to 2.9.2 as part of the LLVM update. It's not clear if this was intentional or not, but if they decide to roll back to 2.12.0, we can revert a0724f6, which was only added to account for this change.

**Note: wheel builds do not work yet. Those will be handled with a follow-up PR.**

### DIFF
--- a/cmake/Modules/CUDA-QX.cmake
+++ b/cmake/Modules/CUDA-QX.cmake
@@ -296,6 +296,6 @@ cudaqx_add_pymodule
 
 #]=======================================================================]
 function(cudaqx_add_pymodule module)
-  nanobind_add_module(${module} ${ARGN})
+  nanobind_add_module(${module} NB_DOMAIN cudaq ${ARGN})
   add_dependencies(cudaqx-pymodules ${module})
 endfunction()
--- a/cmake/Modules/CUDA-QX.cmake
+++ b/cmake/Modules/CUDA-QX.cmake
@@ -194,27 +194,6 @@ function(cudaqx_import_cudaq_targets)
     include("${CUDAQ_CMAKE_DIR}/CUDAQTargets.cmake")
   endif()
 
-  # CUDA-Q built with GCC 12 baked `--param=evrp-mode=legacy` into the
-  # PUBLIC compile options of `cudaq-common` (see runtime/common/CMakeLists.txt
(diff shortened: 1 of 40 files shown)


cmake/Modules/CUDA-QX.cmake
```diff
@@ -194,6 +194,27 @@ function(cudaqx_import_cudaq_targets)
     include("${CUDAQ_CMAKE_DIR}/CUDAQTargets.cmake")
   endif()
 
+  # CUDA-Q built with GCC 12 baked `--param=evrp-mode=legacy` into the
```
bmhowe23: @cketcham2333 this is what I would like to remove, if possible, by making an upstream change to CUDA-Q.

cketcham2333: You should be able to remove this if you incorporate https://github.com/NVIDIA/cuda-quantum/pull/4477

bmhowe23: Fixed in 3d356532f8b7953a5688335d24a0911d0220feb0

anjbur (approved):
LGTM!

PR #536, opened 2026-05-08
Expand TRT decoder YAML config for composite decoding [depends on PR #524]

wsttiger:
Add YAML/config support for TRT decoder runtime options including batch size,
CUDA graph execution, global decoder selection, and PyMatching-specific global
decoder parameters. Wire realtime decoder construction so TRT configs receive
the top-level observable matrix from O_sparse, and pass the same O matrix into
PyMatching global decoder params for composite observable decoding.

Expose the new config fields through Python bindings and heterogeneous_map
round-tripping. Extend YAML tests for TRT config round-trip, runtime parameter
conversion, and O_sparse-to-O injection.

Update test_trt_decoder_composite to support an optional --config-yaml path,
allowing the existing composite demo to construct and run a real TRT+PyMatching
decoder directly from YAML while preserving the original manual CLI path.


test_surface_code_trt.py
```diff
@@ -0,0 +1,91 @@
+# ============================================================================ #
```
melody-ren: This file sits at the root of the repo. I'm not sure if it's intentional? There are also several hardcoded paths used in this file. From the comment, it seems like this file is meant to be a draft. Is it ready for review or should it be removed from the PR?

libs/qec/lib/realtime/config.cpp
```diff
@@ -197,6 +240,23 @@ cudaqx::heterogeneous_map trt_decoder_config::to_heterogeneous_map() const {
...
+          "global_decoder_params currently supports only pymatching.");
+    }
+    config_map.insert(
+        "global_decoder_params",
+        global_decoder_config_to_heterogeneous_map(global_decoder_params));
+  } else if (global_decoder.has_value()) {
+    // trt_decoder attaches a global decoder only when both the decoder name and
+    // parameter map are present. An empty map is valid for decoders that do not
+    // need extra parameters.
+    config_map.insert("global_decoder_params", cudaqx::heterogeneous_map());
```
melody-ren: I think this empty `global_decoder_params` insertion should move out of `trt_decoder_config::to_heterogeneous_map()`.

At the config layer, `global_decoder = "pymatching"` with `global_decoder_params = std::monostate{}` means “use pymatching with default params.” But this serializer turns that state into a map containing `global_decoder_params: {}`. If that heterogeneous map is later deserialized, the presence of the key causes `from_heterogeneous_map()` to materialize `pymatching_decoder_config{}` instead of preserving monostate. So a config-level round trip changes the variant alternative even though the user did not specify params.

The TRT runtime may need an empty params map as an attachment marker, but that is runtime plumbing rather than faithful config serialization. I think `to_heterogeneous_map()` should only emit `global_decoder_params` when params were explicitly configured, and `prepare_decoder_params()` should synthesize the empty runtime map when `global_decoder == "pymatching"` and no params map exists.

libs/qec/lib/realtime/config.cpp
```diff
@@ -209,6 +269,26 @@ trt_decoder_config trt_decoder_config::from_heterogeneous_map(
   GET_ARG(engine_save_path);
   GET_ARG(precision);
   GET_ARG(memory_workspace);
+  GET_ARG(batch_size);
+  GET_ARG(use_cuda_graph);
+  GET_ARG(global_decoder);
+  if (map.contains("global_decoder_params")) {
```
melody-ren, line 281: I'm not sure if it's already checked. But if global decoder config exists but the decoder itself is unset, we should reject it

libs/qec/lib/realtime/realtime_decoding.cpp
```diff
@@ -42,6 +43,35 @@ static std::vector<uint8_t> pack_syndrome_bits(const uint8_t *syndromes,
...
+    return params;
+
+  const auto num_observables = std::count(decoder_config.O_sparse.begin(),
+                                          decoder_config.O_sparse.end(), -1);
+  if (num_observables == 0)
+    return params;
+
+  auto O = cudaq::qec::pcm_from_sparse_vec(
+      decoder_config.O_sparse, num_observables, decoder_config.block_size);
+  params.insert("O", O);
```
melody-ren, line 222: Note: injecting `O` here means the TRT decoder may return observable corrections with length `num_observables` instead of an error vector with length `block_size`. The current realtime `enqueue_syndrome()` path still validates decoder results against `block_size`, so this integration path will throw until the `enqueue_syndrome` path understands observable-returning decoders. I'm patching that separately, but I wanted to flag the dependency because this line is what enables that mode for realtime configs.

melody-ren:
The following were added on top of 26be6b48c450861c5a5631f7f9ec59a3c56425bc:

- `decode_to_errs`: `decode().result` is an error vector; `enqueue_syndrome()` projects it through `O_sparse`.
- `decode_to_obs`: `decode().result` is already an observable vector; `enqueue_syndrome()` applies it directly.

`enqueue_syndrome()` now validates result size against the selected result type, applies corrections through an explicit switch, and logs the result type plus either `Errors` or `Observables`. This avoids treating obs-frame decoder output as error indices.

- Added explicit decoder result-type handling for realtime replay/runtime paths.
  The TRT composite decoder returns observable-frame results, not error-frame results, so callers should not infer meaning from “not errors” or from vector shape. The runtime now uses the decoder result type to decide whether to project errors through `O` or apply observable results directly.

- Expanded decoder stats logging to include `ResultType`, `Errors`, and `Observables`.
  This makes log replay able to tell what `decode().result` meant at the time it was produced, which is required for replaying obs-frame decoders correctly.

- Fixed `replay_decoder_logs.py` to interpret logged result types explicitly.
  Replay now handles `errs` as an error vector and `obs` as an observable vector. It no longer assumes “non-errs means obs”; unknown result-type sets fail clearly instead of being silently misinterpreted.

- Fixed TRT `global_decoder_params` round-tripping.
  `global_decoder_params` is now represented as `variant<monostate, pymatching_decoder_config>`. Serialization omits it for `monostate`, so a config with no params does not round-trip into `global_decoder_params: {}` and come back as default PyMatching params.

- Moved runtime-only TRT parameter synthesis into `prepare_decoder_params()`.
  Serialization stays faithful to the config, while runtime still synthesizes an empty `global_decoder_params` map for `global_decoder: pymatching` when the plugin needs that key to attach the global decoder. This also happens before the no-`O_sparse` early return, so the no-O path still attaches PyMatching correctly.

- Added validation for heterogeneous-map `global_decoder_params` without `global_decoder`.
  That catches the write/read path where params are present but there is no decoder name to interpret them.

- Added/consolidated tests around TRT YAML round-trip, heterogeneous-map conversion, observable matrix injection, monostate global decoder params, no-O runtime prep, and params-without-decoder errors.

libs/qec/lib/realtime/realtime_decoding.cpp
melody-ren: Resolved in 1dff1f1df628654d03c9cddf4dc7d93902b101ab

libs/qec/lib/realtime/config.cpp
melody-ren: Resolved in 626e40d7a66cde20de96bfd52d39041225b33535

libs/qec/lib/realtime/config.cpp
melody-ren: Resolved in 626e40d7a66cde20de96bfd52d39041225b33535

wsttiger (commented):
Overall this is a solid, well-tested change. The `decode_to_errs` / `decode_to_obs` abstraction is clean, the serialization monostate handling is carefully thought through (and documented in-code), and the test coverage is genuinely good — YAML round-trip, the monostate edge cases, `prepare_decoder_params`, and `enqueue_syndrome` routing all have focused tests. Comments below are mostly one must-fix plus robustness/clarity nits.

## Blocking

### 1. `test_surface_code_trt.py` should not be committed (repo root, +91)
This is a draft scratch script at the repository root, with a hardcoded absolute path (`/workspaces/pre-decoder/circuit_Z.stim`), commented-out blocks, and a `beliefmatching` dependency. Its own header says it's a draft that "should be improved and run by the CI" — it won't run in CI and doesn't belong in the tree. Please remove it, or if real coverage is intended, move it under `libs/qec/python/tests/` with fixtures and no absolute paths.

## Correctness / robustness

### 2. `decode_to_obs` branch has no empty-result guard for sliding windows (`libs/qec/lib/decoder.cpp`)
The size validation explicitly permits an empty result in sliding-window mode (`is_sliding_window && !result.empty() && ...`). But the new obs-frame branch then indexes unconditionally:

```cpp
case decode_result_type::decode_to_obs:
  for (std::size_t i = 0; i < num_observables; i++)
    if (decoded_result.result[i]) { ... }
```

If a `decode_to_obs` decoder is ever used in sliding-window mode and returns an empty result (the case validation just allowed), this is an out-of-bounds read. The `decode_to_errs` path inherits the same pre-existing pattern so it's not a regression, but since the obs branch is new code, a cheap guard is worth it:

```cpp
case decode_result_type::decode_to_obs:
  if (decoded_result.result.empty())
    break;  // sliding-window: nothing to apply yet
  ...
```

Today's only `decode_to_obs` user (composite TRT) isn't sliding-window, so this is edge-case hardening rather than an active bug.

### 3. Triple-nested `catch (...)` cascade in `trt_decoder_config::from_heterogeneous_map` (`libs/qec/lib/realtime/config.cpp`)
The three-layer `try { get<global_decoder_config> } catch(...) { try { get<pymatching_config> } catch(...) { get<heterogeneous_map> ... } }` works and is tested, but catch-all on three layers will also swallow genuinely unexpected errors and silently fall through, making future failures hard to diagnose. If `heterogeneous_map` exposes a type-query or typed `try_get`, prefer dispatching on that; at minimum add a comment explaining the three representations being probed (variant blob vs. typed `pymatching_config` vs. raw nested map).

## Clarity / minor

### 4. Redundant dual string descriptors in `enqueue_syndrome`
`result_type_str` (`errs`/`obs`) and `result_type_name` (`decode_to_errs`/`decode_to_obs`) are two near-duplicate strings from the same switch, and the `if (!result_type_name)` sentinel is effectively dead since the switch covers both enum values. Consider collapsing to one descriptor, or add a `default:` and drop the sentinel.

### 5. Tests reach into `lib/` via relative includes
Both `test_decoders_yaml.cpp` and `test_trt_decoder_composite.cpp` `#include "../lib/realtime/realtime_decoding.h"` to reach the now-exported `prepare_decoder_params`. Acceptable for unit tests, but relative `../lib` includes are brittle if files move. Flagging for awareness.

### 6. `ResultType` parsed as a comma-set in the replay script (`libs/qec/utils/replay_decoder_logs.py`)
The log emits a single token (`errs`/`obs`), but the parser builds a set via `split(",")` and checks membership. It works, but modeling a scalar field as a set is slightly surprising — worth a one-line comment that this is intentional, or just treat it as a scalar.

## Things done well (no change needed)
- The monostate serialization rationale (avoid `global_decoder_params: {}` round-tripping into a default `pymatching_config`) is documented in the three places that matter and backed by `TrtDecoderMonostateGlobalDecoderParams`. This is the subtle part of the PR and it's handled carefully.
- `prepare_decoder_params` correctly synthesizes the empty params map *before* the `O_sparse` early-return, with an explanatory comment.
- Python `setter_accepts_none` is a real usability improvement for clearing `std::optional` fields.
- `enqueue_syndrome` size-mismatch error now reports the expected size and result-type name — much more debuggable than the old `block_size`-only message.
- The replay-script comment documenting that obs-frame replay needs TensorRT/GPU/artifacts is a great operational caveat.

## Note
Please confirm the `[depends on #524]` overlap so this doesn't double-land code, and re-scope the final diff once #524 merges.

libs/qec/lib/realtime/realtime_decoding.cpp
```diff
@@ -179,6 +180,45 @@ static std::vector<uint8_t> pack_syndrome_bits(const uint8_t *syndromes,
 
 namespace cudaq::qec::decoding::host {
 
+cudaqx::heterogeneous_map prepare_decoder_params(
+    const cudaq::qec::decoding::config::decoder_config &decoder_config) {
+  auto params = decoder_config.decoder_custom_args_to_heterogeneous_map();
+  if (decoder_config.type != "trt_decoder")
+    return params;
```
kvmto, line 187: batch_size on the realtime single-shot path. The realtime path decodes one syndrome per call (decoder.cpp:359), and with batch_size > 1 trt_decoder::decode() zero-pads to a full batch and discards all but slot 0 (trt_decoder.cpp:835-857). Since this PR exposes batch_size through the realtime config, is batch_size > 1 ever meaningful for realtime, or should we pin it to 1 or CUDAQ_WARN here so a config cannot silently Nx per-syndrome GPU work on the latency-critical path?

libs/qec/lib/realtime/config.cpp
```diff
@@ -190,6 +190,31 @@ single_error_lut_config single_error_lut_config::from_heterogeneous_map(
...
+global_decoder_config global_decoder_config_from_heterogeneous_map(
+    const cudaqx::heterogeneous_map &map,
+    const std::optional<std::string> &global_decoder) {
+  if (global_decoder.has_value() && global_decoder.value() != "pymatching") {
+    throw std::runtime_error(
+        "global_decoder_params currently supports only pymatching.");
+  }
+
+  return pymatching_config::from_heterogeneous_map(map);
+}
```
kvmto, line 216: Is the monostate no-params invariant symmetric on read? Serialization deliberately never emits global_decoder_params: {} for monostate, but this returns a default pymatching_config for an empty map, never monostate. The YAML input side (config.cpp:456-458) behaves the same way. A hand-written global_decoder: pymatching with global_decoder_params: {} therefore deserializes into a real default PyMatching decoder, which is the exact mutation the write side guards against. Should an empty nested map read back as monostate so the invariant is symmetric for hand-authored configs?

libs/qec/lib/decoder.cpp
```diff
@@ -362,36 +364,76 @@ bool decoder::enqueue_syndrome(const uint8_t *syndrome,
...
+    case decode_result_type::decode_to_errs:
+      result_type_str = "errs";
+      result_type_name = "decode_to_errs";
+      expected_result_size = block_size;
+      break;
+    case decode_result_type::decode_to_obs:
+      result_type_str = "obs";
+      result_type_name = "decode_to_obs";
+      expected_result_size = num_observables;
+      break;
```
kvmto, line 384: Should the decode_to_obs contract be enforced, not just documented? Here expected_result_size == get_num_observables() == O_sparse.size(). If a decode_to_obs decoder never calls set_O_sparse(), then num_observables == 0, so an empty result passes validation, the obs loop (lines 413-421) runs zero times, and enqueue_syndrome returns true, which is a silent no-op. Realtime always sets O_sparse, but this is a public base class. Would a one-time decode_to_obs implies num_observables > 0 check be worth it to turn a silent wrong-answer into a clear error?

libs/qec/lib/decoder.cpp
```diff
@@ -362,36 +364,76 @@ bool decoder::enqueue_syndrome(const uint8_t *syndrome,
...
+    case decode_result_type::decode_to_obs:
+      // Observable-frame path: decoder already projected to observables via its
+      // internal "O" matrix; use the result directly.
+      for (std::size_t i = 0; i < num_observables; i++)
+        if (decoded_result.result[i]) {
           if (should_log)
-            log_observable_corrections[i] ^= 1;
+            log_observables.push_back(i);
+          flip_correction(i);
         }
```
kvmto, line 421: Is "obs frames must be hard 0/1" an invariant we should pin down? This uses truthiness (!= 0.0), while the replay tool thresholds at > 0.5 (replay_decoder_logs.py:235-237). They agree only because the composite emits strictly 0.0f or 1.0f. If a future obs-frame decoder emits soft values, live and replay will silently disagree. Should we assert or document that obs-frame results must be hard 0/1?

libs/qec/python/bindings/type_casters.h
```diff
@@ -194,6 +194,17 @@ struct type_caster<cudaqx::heterogeneous_map> {
...
+                       cudaq::qec::decoding::config::global_decoder_config>(
+                       &val)) {
+          if (std::holds_alternative<std::monostate>(*global_cfg)) {
+            result[key.c_str()] = nb::none();
+          } else {
+            result[key.c_str()] = nb::cast(
+                std::get<cudaq::qec::decoding::config::pymatching_config>(
+                    *global_cfg)
+                    .to_heterogeneous_map());
+          }
```
kvmto, line 209: Should the monostate encoding be consistent across layers? to_heterogeneous_map() omits the key entirely for monostate, but this caster emits key: None. Is the divergence between absent and None intentional, and could a Python consumer iterating the dict trip over the None? Should these two monostate encodings be aligned?

libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -329,6 +356,9 @@ class decoder
 
   /// @brief The decoder's D matrix in sparse format
   std::vector<std::vector<uint32_t>> D_sparse;
+
+private:
+  decode_result_type result_type_ = decode_result_type::decode_to_errs;
```
kvmto, line 361: Is the new data member on the public decoder base an ABI concern? This adds a member to cudaq::qec::decoder, an installed header and documented plugin extension point. In-repo everything rebuilds, but any prebuilt out-of-tree decoder plugin would need recompilation. Is that acceptable for this release, or should we note it in the release notes?

libs/qec/utils/replay_decoder_logs.py
```diff
@@ -181,28 +205,60 @@ def parse_decoder_config(config_file, decoders, O_per_decoder):
...
+        ]
+    elif "obs" in result_types:
+        replay_errors_sparse.append([])
+        O_replay = numpy.array([1 if x > 0.5 else 0 for x in result.result],
+                               dtype=numpy.uint8)
+        decoded_observables_sparse = decoded_sparse
+    else:
+        print(f"Error: unsupported ResultType set {sorted(result_types)} "
+              f"in decode_call_idx {decode_call_idx}.")
+        exit(1)
```
kvmto, line 241: How do we verify the obs-frame replay path? This obs branch only runs once the composite decoder is reconstructed, which needs TensorRT, a GPU, and the ONNX or engine artifacts, so it cannot execute in CI. Could we exercise it with a lightweight fake obs-frame decoder using the sample_decoder trick so the path does not silently rot?

kvmto (commented):
Heads up, I have not been through all of it yet. :) open to discuss or drop the comments

libs/qec/include/cudaq/qec/decoder.h
melody-ren: We generally expect the C++ users to rebuild against our latest main/release so this is ok

libs/qec/lib/realtime/config.cpp
melody-ren: > A hand-written global_decoder: pymatching with global_decoder_params: {} therefore deserializes into a real default PyMatching decoder, which is the exact mutation the write side guards against. 

This is intentional and not something we guard against. A `{}` for the config param means "use default" and that is what it gets serializes into, a default pymatching config. What we do guard against in the heterogeneous_map layer is this combo: global_decoder unset (meaning, this key is not in the map) but a global decoder config is provided. We throw in this case because we do not want to make assumptions about what the default global decoder should be. This is somewhat debatable if we should just allow this combo though. 

Regarding the emitting or not emitting monostate part: if global_decoder is unset and no global config param is provided, then the missing global config param should serialize and deserialize to monostate, which means "none is provided". What we do not want is to silently swapping in a pymatching config after monostate is round tripped back.

libs/qec/lib/realtime/realtime_decoding.cpp
melody-ren: Good call. Will add a warning here when batch_size > 1

libs/qec/lib/decoder.cpp
melody-ren: For offline mode: we can't really enforce it here with a one time check because when set_O_sparse is called, or if it's even called at all, is largely decoder dependent. In the offline mode, each decoder needs to decide if/when it sets O. I admit this is non-ideal and puts the onus on the decoder authors to do the right thing. I'm open to suggestions though

For realtime path: set_O_sparse is always called and the silent no-op case is not reachable

libs/qec/lib/decoder.cpp
melody-ren: This is deferred for a later date when we do support soft results.

test_surface_code_trt.py
melody-ren: Agreed that we remove this file (this was an early draft for experimentation), per discussion with @wsttiger

libs/qec/python/bindings/type_casters.h
melody-ren: No, this was not intentional and yes, I do agree that they should be consistent. It's a one line change so I'll include it in this PR

libs/qec/utils/replay_decoder_logs.py
melody-ren: obs return path is tested in the cpp version, using the sample_decoder. The thing is, that decoder doesn't have a python version. We can however use a stub python decoder to test the replay path. I'll log that as a follow-up PR and not a blocker for this one though.

libs/qec/lib/realtime/realtime_decoding.cpp
melody-ren: Added in 321ce20f441dfe5dc373d7d745fa96adb2c6d4c6

libs/qec/python/bindings/type_casters.h
melody-ren: Addressed in 321ce20f441dfe5dc373d7d745fa96adb2c6d4c6

test_surface_code_trt.py
melody-ren: Removed in 32b491407124cef63420709fcd2946aba1c478a2

libs/qec/lib/realtime/realtime_decoding.cpp
kvmto: great, thanks!

libs/qec/lib/realtime/config.cpp
kvmto: Thanks, that distinction makes sense. No change needed, appreciate the clarification.

libs/qec/lib/decoder.cpp
kvmto: after carefully looking at it, I agree. To be clear, I only meant a guard in enqueue_syndrome (the realtime path), not decode(), so it wouldn't touch offline decoders. But watching the wiring, decode_to_obs and O_sparse are both driven by decoder_config.O_sparse, I think realtime can't actually hit this. So, happy to drop it.

libs/qec/python/bindings/type_casters.h
kvmto: great, thanks!

libs/qec/include/cudaq/qec/decoder.h
kvmto: perfect, thanks!

melody-ren (approved):

PR #537, opened 2026-05-08
Fix for CI error when running Python example

anjbur:
Fix for recent [CI failures](https://github.com/NVIDIA/cudaqx/actions/runs/25504210124/job/74888675567?pr=535).


bmhowe23 (approved):
Thanks, @anjbur!

PR #541, opened 2026-05-12
Fix failing test in Python wheels

anjbur:
See test failure [here](https://github.com/NVIDIA/cudaqx/actions/runs/25698909898).

Python packages for TensorRT are installed in `test_wheels.sh`, but only installed for C++ in the general PR workflow. Because of this, Python tests that relied on TensorRT were skipped in PR checks, and the above test didn't fail until the Python wheels were tested. This PR adds Python TensorRT installation to the general PR workflow, meaning this test will no longer be skipped there. It also updates the failing test to match the new behavior from #524.


bmhowe23 (approved):
Thanks, @anjbur!

PR #550, opened 2026-05-18
Sparse parity-check matrix support for decoders

### LINKED ISSUE #379: cudaqx::tensor fails to allocate large PCMs

vedika-saravanan:
Continuation of [#440](https://github.com/NVIDIA/cudaqx/pull/440). Closes [#379](https://github.com/NVIDIA/cudaqx/issues/379).

Introduces `cudaq::qec::sparse_binary_matrix` (CSC/CSR, `uint32_t` indices) as the decoder PCM type, so large PCMs no longer have to be materialized as a dense `cudaqx::tensor`. The sparse path is wired through PCM generation, round slicing, decoder construction, and Python bindings. `sliding_window` canonicalizes once and reads `H.ptr() / H.indices()` directly per window.

## Breaking changes

- **Decoder constructor** now takes `const sparse_binary_matrix &H`. Dense `cudaqx::tensor<uint8_t>` callers still compile via implicit conversion, but the extension-point template changed — **out-of-tree plugins must be recompiled**.
- **`generate_random_pcm` throws above `k_max_dense_pcm_elements` (400M)**. Use `generate_random_pcm_sparse` for larger matrices.
- **`single_error_lut_example`** now uses column `qErr` of H directly instead of XOR-of-other-columns (the old form was only correct when every row of H had even weight). Example plugin only; production `single_error_lut` was already correct.

## Follow-up PRs

- Sparse-aware PCM utilities: migrate sort_pcm_columns, simplify_pcm, reorder_pcm_columns, shuffle_pcm_columns, pcm_extend_to_n_rounds, pcm_to_sparse_*, pcm_from_sparse_*, and detector_error_model::canonicalize_for_rounds to sparse_binary_matrix so the full #379 scale pipeline avoids any dense cudaqx::tensor<uint8_t> allocation.
- Docs: Sphinx entries for the new type/functions and the 400M dense cap.

Note to reviewers: Quick context on why it's larger than the ~1000-line guideline: it's a foundational type-system change rather than an additive feature. The new sparse_binary_matrix type, the decoder constructor breaking change, all the plugin updates, the sparse generator, and the Python bindings all have to land together for the #379 fix to actually be usable end-to-end. Splitting it would create intermediate states where the breaking change has landed but users still can't construct a >50k-mechanism PCM which would mean shipping the API churn without the benefit.

### DIFF
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -1,5 +1,5 @@
 /****************************************************************-*- C++ -*-****
- * Copyright (c) 2025 NVIDIA Corporation & Affiliates.                         *
+ * Copyright (c) 2025 - 2026 NVIDIA Corporation & Affiliates.                  *
  * All rights reserved.                                                        *
  *                                                                             *
  * This source code and the accompanying materials are made available under    *
@@ -109,6 +109,14 @@ get_sorted_pcm_column_indices(const cudaqx::tensor<uint8_t> &pcm,
 /// @param num_syndromes_per_round The number of syndromes per round.
 /// @return True if the PCM is sorted, false otherwise.
 bool pcm_is_sorted(const cudaqx::tensor<uint8_t> &pcm,
+                   std::uint32_t num_syndromes_per_round = 0);
+
+/// @brief Check if a PCM is sorted.
+/// @param sparse_pcm The sparse PCM to check (in the same format as
+/// get_sparse_pcm())
+/// @param num_syndromes_per_round The number of syndromes per round.
+/// @return True if the PCM is sorted, false otherwise.
+bool pcm_is_sorted(const std::vector<std::vector<std::uint32_t>> &sparse_pcm,
                    std::uint32_t num_syndromes_per_round = 0);
 
 /// @brief Reorder the columns of a PCM according to the given column order.
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -187,12 +187,19 @@ get_pcm_for_rounds(const sparse_binary_matrix &pcm,
                    bool straddle_start_round = false,
                    bool straddle_end_round = false);
 
+/// @brief Upper bound on \f$\texttt{rows} \times \texttt{cols}\f$ for the
+/// dense \c generate_random_pcm path. Above this, the call throws and the
+/// caller should switch to \c generate_random_pcm_sparse. Exposed as a named
+/// symbol so tests and users can reference it without grepping the source.
+inline constexpr std::size_t k_max_dense_pcm_elements =
+    static_cast<std::size_t>(400u) * 1024u * 1024u;
+
 /// @brief Generate a random PCM with the given parameters.
 ///
-/// If \f$\texttt{rows} \times \texttt{cols}\f$ exceeds an internal dense limit
-/// (currently 400 million entries), throws \c std::invalid_argument; use
-/// \c generate_random_pcm_sparse for matrices that large without a dense
-/// allocation.
+/// If \f$\texttt{rows} \times \texttt{cols}\f$ exceeds \c
+/// k_max_dense_pcm_elements (currently ~419 million entries), throws \c
+/// std::invalid_argument; use \c generate_random_pcm_sparse for matrices that
+/// large without a dense allocation.
 /// @param n_rounds The number of rounds in the PCM.
 /// @param n_errs_per_round The number of errors per round in the PCM.
 /// @param n_syndromes_per_round The number of syndromes per round in the PCM.
@@ -212,6 +219,20 @@ sparse_binary_matrix
 generate_random_pcm_sparse(std::size_t n_rounds, std::size_t n_errs_per_round,
                            std::size_t n_syndromes_per_round, int weight,
                            std::mt19937_64 &&rng);
+
+/// @brief Return a GF(2)-canonical copy of \p pcm: each column (CSC) or row
+/// (CSR) has its indices sorted ascending, and duplicate indices are XOR-
+/// merged — an index that appears \p k times is kept iff \p k is odd. The
+/// output has the same layout as the input and is idempotent under further
+/// \c canonicalize_pcm calls.
+///
+/// Use this before passing a PCM built from a DEM decomposition (or any other
+/// source that may legitimately emit duplicate indices within a column/row)
+/// to consumers like PyMatching that dispatch on per-column nnz count and
+/// assume at-most-one entry per row per column. \c sparse_binary_matrix
+/// itself accepts duplicates at construction (see its header) so a separate
+/// call is required to canonicalize.
+sparse_binary_matrix canonicalize_pcm(const sparse_binary_matrix &pcm);
 
 /// @brief Randomly permute the columns of a PCM.
 /// @param pcm The PCM to permute.
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -178,6 +178,23 @@ get_pcm_for_rounds(const cudaqx::tensor<uint8_t> &pcm,
 /// reads from \p pcm as ``sparse_binary_matrix`` so the full dense PCM is not
 /// required (only the returned sub-matrix is dense). Parameter meanings match
 /// the dense overload.
+///
+/// @param pcm_is_canonical If true, the caller asserts \p pcm has
+/// sorted-unique per-group indices (i.e. is the output of \c canonicalize_pcm
+/// or was constructed canonically). In that case the per-call
+/// canonicalization step is skipped — useful for callers like
+/// \c sliding_window that canonicalize once at construction and then call
+/// this function in a per-window loop on the same matrix. Default \c false
+/// preserves the original "canonicalize on entry" behavior.
+///
+/// @warning When \p pcm_is_canonical is \c true the precondition is *not*
+/// checked. select_pcm_columns_for_round_range reads \c .front() / \c .back()
+/// of each CSC column to derive first/last round; those are only the true
+/// min/max row if the column list is sorted-unique. Passing a non-canonical
+/// PCM with this flag (e.g. a raw DEM decomposition, a hand-built sparse
+/// matrix with duplicate or unsorted indices, or a canonical CSR that the
+/// caller forgot to re-canonicalize after conversion) silently produces
+/// wrong round assignments. If unsure, leave the flag \c false.
 /// @return A tuple with the new PCM with the columns in the range [start_round,
 /// end_round], the first column included, and the last column included.
 std::tuple<cudaqx::tensor<uint8_t>, std::uint32_t, std::uint32_t>
@@ -185,7 +202,8 @@ get_pcm_for_rounds(const sparse_binary_matrix &pcm,
                    std::uint32_t num_syndromes_per_round,
                    std::uint32_t start_round, std::uint32_t end_round,
                    bool straddle_start_round = false,
-                   bool straddle_end_round = false);
+                   bool straddle_end_round = false,
+                   bool pcm_is_canonical = false);
 
 /// @brief Upper bound on \f$\texttt{rows} \times \texttt{cols}\f$ for the
 /// dense \c generate_random_pcm path. Above this, the call throws and the
@@ -214,7 +232,8 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
 
 /// @brief Same distribution as generate_random_pcm, but constructs a CSC
 /// sparse_binary_matrix directly without allocating a dense rank-2 tensor.
-/// Intended for large PCMs where dense allocation fails (#379 style sizes).
+/// Intended for large PCMs whose dense form would exceed
+/// k_max_dense_pcm_elements.
 sparse_binary_matrix
 generate_random_pcm_sparse(std::size_t n_rounds, std::size_t n_errs_per_round,
                            std::size_t n_syndromes_per_round, int weight,
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -191,7 +191,8 @@ get_pcm_for_rounds(const sparse_binary_matrix &pcm,
 ///
 /// If \f$\texttt{rows} \times \texttt{cols}\f$ exceeds an internal dense limit
 /// (currently 400 million entries), throws \c std::invalid_argument; use
-/// \c generate_random_pcm_sparse for matrices that large without a dense allocation.
+/// \c generate_random_pcm_sparse for matrices that large without a dense
+/// allocation.
 /// @param n_rounds The number of rounds in the PCM.
 /// @param n_errs_per_round The number of errors per round in the PCM.
 /// @param n_syndromes_per_round The number of syndromes per round in the PCM.
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -8,6 +8,7 @@
 #pragma once
 
 #include "cuda-qx/core/tensor.h"
+#include "sparse_binary_matrix.h"
 #include <limits>
 #include <random>
 
@@ -185,6 +186,14 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
                                             std::size_t n_errs_per_round,
                                             std::size_t n_syndromes_per_round,
                                             int weight, std::mt19937_64 &&rng);
+
+/// @brief Same distribution as generate_random_pcm, but constructs a CSC
+/// sparse_binary_matrix directly without allocating a dense rank-2 tensor.
+/// Intended for large PCMs where dense allocation fails (#379 style sizes).
+sparse_binary_matrix
+generate_random_pcm_sparse(std::size_t n_rounds, std::size_t n_errs_per_round,
+                           std::size_t n_syndromes_per_round, int weight,
+                           std::mt19937_64 &&rng);
 
 /// @brief Randomly permute the columns of a PCM.
 /// @param pcm The PCM to permute.
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -114,7 +114,7 @@ bool pcm_is_sorted(const cudaqx::tensor<uint8_t> &pcm,
 
 /// @brief Check if a PCM is sorted.
 /// @param sparse_pcm The sparse PCM to check (in the same format as
-/// get_sparse_pcm())
+/// dense_to_sparse())
 /// @param num_syndromes_per_round The number of syndromes per round.
 /// @return True if the PCM is sorted, false otherwise.
 bool pcm_is_sorted(const std::vector<std::vector<std::uint32_t>> &sparse_pcm,
@@ -180,21 +180,21 @@ get_pcm_for_rounds(const cudaqx::tensor<uint8_t> &pcm,
 /// the dense overload.
 ///
 /// @param pcm_is_canonical If true, the caller asserts \p pcm has
-/// sorted-unique per-group indices (i.e. is the output of \c canonicalize_pcm
+/// sorted-unique per-group indices (i.e. is the output of `canonicalize_pcm`
 /// or was constructed canonically). In that case the per-call
 /// canonicalization step is skipped — useful for callers like
-/// \c sliding_window that canonicalize once at construction and then call
-/// this function in a per-window loop on the same matrix. Default \c false
+/// `sliding_window` that canonicalize once at construction and then call
+/// this function in a per-window loop on the same matrix. Default `false`
 /// preserves the original "canonicalize on entry" behavior.
 ///
-/// @warning When \p pcm_is_canonical is \c true the precondition is *not*
-/// checked. select_pcm_columns_for_round_range reads \c .front() / \c .back()
+/// @warning When \p pcm_is_canonical is `true` the precondition is *not*
+/// checked. `select_pcm_columns_for_round_range` reads `.front()` / `.back()`
 /// of each CSC column to derive first/last round; those are only the true
 /// min/max row if the column list is sorted-unique. Passing a non-canonical
 /// PCM with this flag (e.g. a raw DEM decomposition, a hand-built sparse
 /// matrix with duplicate or unsorted indices, or a canonical CSR that the
 /// caller forgot to re-canonicalize after conversion) silently produces
-/// wrong round assignments. If unsure, leave the flag \c false.
+/// wrong round assignments. If unsure, leave the flag `false`.
 /// @return A tuple with the new PCM with the columns in the range [start_round,
 /// end_round], the first column included, and the last column included.
 std::tuple<cudaqx::tensor<uint8_t>, std::uint32_t, std::uint32_t>
@@ -206,18 +206,21 @@ get_pcm_for_rounds(const sparse_binary_matrix &pcm,
                    bool pcm_is_canonical = false);
 
 /// @brief Upper bound on \f$\texttt{rows} \times \texttt{cols}\f$ for the
-/// dense \c generate_random_pcm path. Above this, the call throws and the
-/// caller should switch to \c generate_random_pcm_sparse. Exposed as a named
+/// dense `generate_random_pcm` path. Above this, the call throws and the
+/// caller should switch to `generate_random_pcm_sparse`. Exposed as a named
 /// symbol so tests and users can reference it without grepping the source.
 inline constexpr std::size_t k_max_dense_pcm_elements =
     static_cast<std::size_t>(400u) * 1024u * 1024u;
 
 /// @brief Generate a random PCM with the given parameters.
 ///
-/// If \f$\texttt{rows} \times \texttt{cols}\f$ exceeds \c
-/// k_max_dense_pcm_elements (currently ~419 million entries), throws \c
-/// std::invalid_argument; use \c generate_random_pcm_sparse for matrices that
-/// large without a dense allocation.
+/// The PCM has shape `(n_rounds * n_syndromes_per_round) × (n_rounds *
+/// n_errs_per_round)`. If the product
+/// \f$n\_rounds^2 \times n\_syndromes\_per\_round \times n\_errs\_per\_round\f$
+/// exceeds `k_max_dense_pcm_elements`
+/// (\f$400 \times 1024 \times 1024 \approx 4.19 \times 10^8\f$ entries),
+/// throws `std::invalid_argument`; use `generate_random_pcm_sparse` for
+/// matrices that large without a dense allocation.
 /// @param n_rounds The number of rounds in the PCM.
 /// @param n_errs_per_round The number of errors per round in the PCM.
 /// @param n_syndromes_per_round The number of syndromes per round in the PCM.
@@ -243,12 +246,12 @@ generate_random_pcm_sparse(std::size_t n_rounds, std::size_t n_errs_per_round,
 /// (CSR) has its indices sorted ascending, and duplicate indices are XOR-
 /// merged — an index that appears \p k times is kept iff \p k is odd. The
 /// output has the same layout as the input and is idempotent under further
-/// \c canonicalize_pcm calls.
+/// `canonicalize_pcm` calls.
 ///
 /// Use this before passing a PCM built from a DEM decomposition (or any other
 /// source that may legitimately emit duplicate indices within a column/row)
 /// to consumers like PyMatching that dispatch on per-column nnz count and
-/// assume at-most-one entry per row per column. \c sparse_binary_matrix
+/// assume at-most-one entry per row per column. `sparse_binary_matrix`
 /// itself accepts duplicates at construction (see its header) so a separate
 /// call is required to canonicalize.
 sparse_binary_matrix canonicalize_pcm(const sparse_binary_matrix &pcm);
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -174,7 +174,24 @@ get_pcm_for_rounds(const cudaqx::tensor<uint8_t> &pcm,
                    bool straddle_start_round = false,
                    bool straddle_end_round = false);
 
+/// @brief Same semantics as the overload taking a dense tensor \p pcm, but
+/// reads from \p pcm as ``sparse_binary_matrix`` so the full dense PCM is not
+/// required (only the returned sub-matrix is dense). Parameter meanings match
+/// the dense overload.
+/// @return A tuple with the new PCM with the columns in the range [start_round,
+/// end_round], the first column included, and the last column included.
+std::tuple<cudaqx::tensor<uint8_t>, std::uint32_t, std::uint32_t>
+get_pcm_for_rounds(const sparse_binary_matrix &pcm,
+                   std::uint32_t num_syndromes_per_round,
+                   std::uint32_t start_round, std::uint32_t end_round,
+                   bool straddle_start_round = false,
+                   bool straddle_end_round = false);
+
 /// @brief Generate a random PCM with the given parameters.
+///
+/// If \f$\texttt{rows} \times \texttt{cols}\f$ exceeds an internal dense limit
+/// (currently 400 million entries), throws \c std::invalid_argument; use
+/// \c generate_random_pcm_sparse for matrices that large without a dense allocation.
 /// @param n_rounds The number of rounds in the PCM.
 /// @param n_errs_per_round The number of errors per round in the PCM.
 /// @param n_syndromes_per_round The number of syndromes per round in the PCM.
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -248,12 +248,10 @@ generate_random_pcm_sparse(std::size_t n_rounds, std::size_t n_errs_per_round,
 /// output has the same layout as the input and is idempotent under further
 /// `canonicalize_pcm` calls.
 ///
-/// Use this before passing a PCM built from a DEM decomposition (or any other
-/// source that may legitimately emit duplicate indices within a column/row)
-/// to consumers like PyMatching that dispatch on per-column nnz count and
-/// assume at-most-one entry per row per column. `sparse_binary_matrix`
-/// itself accepts duplicates at construction (see its header) so a separate
-/// call is required to canonicalize.
+/// Use this when a PCM source, such as a DEM decomposition, may legitimately
+/// emit duplicate indices within a column/row and the caller wants to apply
+/// GF(2) duplicate-collapse semantics before passing the matrix to consumers
+/// that require at-most-one entry per row per column.
 sparse_binary_matrix canonicalize_pcm(const sparse_binary_matrix &pcm);
 
 /// @brief Randomly permute the columns of a PCM.
--- a/libs/qec/include/cudaq/qec/sparse_binary_matrix.h
+++ b/libs/qec/include/cudaq/qec/sparse_binary_matrix.h
@@ -1,0 +1,124 @@
+/****************************************************************-*- C++ -*-****
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#pragma once
+
+#include "cuda-qx/core/tensor.h"
+#include <cstdint>
+#include <vector>
+
+namespace cudaq::qec {
+
+/// @brief Storage layout for the sparse PCM: column-major (CSC) or row-major
+/// (CSR). All non-zero entries are assumed to be 1; values are not stored.
+enum class sparse_binary_matrix_layout { csc, csr };
+
+/// @brief Sparse parity-check matrix in either CSC or CSR form.
+/// Index types are uint32_t; non-zero values are always 1 and are not stored.
+class sparse_binary_matrix {
+public:
+  using index_type = std::uint32_t;
+
+  /// @brief Construct a sparse PCM in CSC form.
+  /// @param num_rows Number of rows.
+  /// @param num_cols Number of columns.
+  /// @param col_ptrs Column pointer array (length num_cols + 1); column j has
+  /// indices in \p row_indices[col_ptrs[j] .. col_ptrs[j+1]-1].
+  /// @param row_indices Row indices of non-zeros (length nnz).
+  static sparse_binary_matrix from_csc(index_type num_rows, index_type num_cols,
+                              std::vector<index_type> col_ptrs,
+                              std::vector<index_type> row_indices);
+
+  /// @brief Construct a sparse PCM in CSR form.
+  /// @param num_rows Number of rows.
+  /// @param num_cols Number of columns.
+  /// @param row_ptrs Row pointer array (length num_rows + 1); row i has
+  /// indices in \p col_indices[row_ptrs[i] .. row_ptrs[i+1]-1].
+  /// @param col_indices Column indices of non-zeros (length nnz).
+  static sparse_binary_matrix from_csr(index_type num_rows, index_type num_cols,
+                              std::vector<index_type> row_ptrs,
+                              std::vector<index_type> col_indices);
+
+  /// @brief Construct from nested CSC: \p nested[j] is the list of row indices
+  /// for column j; \p nested.size() must equal \p num_cols.
+  static sparse_binary_matrix from_nested_csc(
+      index_type num_rows, index_type num_cols,
+      const std::vector<std::vector<index_type>> &nested);
+
+  /// @brief Construct from nested CSR: \p nested[i] is the list of column
+  /// indices for row i; \p nested.size() must equal \p num_rows.
+  static sparse_binary_matrix from_nested_csr(
+      index_type num_rows, index_type num_cols,
+      const std::vector<std::vector<index_type>> &nested);
+
+  /// @brief Construct a sparse PCM from a dense PCM tensor (rows x columns).
+  /// Any non-zero entry is treated as 1.
+  /// @param dense Dense parity-check matrix; must have rank 2.
+  /// @param layout Storage layout for the sparse representation (default CSC).
+  sparse_binary_matrix(const cudaqx::tensor<std::uint8_t> &dense,
+             sparse_binary_matrix_layout layout = sparse_binary_matrix_layout::csc);
+
+  // Default constructor
+  sparse_binary_matrix() = default;
+
+  // Copy constructor
+  sparse_binary_matrix(const sparse_binary_matrix&) = default;
+
+  // Move constructor
+  sparse_binary_matrix(sparse_binary_matrix&&) noexcept = default;
+
+  // Copy assignment operator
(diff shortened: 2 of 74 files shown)


libs/qec/include/cudaq/qec/sparse_binary_matrix.h
```diff
@@ -0,0 +1,121 @@
+/****************************************************************-*- C++ -*-****
```
sacpis, line 1: nit: Can we please make sure these headers are same in all files?

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
```diff
@@ -118,30 +120,35 @@ class pymatching : public decoder {
...
+    user_graph = pm::UserGraph(H.num_rows());
 
-    auto sparse = cudaq::qec::dense_to_sparse(H);
+    // PyMatching dispatches on per-column nnz ∈ {1, 2}; canonicalize so
+    // GF(2)-duplicates from DEM decompositions collapse first.
+    std::vector<std::vector<std::uint32_t>> H_e2d =
+        canonicalize_pcm(H).to_nested_csc();
     std::size_t col_idx = 0;
-    for (auto &col : sparse) {
+    for (std::size_t col = 0; col < block_size; col++) {
```
sacpis: Seems like col and col_idx are being incremented in lockstep. Do we need both then?

libs/qec/include/cudaq/qec/sparse_binary_matrix.h
```diff
@@ -0,0 +1,121 @@
...
+ ******************************************************************************/
+#pragma once
+
+#include "cuda-qx/core/tensor.h"
+#include <cstdint>
+#include <vector>
+
+namespace cudaq::qec {
+
+/// @brief Storage layout for the sparse PCM: column-major (CSC) or row-major
```
sacpis: nit: Should we also add `Compressed Sparse Column (CSC)` and `Compressed Sparse Row (CSR)`?

libs/qec/include/cudaq/qec/pcm_utils.h
```diff
@@ -111,6 +112,14 @@ get_sorted_pcm_column_indices(const cudaqx::tensor<uint8_t> &pcm,
 bool pcm_is_sorted(const cudaqx::tensor<uint8_t> &pcm,
                    std::uint32_t num_syndromes_per_round = 0);
 
+/// @brief Check if a PCM is sorted.
+/// @param sparse_pcm The sparse PCM to check (in the same format as
+/// get_sparse_pcm())
+/// @param num_syndromes_per_round The number of syndromes per round.
+/// @return True if the PCM is sorted, false otherwise.
+bool pcm_is_sorted(const std::vector<std::vector<std::uint32_t>> &sparse_pcm,
```
sacpis, line 120: nit: would have named it as `is_pcm_sorted`, just for readability.

libs/qec/include/cudaq/qec/pcm_utils.h
```diff
@@ -165,7 +174,50 @@ get_pcm_for_rounds(const cudaqx::tensor<uint8_t> &pcm,
...
+///
+/// @param pcm_is_canonical If true, the caller asserts \p pcm has
+/// sorted-unique per-group indices (i.e. is the output of \c canonicalize_pcm
+/// or was constructed canonically). In that case the per-call
+/// canonicalization step is skipped — useful for callers like
+/// \c sliding_window that canonicalize once at construction and then call
+/// this function in a per-window loop on the same matrix. Default \c false
+/// preserves the original "canonicalize on entry" behavior.
+///
+/// @warning When \p pcm_is_canonical is \c true the precondition is *not*
```
sacpis: Do we need these formatting (`\p`, '\c' etc) of words in the comments? I don't see it in the old / existing code.

libs/qec/include/cudaq/qec/pcm_utils.h
```diff
@@ -165,7 +174,50 @@ get_pcm_for_rounds(const cudaqx::tensor<uint8_t> &pcm,
...
+/// @brief Upper bound on \f$\texttt{rows} \times \texttt{cols}\f$ for the
+/// dense \c generate_random_pcm path. Above this, the call throws and the
+/// caller should switch to \c generate_random_pcm_sparse. Exposed as a named
+/// symbol so tests and users can reference it without grepping the source.
+inline constexpr std::size_t k_max_dense_pcm_elements =
+    static_cast<std::size_t>(400u) * 1024u * 1024u;
+
 /// @brief Generate a random PCM with the given parameters.
+///
+/// If \f$\texttt{rows} \times \texttt{cols}\f$ exceeds \c
```
sacpis: I think we can put the formula here.

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
```diff
@@ -6,15 +6,17 @@
...
 #include <algorithm>
 #include <cassert>
 #include <map>
 #include <vector>
 
+#include "cudaq/qec/decoder.h"
+#include "cudaq/qec/pcm_utils.h"
+
+#include "pymatching/sparse_blossom/driver/mwpm_decoding.h"
+#include "pymatching/sparse_blossom/driver/user_graph.h"
```
sacpis: nit: Was this re-arrangement of headers because of some formatting tool? I see different sequence in other files.

libs/qec/lib/decoder.cpp
```diff
@@ -479,7 +480,7 @@ void decoder::reset_decoder() {
 }
 
 std::unique_ptr<decoder> get_decoder(const std::string &name,
-                                     const cudaqx::tensor<uint8_t> &H,
+                                     const sparse_binary_matrix &H,
```
sacpis: nit: `cudaq::qec::sparse_binary_matrix`. Let's make sure we are using the namespace elsewhere as well.

libs/qec/lib/pcm_utils.cpp
```diff
@@ -1,15 +1,58 @@
...
+    const std::vector<std::vector<std::uint32_t>> &row_indices,
+    std::uint32_t num_syndromes_per_round, std::uint32_t start_round,
+    std::uint32_t end_round, bool straddle_start_round, bool straddle_end_round,
+    std::vector<std::uint32_t> &columns_in_range_out,
+    std::uint32_t &first_column, std::uint32_t &last_column) {
+  columns_in_range_out.clear();
+  for (std::size_t c = 0; c < row_indices.size(); c++) {
+    const auto &rows_for_this_column = row_indices[c];
+    if (rows_for_this_column.size() == 0)
+      continue;
```
sacpis: For simplicity, we can write it as (`rows_for_this_column` is a bit confusing for me)

```
const auto &col = row_indices[c];
if (col.empty())
    continue;
```

libs/qec/lib/pcm_utils.cpp
```diff
@@ -1,15 +1,58 @@
...
+    auto first_round = rows_for_this_column.front() / num_syndromes_per_round;
+    auto last_round = rows_for_this_column.back() / num_syndromes_per_round;
+    if (first_round >= start_round && last_round <= end_round)
+      columns_in_range_out.push_back(static_cast<std::uint32_t>(c));
+    else if (straddle_start_round && first_round <= start_round &&
+             last_round >= start_round)
+      columns_in_range_out.push_back(static_cast<std::uint32_t>(c));
+    else if (straddle_end_round && first_round <= end_round &&
+             last_round >= end_round)
+      columns_in_range_out.push_back(static_cast<std::uint32_t>(c));
```
sacpis: As we are repeating the same push_back for all 3 cases, we can do

```
if (last_round < start_round || first_round > end_round)
    continue;

const bool fully_inside = first_round >= start_round && last_round <= end_round;
if (fully_inside ||
          (straddle_start_round && first_round <= start_round &&
           last_round >= start_round) ||
          (straddle_end_round && first_round <= end_round &&
           last_round >= end_round))
    columns_in_range_out.push_back(static_cast<std::uint32_t>(c));
```

libs/qec/lib/pcm_utils.cpp
```diff
@@ -1,15 +1,58 @@
...
+  }
+
+  first_column = std::numeric_limits<std::uint32_t>::max();
+  last_column = std::numeric_limits<std::uint32_t>::min();
+  for (auto c : columns_in_range_out) {
+    first_column = std::min(first_column, c);
+    last_column = std::max(last_column, c);
+  }
```
sacpis, line 128: Can be written as

```
first_column = std::numeric_limits<std::uint32_t>::max();
last_column = 0;
if (!columns_in_range_out.empty()) {
    first_column = columns_in_range_out.front();
    last_column  = columns_in_range_out.back();
}
```

libs/qec/lib/pcm_utils.cpp
```diff
@@ -331,47 +389,81 @@ get_pcm_for_rounds(const cudaqx::tensor<uint8_t> &pcm,
...
+  if (first_row_to_keep >= pcm.num_rows()) {
+    throw std::invalid_argument(
+        "get_pcm_for_rounds: first_row_to_keep is greater than the number of "
+        "rows in PCM");
+  }
+  if (last_row_to_keep >= pcm.num_rows()) {
+    throw std::invalid_argument(
+        "get_pcm_for_rounds: last_row_to_keep is greater than the number of "
+        "rows in PCM");
+  }
```
sacpis: nit: This code could go into `validate_parameters()` function.

libs/qec/lib/pcm_utils.cpp
```diff
@@ -137,32 +191,36 @@ bool pcm_is_sorted(const cudaqx::tensor<uint8_t> &pcm,
...
   return row_indices;
 }
 
+/// @brief Return a sparse representation of the PCM.
+/// @return A vector of vectors that sparsely represents the PCM. The size of
+/// the outer vector is the number of columns in the PCM, and the i-th element
+/// contains an inner vector of the row indices of the non-zero elements in the
+/// i-th column of the PCM.
+std::vector<std::vector<std::uint32_t>>
+get_sparse_pcm(const cudaqx::tensor<uint8_t> &pcm) {
```
bmhowe23: Curious: why was this 1-line wrapper added (get_sparse_pcm wrapping dense_to_sparse)? It looks like the original get_sparse_pcm was doing the same thing without the wrapper.

bmhowe23 (commented):
Thanks, Vedika. I think these changes are looking pretty good. Could you please prepare any necessary changes to the nv-qldpc-decoder (closed source) before we merge this?

libs/qec/lib/pcm_utils.cpp
```diff
@@ -416,6 +531,134 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
+  if (n_rounds != 0 && (n_errs_per_round > kMaxDim / n_rounds ||
+                        n_syndromes_per_round > kMaxDim / n_rounds)) {
+    throw std::invalid_argument(
+        "generate_random_pcm_sparse: n_rounds * n_errs_per_round and "
+        "n_rounds * n_syndromes_per_round must each fit in "
+        "sparse_binary_matrix::index_type (uint32_t); got n_rounds=" +
+        std::to_string(n_rounds) +
+        ", n_errs_per_round=" + std::to_string(n_errs_per_round) +
+        ", n_syndromes_per_round=" + std::to_string(n_syndromes_per_round));
+  }
```
sacpis: nit: can be moved out to a validate function.

sacpis (commented):
In the middle of the review....

libs/qec/lib/pcm_utils.cpp
```diff
@@ -416,6 +531,134 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
+        nested[c_ix].push_back(
+            static_cast<sparse_binary_matrix::index_type>(global_row));
+      }
+    }
+  }
+
+  // Sort per column so downstream consumers reading .front()/.back() as
+  // min/max row get correct round assignments.
+  for (auto &col : nested)
+    std::sort(col.begin(), col.end());
```
sacpis, line 628: I think here we can use insertion sort to keep it sorted without a final sweep.

sacpis (commented):
Thanks @vedika-saravanan. Looks good overall but would like you to give it a thought to optimize it as it will help us with runtime and hot-paths.

libs/qec/include/cudaq/qec/sparse_binary_matrix.h
vedika-saravanan: sparse_binary_matrix.h shows 2026 only because the file is new in this PR, matches the cudaqx convention for new files (e.g. experiments.cpp, stabilizer_utils.cpp). Resolving this.

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
vedika-saravanan: Good catch. Pre-PR the loop was for (auto &col : sparse) where col was a vector reference and col_idx was the separate integer index. When it was refactored to integer indexing, col became the index and col_idx became redundant but was left behind. Removed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/include/cudaq/qec/sparse_binary_matrix.h
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/include/cudaq/qec/pcm_utils.h
vedika-saravanan: Agree it reads better as is_pcm_sorted, but pcm_is_sorted is already part of the public C++ and Python API (qec.pcm_is_sorted) and is documented in three Sphinx files. Renaming would be a user-facing API break across 10 files, including the docs site. Out of scope for this PR, happy to do it as a separate rename PR

libs/qec/include/cudaq/qec/pcm_utils.h
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/include/cudaq/qec/pcm_utils.h
vedika-saravanan: Good idea made it concrete. Fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/decoder.cpp
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: Good catch, that wrapper shouldn't have been left in. dense_to_sparse is the real implementation, the wrapper isn't in the public header. removed, and fixed the stale get_sparse_pcm() reference in the header comment. Fixed in [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: fixed it [058c652](https://github.com/NVIDIA/cudaqx/pull/550/commits/058c65211125d085dc0fee978383a165fb4234d4).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: Current collect-then-sort is O(w log w) per column vs O(w²) for insertion-sort. At typical QEC weights they're equivalent, but weight can go up to n_syndromes_per_round where std::sort scales noticeably better. Leaning on keeping the current form for asymptotic safety, happy to revisit if needed.

vedika-saravanan:
> Thanks, Vedika. I think these changes are looking pretty good. Could you please prepare any necessary changes to the nv-qldpc-decoder (closed source) before we merge this?

@bmhowe23, I have raised a merge request for necessary changes in nv-qldpc-decoder (219)

libs/qec/include/cudaq/qec/sparse_binary_matrix.h
sacpis: Oh I was talking about the `/***` line.
Please see this one
libs/qec/lib/decoders/plugins/example/single_error_lut_example.cpp

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -196,19 +199,25 @@ sliding_window::sliding_window(const cudaqx::tensor<uint8_t> &H,
...
+  // this->H is canonical CSC (ctor init list), so skip the per-call
+  // canonicalize in get_pcm_for_rounds.
   for (std::size_t w = 0; w < num_windows; ++w) {
     std::size_t start_round = w * step_size;
     std::size_t end_round = start_round + window_size - 1;
     auto [H_round, first_column, last_column] = cudaq::qec::get_pcm_for_rounds(
-        H, num_syndromes_per_round, start_round, end_round,
-        straddle_start_round, straddle_end_round);
+        this->H, num_syndromes_per_round, start_round, end_round,
+        straddle_start_round, straddle_end_round, /*pcm_is_canonical=*/true);
```
sacpis, line 220: Instead of calling `get_pcm_for_rounds` through the num_windows, where internally ever call does `pcm.to_nested_csc()` which does a new vector<vector<uint32_t>> sized num_cols and a copy as well. I think here we can add an overload that reads off the canonical CSC arrays (ptr_ and/or indices_).

libs/qec/lib/pcm_utils.cpp
```diff
@@ -1,15 +1,121 @@
...
+    throw std::invalid_argument(
+        "get_pcm_for_rounds: last_row_to_keep is greater than the number of "
+        "rows in PCM");
+  }
+}
+
+/// Select PCM column indices with any nonzero in rows belonging to rounds
+/// [start_round, end_round], honoring straddle flags (same logic as dense
+/// `get_pcm_for_rounds`).
+void select_pcm_columns_for_round_range(
```
sacpis, line 97: A quick thought. Instead of scanning every column to find the overlaps, how about we compute `(first_round, last_round)` once and then do a binary search over a sorted first_round array. This will reduce a ton of looping here.

vedika-saravanan: Valid optimization for sliding_window's multi-call pattern (cached (first_round, last_round) array with binary search). For a single call it's net-neutral. Same scope concern, deferring to the perf follow-up PR with the CSC-direct change so both can be profiled together.

libs/qec/lib/decoders/sliding_window.cpp
vedika-saravanan: It is a real win, I agree with you, but the helper + dispatch surface is non-trivial. Given how large this PR already is, leaning on deferring to a perf follow-up PR after these lands so we can profile sliding_window's hot path first. Tracking alongside the sparse-aware PCM utility migration.

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
```diff
@@ -118,32 +118,32 @@ class pymatching : public decoder {
...
-    user_graph = pm::UserGraph(H.shape()[0]);
+    user_graph = pm::UserGraph(H.num_rows());
 
-    auto sparse = cudaq::qec::dense_to_sparse(H);
-    std::size_t col_idx = 0;
-    for (auto &col : sparse) {
+    // PyMatching dispatches on per-column nnz ∈ {1, 2}; canonicalize so
+    // GF(2)-duplicates from DEM decompositions collapse first.
+    std::vector<std::vector<std::uint32_t>> H_e2d =
+        canonicalize_pcm(H).to_nested_csc();
```
bmhowe23: I don't think this is safe. That is - if a user provides `H` with a specific column ordering, the decoder must return predicted errors matching that column ordering. If we canonicalize the PCM, we will potentially both reorder and drop duplicate columns. More concretely, if canonicalization reorders the columns of H from [0,1,2,3...] to [0,1,3,2,...], then the user will interpret a vector returned by `decode` that has the 3rd bit set to mean that error 2 happened when in fact this canonicalization will mean that error 3 happened.

libs/qec/lib/decoders/sliding_window.cpp
bmhowe23: I agree that this can wait until a follow-up PR.

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -28,16 +33,49 @@
...
+sparse_binary_matrix_from_py_dict(const nb::dict &d) {
+  if (!d.contains("layout") || !d.contains("num_rows") ||
+      !d.contains("num_cols") || !d.contains("nested"))
+    throw std::runtime_error(
+        "Sparse H dict must have keys: layout, num_rows, num_cols, "
+        "nested. Use layout \"nested_csc\": nested has one list per COLUMN "
+        "with ROW indices where that column has a one; \"nested_csr\": one "
+        "list per ROW with COLUMN indices for that row.");
+  std::string layout = nb::cast<std::string>(d["layout"]);
+  auto num_rows = static_cast<sparse_binary_matrix::index_type>(
```
melody-ren: This is a risky cast. Casting to uint32_t could silently narrow the input which can be 64 bit. The check `from_nested_csc` later won't catch this because num_rows is already narrowed by then. We should check before doing the cast with something like this:

`auto raw = nb::cast<std::size_t>(d["num_rows"]); if (raw > std::numeric_limits<sparse_binary_matrix::index_type>::max())
  throw std::runtime_error("num_rows exceeds uint32_t range");
auto num_rows = static_cast<sparse_binary_matrix::index_type>(raw);`

libs/qec/lib/pcm_utils.cpp
```diff
@@ -416,6 +545,116 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
+      bool all_errors_in_this_round = dis(rng) ? true : false;
+      if (r == n_rounds - 1)
+        all_errors_in_this_round = true;
+      std::size_t row_max = all_errors_in_this_round
+                                ? n_syndromes_per_round
+                                : 2 * n_syndromes_per_round;
+      // See generate_random_pcm re: row_max==0 row_dis UB.
+      if (row_max == 0)
+        continue;
+      std::uniform_int_distribution<> row_dis(0, row_max - 1);
```
melody-ren:
```suggestion
      std::uniform_int_distribution<sparse_binary_matrix::index_type> row_dis(0, row_max - 1);
```

Leaving the `<>` empty means it's `int`, but `row_max` is size_t. If row_max -1 > INT_MAX, it'll silently truncate.

libs/qec/lib/pcm_utils.cpp
```diff
@@ -384,8 +490,28 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
+    throw std::invalid_argument(
+        "generate_random_pcm: weight must be in [0, n_syndromes_per_round]; "
+        "got weight=" +
+        std::to_string(weight) +
+        ", n_syndromes_per_round=" + std::to_string(n_syndromes_per_round));
+  }
   std::size_t n_cols = n_rounds * n_errs_per_round;
   std::size_t n_rows = n_rounds * n_syndromes_per_round;
+  const std::size_t n_elems = n_rows * n_cols;
+  if (n_elems > k_max_dense_pcm_elements) {
```
melody-ren: `n_col`, `n_rows` and `n_elems` might overflow and wrap before this check. We need to check the multiplication before using the result

libs/qec/lib/sparse_binary_matrix.cpp
```diff
@@ -0,0 +1,391 @@
...
+  num_rows_ = static_cast<index_type>(nrows);
+  num_cols_ = static_cast<index_type>(ncols);
+  layout_ = layout;
+
+  // size_t loops + `static_cast<size_t>(dim) + 1` ptr sizing avoid UINT32_MAX
+  // wraparound; see from_csc.
+  if (layout_ == sparse_binary_matrix_layout::csc) {
+    std::vector<index_type> col_nnz(num_cols_, 0);
+    for (std::size_t r = 0; r < num_rows_; ++r) {
+      const auto *row_ptr = &dense.at({r, std::size_t{0}});
```
melody-ren, line 265: The inner loop could be empty (as in, zero columns) so we might be dereferencing an element that doesn't exist. A safer pattern would be to check if num_cols_ is zero before getting the row_ptr. I think there are several other similar cases like this one elsewhere in this file.

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -321,11 +366,37 @@ void bindDecoder(nb::module_ &mod) {
...
+        }
+
+        cudaq::qec::sparse_binary_matrix H_sparse;
+
+        auto make_sparse_from_dense =
+            [](const nb::ndarray<nb::numpy, uint8_t> &arr) {
+              if (arr.ndim() != 2 || arr.itemsize() != sizeof(uint8_t))
+                throw std::runtime_error(
+                    "Parity check matrix must be 2-dimensional uint8.");
+              if (arr.stride(0) == (int64_t)sizeof(uint8_t))
```
melody-ren: This stride check will reject valid matrix size of (N,1) row-major arrays and doesn't do stride(1) check. I suggest to use the existing helpers `cudaqx::toTensor` or `cudaqx::pcmToTensor`.

libs/qec/lib/pcm_utils.cpp
```diff
@@ -416,6 +545,116 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
+                           std::size_t n_syndromes_per_round, int weight,
+                           std::mt19937_64 &&rng) {
+  validate_random_pcm_sparse_params(n_rounds, n_errs_per_round,
+                                    n_syndromes_per_round, weight);
+  std::size_t n_cols = n_rounds * n_errs_per_round;
+  std::size_t n_rows = n_rounds * n_syndromes_per_round;
+
+  std::uniform_int_distribution<> dis(0, 1);
+
+  std::vector<std::vector<sparse_binary_matrix::index_type>> nested(n_cols);
```
melody-ren, line 588: We should add a bound check on total `nnz` inside `validate_random_pcm_sparse_params` before allocating `nested`. The current checks ensure `n_rows` and `n_cols` fit in `index_type`, but `nnz = n_cols * weight` can still exceed `index_type::max()`. Since `from_nested_csc` only rejects that after `nested` has already been allocated/populated, oversized input can OOM before reaching the validator.

libs/qec/lib/sparse_binary_matrix.cpp
```diff
@@ -0,0 +1,391 @@
...
+  std::vector<std::vector<index_type>> out(num_cols_);
+  if (layout_ == sparse_binary_matrix_layout::csc) {
+    for (std::size_t j = 0; j < num_cols_; ++j) {
+      out[j].assign(indices_.begin() + ptr_[j], indices_.begin() + ptr_[j + 1]);
+    }
+  } else {
+    for (std::size_t i = 0; i < num_rows_; ++i) {
+      for (index_type p = ptr_[i]; p < ptr_[i + 1]; ++p) {
+        index_type j = indices_[p];
+        out[j].push_back(static_cast<index_type>(i));
```
melody-ren, line 429: Here each inner vector is dynamically allocated. I wonder if we can do some cheap optimization here by doing  a one-pass on each inner vector and count the number of nonzeros, then reserve space for the inner vector.  Similar idea goes to the `to_nested_csr` function

libs/qec/lib/pcm_utils.cpp
```diff
@@ -118,6 +224,17 @@ std::vector<std::uint32_t> get_sorted_pcm_column_indices(
   return column_order;
 }
 
+bool pcm_is_sorted(const std::vector<std::vector<std::uint32_t>> &sparse_pcm,
```
melody-ren, line 240: [For follow-up] `pcm_is_sorted` is a predicate, but it currently answers the question by building the full sorted column permutation via `get_sorted_pcm_column_indices` and checking whether it is identity. That works, but it couples a yes/no check to a heavier transformation helper and does `O(n log n)` work plus an allocation. Another thing is, today `get_sorted_pcm_column_indices` does not mutate `sparse_pcm`, but `pcm_is_sorted` calls it and may have unintended side-effects should `get_sorted_pcm_column_indices` behaviour changes in the future. Ideally, we'd want to decouple them. 

Could we extract the column-order comparator used by `get_sorted_pcm_column_indices` into a shared helper, then use that comparator in both places? `get_sorted_pcm_column_indices` can still sort the permutation, while `pcm_is_sorted` can do an `O(n)` adjacent inversion check with the exact same ordering definition.

libs/qec/lib/pcm_utils.cpp
```diff
@@ -416,6 +545,116 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
   return pcm;
 }
 
+sparse_binary_matrix
+generate_random_pcm_sparse(std::size_t n_rounds, std::size_t n_errs_per_round,
```
melody-ren, line 578: [For follow-up] It seems that this function and `generate_random_pcm` shares the same logic to generate random columns. I wonder if we can extract the common logic into a helper? (Good for maintainability, prevents future drift and subtle differences).

libs/qec/lib/pcm_utils.cpp
```diff
@@ -416,6 +545,116 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
+      }
+    }
+  }
+
+  // Sort per column so downstream consumers reading .front()/.back() as
+  // min/max row get correct round assignments.
+  for (auto &col : nested)
+    std::sort(col.begin(), col.end());
+
+  return sparse_binary_matrix::from_nested_csc(
```
melody-ren, line 630: [For follow-up]: the construction flow of `generate_random_pcm_sparse` seems to be this: allocate for `nested` > sorts each inner vector > `from_nested_csc` allocates/copies again into flat CSC storage. Since this generator knows each column' number of entries, I think we should be able to build `col_ptrs` and the flat `row_indices` array directly in a follow-up and avoid the nested intermediate allocations.

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
vedika-saravanan: Good catch, I checked this carefully.
canonicalize_pcm does not reorder columns: it iterates columns g=0..num_cols-1 and canonicalizes each column in place (sort + GF(2) duplicate collapse). So the mapping result[c] -> input column c is preserved; this is row-index normalization within a column, not a column permutation.

We call it to support non-canonical sparse inputs (e.g. DEM-style duplicates). The new test test_pymatching_accepts_non_canonical_sparse_input covers [0,0,0] -> [0]; without canonicalization the wrapper would reject that column for having 3 ones.

Edge cases:

- [1,1] canonicalizes to an empty column, which the wrapper rejects (has 0 ones) instead of treating it as a 2-entry column.
- `>2` distinct indices still reject (has 3 ones) either way.
- Distinct encodings that are GF(2)-equivalent can canonicalize to the same edge key, and edge2col_idx keeps the last column. That behavior is GF(2)-consistent but can change which column index is reported on decode.

libs/qec/python/bindings/py_decoder.cpp
vedika-saravanan: Good catch, fixed https://github.com/NVIDIA/cudaqx/pull/550/changes/174519cdc5543645035741938aa70446ca67dd7c. Added a checked_narrow_to_index_type(size_t, const char*) helper that throws with the field name and offending value if the input exceeds uint32_t range, and applied it to num_rows, num_cols, and the inner row indices in nested[i][j] (same silent-narrowing risk for any of the three).

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: fixed both the sparse and dense paths in https://github.com/NVIDIA/cudaqx/commit/174519cdc5543645035741938aa70446ca67dd7c. Sparse uses sparse_binary_matrix::index_type (uint32_t) since that's the type bounding the matrix; dense uses std::size_t since the tensor is size_t-indexed. The dense path is safe today because the 400M element cap keeps row_max well below INT_MAX, but the silent-narrowing risk was latent, so fixing both for consistency.

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: applied an overflow-safe cap check in (a > limit / b) form (same pattern as validate_random_pcm_sparse_params) https://github.com/NVIDIA/cudaqx/commit/174519cdc5543645035741938aa70446ca67dd7c

libs/qec/lib/sparse_binary_matrix.cpp
vedika-saravanan: fixed both. Added an early-return at the top of the dense ctor body when either num_rows_ == 0 or num_cols_ == 0: it sets ptr_ to num_groups + 1 zeros (a valid canonical empty matrix) and leaves indices_ empty. That single guard covers all four &dense.at({r, 0}) call sites (CSC and CSR paths, each in the count and build passes). Same pattern applied in to_dense to skip the memset when num_cols_ == 0 — the write loops below were already no-ops in that case. https://github.com/NVIDIA/cudaqx/commit/174519cdc5543645035741938aa70446ca67dd7c

libs/qec/python/bindings/py_decoder.cpp
vedika-saravanan: switched to cudaqx::pcmToTensor, which does the proper per-dimension stride check and skips degenerate dimensions via if (H.shape(i) > 1 && ...). That fixes the (N, 1) false-rejection and centralizes the validation. Existing column-major rejection still throws (a Fortran-order array fails the innermost-stride check). https://github.com/NVIDIA/cudaqx/commit/174519cdc5543645035741938aa70446ca67dd7c

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: added an nnz bound check in validate_random_pcm_sparse_params in overflow-safe (weight > kMaxDim / n_cols) form, after the existing dimension check. Rejects before nested is allocated/populated, so oversized input now throws cleanly instead of OOMing en route to from_nested_csc. https://github.com/NVIDIA/cudaqx/commit/174519cdc5543645035741938aa70446ca67dd7c

libs/qec/lib/sparse_binary_matrix.cpp
vedika-saravanan: Real win, reserve-then-push avoids the doubling-realloc overhead. Same family as the CSC-direct and binary-search optimizations from earlier comments, so leaning on bundling all three into a perf follow-up PR after this land. That lets us profile sliding_window and the conversion paths together and confirm the optimizations matter at user workloads. Tracking alongside the sparse-aware PCM utility migration.

melody-ren: Sounds good to me.

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
bmhowe23: > I checked this carefully. canonicalize_pcm does not reorder columns

I disagree. https://github.com/NVIDIA/cudaqx/blob/da7f89a6a869285c5111156dd189eefdac9b4072/libs/qec/lib/detector_error_model.cpp#L164 shows where it is reordering the columns in the detector_error_matrix and https://github.com/NVIDIA/cudaqx/blob/da7f89a6a869285c5111156dd189eefdac9b4072/libs/qec/lib/detector_error_model.cpp#L168 shows where it is reordering the columns in the observables_flips_matrix.

vedika-saravanan: you're right, sorry for the oversight. dropped canonicalize_pcm(H), so caller column order is preserved end-to-end

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
```diff
@@ -120,29 +119,52 @@ class pymatching : public decoder {
...
     for (std::size_t col = 0; col < block_size; col++) {
       double weight = 1.0;
       if (col < error_rate_vec.size()) {
         weight = -std::log(error_rate_vec[col] / (1.0 - error_rate_vec[col]));
       }
-      if (H_e2d[col].size() == 2) {
-        edge2col_idx[make_canonical_edge(H_e2d[col][0], H_e2d[col][1])] = col;
-        user_graph.add_or_merge_edge(H_e2d[col][0], H_e2d[col][1],
+
+      // Collapse GF(2)-duplicate row indices within this column only (handles
```
bmhowe23: I'd recommend we follow cuSparse's lead here and say that duplicate entries are not allowed. This would simplify this code, removing the sorts and traversals.

https://docs.nvidia.com/cuda/cusparse/index.htm
> The cuSPARSE routines assume that the indices are provided in increasing order and that each index appears only once. In the opposite case, the correctness of the computation is not always ensured.

If you *really* want to retain any sort of validation, I'd suggest putting that as a method in `cudaq::qec::sparse_binary_matrix` and just throw an exception if it's violated. (That should be a simple 1-pass traversal of each non-zero element.)

vedika-saravanan: thank you for the feedback. I agree, I have updated PyMatching to reject non-canonical sparse inputs instead of canonicalizing internally. It now validates that sparse indices are strictly increasing, so duplicate entries are not allowed.

libs/qec/include/cudaq/qec/pcm_utils.h
```diff
@@ -178,6 +233,27 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
+/// (CSR) has its indices sorted ascending, and duplicate indices are XOR-
+/// merged — an index that appears \p k times is kept iff \p k is odd. The
+/// output has the same layout as the input and is idempotent under further
+/// `canonicalize_pcm` calls.
+///
+/// Use this when a PCM source, such as a DEM decomposition, may legitimately
+/// emit duplicate indices within a column/row and the caller wants to apply
+/// GF(2) duplicate-collapse semantics before passing the matrix to consumers
+/// that require at-most-one entry per row per column.
+sparse_binary_matrix canonicalize_pcm(const sparse_binary_matrix &pcm);
```
bmhowe23, line 255: [ok for follow up, but before the next release] I'd recommend that we move this into a canonicalize function inside sparse_binary_matrix.

bmhowe23 (approved):
Thanks, @vedika-saravanan.

libs/qec/lib/decoders/sliding_window.cpp
vedika-saravanan: Created a GHI for followup: https://github.com/NVIDIA/cudaqx/issues/582. Makrking the comment resolved.

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: Created a GHI for followup: https://github.com/NVIDIA/cudaqx/issues/582. Makrking the comment resolved.

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: Created a GHI for followup: https://github.com/NVIDIA/cudaqx/issues/582. Makrking the comment resolved.

libs/qec/include/cudaq/qec/pcm_utils.h
vedika-saravanan: Created a GHI for followup: https://github.com/NVIDIA/cudaqx/issues/582. Makrking the comment resolved.

libs/qec/lib/pcm_utils.cpp
vedika-saravanan: Created a GHI for followup: https://github.com/NVIDIA/cudaqx/issues/582. Makrking the comment resolved.

PR #553, opened 2026-05-19
Use canonical soft-to-hard decoder conversion

### LINKED ISSUE #552: Decoders should use the canonical soft-to-hard conversion rule
The library already defines a canonical soft→hard helper at [libs/qec/include/cudaq/qec/decoder.h:296](vscode-webview://1222tmce6kuos244oulmcscgjlvuimvco1hkgafnmvdr6ved4etn/libs/qec/include/cudaq/qec/decoder.h#L296) with a documented contract — "Values >= thresh are assigned 1/true, all others 0/false". However, the following decoders inlines its own threshold loop with a manual comparison `if (syndrome[i] >= 0.5)`:

```
trt_decoder.cpp
lut.cpp
single_error_lut_example.cpp
pymatching.cpp:173 (this one actually does `if(syndrome[i]) > 0.5`!)
```
Action needed: these decoders should use the canonical helpers from the header `decoder.h`

melody-ren:
**Summary**
Use the canonical `cudaq ::qec ::convert_vec_soft_to_hard `helper for decoder syndrome binarization instead of inline threshold checks.

**Changes**

- Update LUT, example LUT, PyMatching, and TRT decoder input paths to use the shared soft-to-hard conversion helper.
- Align PyMatching’s boundary behavior with the documented helper contract: values >= 0.5 map to 1.
- Add focused boundary tests for 0.5 conversion behavior in C++ and Python.
- Keep TRT float inputs unchanged so soft-valued model inputs are still passed through as before.

**Note**
This PR intentionally leaves TRT decoder output conversion helpers unchanged. Those paths still use local threshold logic and should be revisited in a later PR to decide whether output binarization should also be centralized around the canonical conversion helper.

Resolves #552

### DIFF
--- a/libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
+++ b/libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
@@ -167,12 +167,10 @@ public:
     auto t1 = std::chrono::high_resolution_clock::now();
 #endif
 
-    std::vector<uint8_t> hard_syndrome;
-    cudaq::qec::convert_vec_soft_to_hard(syndrome, hard_syndrome);
     std::vector<uint64_t> detection_events;
-    detection_events.reserve(hard_syndrome.size());
-    for (size_t i = 0; i < hard_syndrome.size(); i++)
-      if (hard_syndrome[i])
+    detection_events.reserve(syndrome.size());
+    for (size_t i = 0; i < syndrome.size(); i++)
+      if (cudaq::qec::convert_soft_to_hard(syndrome[i]))
         detection_events.push_back(i);
 #if PERFORM_TIMING
     auto t2 = std::chrono::high_resolution_clock::now();
--- a/libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
+++ b/libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
@@ -167,10 +167,12 @@ public:
     auto t1 = std::chrono::high_resolution_clock::now();
 #endif
 
+    std::vector<uint8_t> hard_syndrome;
+    cudaq::qec::convert_vec_soft_to_hard(syndrome, hard_syndrome);
     std::vector<uint64_t> detection_events;
-    detection_events.reserve(syndrome.size());
-    for (size_t i = 0; i < syndrome.size(); i++)
-      if (syndrome[i] > 0.5)
+    detection_events.reserve(hard_syndrome.size());
+    for (size_t i = 0; i < hard_syndrome.size(); i++)
+      if (hard_syndrome[i])
         detection_events.push_back(i);
 #if PERFORM_TIMING
     auto t2 = std::chrono::high_resolution_clock::now();
--- a/libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
+++ b/libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
@@ -926,15 +926,19 @@ std::vector<decoder_result> trt_decoder::decode_batch_impl(
       // Prepare input batch. For float input we preserve soft (raw) values;
       // for uint8 we binarize to 0/1.
       std::vector<IoType> input_host(impl_->input_size);
+      std::vector<uint8_t> hard_syndrome;
       for (size_t batch_idx = 0; batch_idx < actual_batch; ++batch_idx) {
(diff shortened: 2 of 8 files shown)


libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
```diff
@@ -167,10 +167,12 @@ class pymatching : public decoder {
     auto t1 = std::chrono::high_resolution_clock::now();
 #endif
 
+    std::vector<uint8_t> hard_syndrome;
```
bmhowe23: My concern about changes like is that we now require an extra memory allocation (and deallocation) plus reading/writing to a temporary array when it isn't really needed. This is in the "hot path", so I'd rather not add things that aren't necessary. If you want to use common code from decoder.h for clarity, I'd recommend making a templated inline `soft_to_bool(t_soft)`, or something like that.

libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
```diff
@@ -926,15 +926,19 @@ std::vector<decoder_result> trt_decoder::decode_batch_impl(
       // Prepare input batch. For float input we preserve soft (raw) values;
       // for uint8 we binarize to 0/1.
       std::vector<IoType> input_host(impl_->input_size);
+      std::vector<uint8_t> hard_syndrome;
```
bmhowe23: For this one, it also creates extra dynamically allocated temporary storage that isn't really needed. Perhaps we need an override of `convert_vec_soft_to_hard(const std::vector<t_soft> &in, t_hard *ptr_out)?

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
melody-ren: I agree that it's a bit of a waste to have a vector allocated here just to use the canonical form. I'll make a scalar version of the canonical function so that the common form can be used and we don't have this allocation here. 

In general I prefer to use the common code so that we won't have subtle differences where some decoders do `>=0.5` and others do `>0.5`

libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
melody-ren: Agree and will update this to remove the need for extra allocation

melody-ren: Addressed in eb116fb66d9f2e39962341904499bc7756837240

libs/qec/lib/decoders/plugins/pymatching/pymatching.cpp
melody-ren: Addressed in eb116fb66d9f2e39962341904499bc7756837240

wsttiger (approved):
Thanks for this — nice cleanup. Centralizing the threshold rule behind a single canonical helper is the right call, and the scalar `convert_soft_to_hard()` is a good response to the earlier hot-path/allocation feedback. Routing the vector/tensor/2D variants through it removes the `>=` vs `>` drift across decoders, and the `if constexpr` hoist in the TRT path is a nice incidental tidy-up.

One item to address (already reflected by the red CI):

- **The new Python test is failing CI** (all four QEC build/test jobs). `test_single_error_lut_example_uses_canonical_threshold` does `assert result.result == [1.0, 0.0]`, but `result.result` is an `np.ndarray`, so the comparison yields an element-wise array and `assert` raises `ValueError: The truth value of an array with more than one element is ambiguous`. The decoder logic is correct — only the assertion is wrong. Suggest:
  ```python
  assert np.array_equal(result.result, [1.0, 0.0])
  ```
  Worth re-running the QEC Python suite after the fix, since the job bails on first failure and this `ndarray == list` pattern could recur.

Two non-blocking notes:

- The pymatching change flips the tie-break at exactly 0.5 (`> 0.5` → `>= 0.5`), so 0.5 now registers a detection event. It's intentional and matches the documented contract, but since it's a real behavioral change inside a refactor, it'd be good to call it out explicitly in the PR description.
- Minor naming asymmetry: scalar `convert_soft_to_hard` returns `bool` while `convert_vec_soft_to_hard` targets an integral `t_hard`. Fine to keep for symmetry — just flagging.

Otherwise the code itself looks good to me.

PR #558, opened 2026-05-21
Return Python decode result as NumPy arrays

melody-ren:
## Description

Updates Python decoder result access so decoded correction data is exposed as NumPy arrays instead of Python lists.

This is a breaking Python API change: `DecoderResult.result` now returns a 1-D NumPy array. This affects:

- `decoder.decode(...).result`
- `decoder.decode_async(...).get().result`
- tuple unpacking of `DecoderResult`
- `decoder.decode_batch(...)[i].result`
- iteration over `BatchDecoderResult`

Existing code that reads, indexes, iterates, or numerically consumes `result` should generally continue to work. Code that requires `result` to be an actual Python `list`, mutates it with list APIs, compares it with list equality, or serializes it directly should migrate to NumPy semantics or call `.tolist()`.

Also updates `decode_batch(...)` to return `BatchDecoderResult` instead of `list[DecoderResult]`.

`BatchDecoderResult` exposes vectorized batch fields:

- `result`: 2D NumPy array
- `converged`: 1D NumPy bool array
- `opt_results`: list-like per-shot optional results

Compatibility for existing batch consumers is preserved through indexing, slicing, `len(...)`, and iteration:

- `batch[i]` materializes a per-shot `DecoderResult`
- `batch[a:b]` returns another `BatchDecoderResult`
- `for r in batch` continues to yield per-shot `DecoderResult` objects

This avoids forcing user code onto the old per-shot Python list extraction path when the natural batch output is already array-shaped.

Python decoder plugin authors with custom `decode_batch(...)` overrides must now return `BatchDecoderResult`, not `list[DecoderResult]`.

User-facing Sphinx documentation will be updated in a separate docs PR, per project guidance that feature docs should not land before release.

This change is intended to reduce Python-side result extraction overhead for batch decoding by allowing users to read batch results directly as NumPy arrays instead of iterating through a list of `DecoderResult` objects.

Minibench run on commit `246114b0a33e9a58ff3a96e9abb77ac72e945fd1` using the C++ `single_error_lut` decoder.

The benchmark uses a CUDA-Q/Stim Steane memory syndrome workload, then splits the fixed shot set into chunks of `batch_size`. Each timed repeat loops over all chunks and measures wall-clock time from calling `decode_batch(...)` through reading the decoded result.

Two current-branch access patterns are compared:

- `numpy_view`: `decoded = decoder.decode_batch(chunk)`, then read `decoded.result` and `decoded.converged`
- `compat_iteration`: `decoded = decoder.decode_batch(chunk)`, then iterate `for r in decoded` and read `r.result` and `r.converged` (this is the backward compatible list access pattern)

Each row reports the median of 5 measured repeats. Speedup is `compat_iteration median / numpy_view median`.

| shots | batch size | decode_batch calls | result width | NumPy view median ms | compat iteration median ms | speedup |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1000 | 1 | 1000 | 697 | 3.892 | 7.234 | 1.86x |
| 1000 | 32 | 32 | 697 | 1.990 | 4.111 | 2.07x |
| 1000 | 256 | 4 | 697 | 2.399 | 3.936 | 1.64x |
| 3000 | 1 | 3000 | 697 | 11.748 | 21.462 | 1.83x |
| 3000 | 32 | 94 | 697 | 6.040 | 12.162 | 2.01x |
| 3000 | 256 | 12 | 697 | 5.923 | 11.818 | 1.99x |

In this minibench, direct NumPy batch access is about `1.6x-2.1x` faster than compatibility iteration for the measured `single_error_lut` cases.

### DIFF
--- a/libs/qec/python/bindings/py_decoder.cpp
+++ b/libs/qec/python/bindings/py_decoder.cpp
@@ -231,11 +231,14 @@ nb::object decoderResultsToNumpy(const std::vector<decoder_result> &results) {
   // not rely on result.shape[1] when the batch is empty.
   const auto result_size = num_results == 0 ? 0 : results.front().result.size();
 
-  for (const auto &result : results) {
-    if (result.result.size() != result_size) {
-      throw std::runtime_error(
+  for (std::size_t i = 0; i < results.size(); ++i) {
+    const auto actual_size = results[i].result.size();
+    if (actual_size != result_size) {
+      throw std::runtime_error(fmt::format(
           "Cannot return decode_batch results as a NumPy array because result "
-          "vectors have inconsistent sizes.");
+          "vectors have inconsistent sizes: expected row width {}, but row {} "
+          "has width {}.",
+          result_size, i, actual_size));
     }
   }
 
--- a/libs/qec/python/bindings/py_decoder.cpp
+++ b/libs/qec/python/bindings/py_decoder.cpp
@@ -16,6 +16,7 @@
 #include <filesystem>
 #include <limits>
 #include <link.h>
+#include <memory>
 #include <nanobind/nanobind.h>
 #include <nanobind/ndarray.h>
 #include <nanobind/stl/complex.h>
@@ -32,8 +33,10 @@
 #include "cuda-qx/core/kwargs_utils.h"
 #include "cuda-qx/core/library_utils.h"
 #include "type_casters.h"
+#include <algorithm>
 
 namespace nb = nanobind;
+using namespace nanobind::literals;
 using namespace cudaqx;
 
 namespace cudaq::qec {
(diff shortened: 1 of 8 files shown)


libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -32,8 +33,10 @@
 #include "cuda-qx/core/kwargs_utils.h"
 #include "cuda-qx/core/library_utils.h"
 #include "type_casters.h"
+#include <algorithm>
 
 namespace nb = nanobind;
+using namespace nanobind::literals;
```
bmhowe23: Nit: `using` more than 1 namespace in a single file brings in some risk of collisions. If you want a better abbreviation, perhaps this would be better (along with whatever downstream changes this induces).
```suggestion
namespace nbl = nanobind::literals;
```

bmhowe23 (commented):
Thanks, Melody. One minor nit below. Also, I submitted https://github.com/NVIDIA/cudaqx/actions/runs/26315316591 just to ensure good test coverage on both CPU and GPU machines for the tensor network decoder.

libs/qec/python/bindings/py_decoder.cpp
melody-ren: Dropped the namespace entirely since there are only two call sites. Resolved in e8bb8ca7203e1066fd19a6144a2e38523ae58eab

melody-ren:
Wheel job here: https://github.com/NVIDIA/cudaqx/actions/runs/26325036958

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -121,12 +329,28 @@ void bindDecoder(nb::module_ &mod) {
         True if the decoder successfully found a valid correction chain,
         False if the decoder failed to converge or exceeded iteration limits.
     )pbdoc")
-      .def_rw("result", &decoder_result::result, R"pbdoc(
+      .def_prop_rw(
+          "result",
+          [](const decoder_result &self) {
+            return decoderResultToNumpy(self.result);
```
wsttiger, line 338: DecoderResult.result changes from list to np.ndarray for single-shot decode() too.
The PR is framed as a decode_batch(...) return-type change, but here this changes DecoderResult.result itself to return a fresh NumPy array. That affects normal decoder.decode(...), decode_async().get(), tuple unpacking, and batch[i]. Existing user code expecting a real list from single-shot decode, list equality, or isinstance(res.result, list) will break. If this is intentional, the PR description and release notes should call out the broader API break; otherwise keep DecoderResult.result as the old list surface and only expose arrays on BatchDecoderResult.

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -94,6 +96,197 @@ std::unordered_map<std::string,
...
+  const auto num_results = results.size();
+  // Empty batch yields shape (0, 0). The per-shot width is unknown without
+  // running a decode (and depends on decoder mode — e.g. decode_to_observables
+  // produces num_observables-wide rows, not block_size-wide). Callers should
+  // not rely on result.shape[1] when the batch is empty.
+  const auto result_size = num_results == 0 ? 0 : results.front().result.size();
+
+  for (const auto &result : results) {
+    if (result.result.size() != result_size) {
+      throw std::runtime_error(
```
wsttiger: Optional DX nit: when rejecting ragged decoder results, could the error include the expected row width, offending index, and actual width? The current message is correct, but if a custom C++ decoder accidentally returns mixed-width rows, this would make the source much easier to diagnose.

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
```diff
@@ -416,15 +417,17 @@ def decode(
...
         """Decode a batch of detection events.
 
         Args:
             syndrome_batch (np.ndarray): A numpy array of shape (batch_size, syndrome_length) where each row is a detection event.
 
         Returns:
-            list[qec.DecoderResult]: list of results for each detection event in the batch.
-                The probabilities that the logical observable flipped for each syndrome.
+            qec.BatchDecoderResult: batched results for each detection event in
+                the batch. The `result` field has shape (batch_size, 1) and
```
wsttiger, line 428: Tiny doc nit: this docstring says `result` has shape `(batch_size, 1)`, which is true for this decoder. It might be worth saying that this decoder produces single-column batch results, while `BatchDecoderResult.result` in general has one row per shot and decoder-specific width.

libs/qec/python/bindings/py_decoder.cpp
melody-ren: Thanks, this is a good catch. The broader `DecoderResult.result` change is intentional: we want decoded correction data to use NumPy consistently, including single-shot `decode(...)`, `decode_async(...).get()`, tuple unpacking, and per-shot results materialized from batch indexing/iteration.

I updated the PR description to call this out explicitly as a breaking Python API change and to describe the migration path for list-specific code (`.tolist()`, NumPy comparisons, NumPy serialization handling, etc.).

libs/qec/python/bindings/py_decoder.cpp
melody-ren: Agreed. Patched in 1426768b9884248e8ce179999cc035766ef55744

libs/qec/python/cudaq_qec/plugins/decoders/tensor_network_decoder.py
melody-ren: I’m inclined to leave this one as-is because the docstring is for the tensor-network decoder specifically, and for this decoder the batch result is single-column. The general `BatchDecoderResult.result` contract is documented on `BatchDecoderResult` itself as one row per shot with decoder-specific width.

wsttiger (approved):

PR #563, opened 2026-05-23
Bump CUDAQ commit and set library mode in cudaqx_set_target

Renaud-K:
1. Bumping the CUDAQ commit because it is the one I have been validating my cudaq changes against. I know for it to be good up to that point.
2. Ahead of the runtime refactor, I need to make `CUDAQ_LIBRARY_MODE` explicit in `cudaqx_set_target`. I am hoping that  `cudaqx_set_target` is not a feature.

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "67404cea365c0f78a9319f89ef634afca15011c8"
+    "ref": "b04dee5177fcdc0b2370a1f983cf6d4d75f9893b"
   }
 }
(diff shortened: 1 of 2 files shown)


bmhowe23:
> 1. Bumping the CUDAQ commit because it is the one I have been validating my cudaq changes against. I know for it to be good up to that point.
> 2. Ahead of the runtime refactor, I need to make `CUDAQ_LIBRARY_MODE` explicit in `cudaqx_set_target`. I am hoping that  `cudaqx_set_target` is not a feature.

Dependence on library mode was intentionally removed in https://github.com/NVIDIA/cudaqx/pull/523 in preparation for other CUDA-Q changes, so we should probably discuss a path forward here.

Renaud-K:
Removing `CUDAQ_LIBRARY_MODE` is not the same as turning off library mode. Especially for these tests which are built with the C++ compiler and do not use the kernel builder. They will always be running in some form of library mode. The only way to eliminate library mode is to use the `nvq++` driver. I have asked @bettinaheim if we could discuss this in a meeting.

Renaud-K:
Ben has a better fix that makes the test use `nvq++`, so we do not have to rely on library mode.

PR #565, opened 2026-05-25
fix: To throw an error if no noise is provided to DEM Python binding functions

kaiqiy-nv:
For the DEM Python binding functions at [L541](https://github.com/NVIDIA/cudaqx/blob/d90d5b914baf2341333b98b24c9754b691aef268/libs/qec/python/bindings/py_code.cpp#L541), [L567](https://github.com/NVIDIA/cudaqx/blob/d90d5b914baf2341333b98b24c9754b691aef268/libs/qec/python/bindings/py_code.cpp#L567) and [L593](https://github.com/NVIDIA/cudaqx/blob/d90d5b914baf2341333b98b24c9754b691aef268/libs/qec/python/bindings/py_code.cpp#L593), the doc says "The noise model is optional and defaults to no noise.", but if the noise is really not provided, it becomes `std::nullopt` in the lambda function, and dereferencing `std::nullopt` is an undefined behaviour.

This PR tries to fix this bug in a way that will throw an error message if no noise model is provided.
Thanks for reviewing it.

### DIFF
--- a/libs/qec/python/bindings/py_code.cpp
+++ b/libs/qec/python/bindings/py_code.cpp
@@ -538,6 +538,10 @@ void bindCode(nb::module_ &mod) {
       "dem_from_memory_circuit",
       [](code &code, operation op, std::size_t numRounds,
          std::optional<cudaq::noise_model> noise = std::nullopt) {
+        if (!noise)
+          throw std::runtime_error(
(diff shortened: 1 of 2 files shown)


bmhowe23 (approved):
Thanks, @kaiqiy-nv.

PR #569, opened 2026-05-28
Add nightly build using latest cuda-quantum main

anjbur:
## Description

Adds a scheduled nightly run against the latest commit on `cuda-quantum/main`, using the full PR workflow build and test coverage: all libs, QEC, solvers, and docs.

Changes:
- Expands the `get-cudaq-version` action to accept an optional repo/ref override, rather than solely reading the `.cudaq_version` file. Similarly expands the `build-lib` action so that the QEC library's realtime CUDA-Q build can use this same override.
- Makes `all_libs.yaml`, `lib_qec.yaml`, and `lib_solvers.yaml` workflows callable from outside a PR context. To enable this, replaces the PR-specific CUDA-Q cache key with a generic cache key suffix
- Allows docs builds triggered from the PR workflow to use the shared PR CUDA-Q build cache key
- Adds a docs workflow input controlling whether docs are deployed after build, preventing nightly runs on `main` from accidentally deploying docs built against a CUDA-Q version other than the one pinned on `main`.
- Moves `apt update` earlier in build steps and adds missing ones. This fixes nightly workflow failures caused by repo/ref overrides skipping the old `get-cudaq-version` path that previously refreshed apt metadata. Adds `jq` install to the QEC build workflow for the same reason.

### Test runs of nightly pipeline:
- Latest run, against b6b3592. Failing in an expected way, due to latest changes on `cuda-quantum`. https://github.com/NVIDIA/cudaqx/actions/runs/27040551724
- Run against 7902410 (and against 190806e on `cuda-quantum` side). Passing. https://github.com/NVIDIA/cudaqx/actions/runs/26835136003
- Limited debugging run to check new cache cleanup step. Old keys were indeed deleted. Against 
3e3b39c. https://github.com/NVIDIA/cudaqx/actions/runs/27043949776/job/79825922678

No CI build time impact other than the overnight scheduled workflow.

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


.github/workflows/all_libs.yaml
```diff
@@ -24,17 +36,6 @@ jobs:
...
-        id: get-pr-info
-        env:
-          GH_TOKEN: ${{ github.token }}
-        uses: nv-gha-runners/get-pr-info@main
-
-      - name: Export PR info
-        id: export-pr-info
-        run: |
-          echo "pr_number=${{ fromJSON(steps.get-pr-info.outputs.pr-info).number }}" >> $GITHUB_OUTPUT
-
```
anjbur, line 37: Refactored into `pr_workflow.yaml`, where it's used to generate the `cache_key_suffix` value passed to this workflow.

.github/workflows/lib_qec.yaml
```diff
@@ -24,17 +36,6 @@ jobs:
...
-        id: get-pr-info
-        env:
-          GH_TOKEN: ${{ github.token }}
-        uses: nv-gha-runners/get-pr-info@main
-
-      - name: Export PR info
-        id: export-pr-info
-        run: |
-          echo "pr_number=${{ fromJSON(steps.get-pr-info.outputs.pr-info).number }}" >> $GITHUB_OUTPUT
-
```
anjbur, line 37: Refactored into `pr_workflow.yaml`, where it's used to generate the `cache_key_suffix` value passed to this workflow.

.github/workflows/lib_solvers.yaml
```diff
@@ -24,17 +36,6 @@ jobs:
...
-        id: get-pr-info
-        env:
-          GH_TOKEN: ${{ github.token }}
-        uses: nv-gha-runners/get-pr-info@main
-
-      - name: Export PR info
-        id: export-pr-info
-        run: |
-          echo "pr_number=${{ fromJSON(steps.get-pr-info.outputs.pr-info).number }}" >> $GITHUB_OUTPUT
-
```
anjbur, line 37: Refactored into `pr_workflow.yaml`, where it's used to generate the `cache_key_suffix` value passed to this workflow.

.github/workflows/nightly_tests.yaml
bmhowe23, line 1: It looks like it uses the cache while running this job this, which makes sense, but can you please clarify whether the cache is cleaned? I'm a bit worried that if we run this night after night that the cache will fill up and then purge the CUDA-Q build that is still referenced in .cudaq_version on `main`.

anjbur: Great point, thank you. I've disabled CUDA-Q ccache saves for the nightly runs so that we don't keep compiler caches from unpinned `cuda-quantum` versions. I've also added a cleanup step so that all CUDA-Q build caches and library ccaches are deleted after each nightly run. You can see that diff [here](https://github.com/NVIDIA/cudaqx/pull/569/changes/b956ae6f8fcf31b5aea17f9118fb5d6ef44a584f..a82ed63cdeca9568dca77aea3ff5b85349cffbe3).

I'm also investigating some other cache cleanup bugs/inefficiencies and will handle those in a follow-up PR.

bmhowe23 (approved):
Thanks, @anjbur!

PR #546, opened 2026-05-15
Add chromobius decoder to the decoder plugins

wsttiger:
## Summary

Adds a CUDA-QX QEC decoder plugin for Chromobius and wires it into the existing decoder factory/plugin flow.

Chromobius is DEM-native, so the public construction path uses raw Stim detector error model text instead of a parity-check matrix:

```cpp
cudaq::qec::get_decoder("chromobius", std::string_view{dem_text}, params)
```

```python
qec.get_decoder("chromobius", dem_text, **params)
```

The wrapper compiles a Chromobius decoder from the Stim DEM, bit-packs detector syndromes into Chromobius' expected input format, and returns predicted observable flips directly.

## Changes

- Add `chromobius` decoder plugin under `libs/qec/lib/decoders/plugins/chromobius`
- Wire the plugin into the QEC library build
- Fetch/build Chromobius as an internal plugin dependency
- Register Chromobius through the `decoder_init` factory path
- Require DEM text initialization for Chromobius and reject PCM construction with a clear error
- Expose Chromobius behavior options through CUDA-QX decoder params
- Return observable predictions directly and configure observable plumbing accordingly
- Add C++ unit tests for construction, decode behavior, PCM rejection, and malformed DEM handling
- Add small upstream-derived Chromobius reference data with provenance documentation
- Add C++ and Python correctness tests for single-shot and batched decode paths
- Keep the GCC 12 GTest workaround private so it does not leak into CUDA test builds

## Testing

Local gb200 verification:

- `ninja -C build`
- `ctest --test-dir build -R "Chromobius|ChromobiusCorrectness" --output-on-failure`
- `pytest -q libs/qec/python/tests/test_chromobius.py`
- `pytest -q libs/qec/python/tests`
- `pytest -q libs/qec/python/tests/test_code.py::test_noisy_simulation`
- `git diff --check`
- `clang-format --dry-run --Werror ...`

GitHub Actions PR checks are passing, including QEC, QEC GPU tests, All libs, Solvers, Docs, formatting, DCO, and Verify PR.

PR #571, opened 2026-05-28
Support Stim DEM strings in get_decoder

vedika-saravanan:
## Description

Adds unified decoder initialization from either a parity-check matrix or raw Stim detector error model text.

This supports both existing PCM-based decoders and DEM-native decoders needed by #546. PCM-based decoders can parse DEM text into `H`, `O`, and `error_rate_vec` defaults, while DEM-native decoders can consume the raw DEM text without going through a lossy matrix conversion.

### API (`cudaq/qec/decoder.h`)

- Adds `decoder_init = std::variant<sparse_binary_matrix, std::string>` as the decoder registry construction input.
- Keeps existing PCM construction paths:
  - `get_decoder(name, H, options)`
  - `decoder::get(name, H, options)`
- Adds DEM string construction through the same entry points:
  - `get_decoder(name, dem_text, options)`
  - `decoder::get(name, dem_text, options)`
- Adds `dem_from_stim_text(dem_text)` to parse Stim DEM text into `detector_error_model`.
- Adds helper routing for PCM-based decoders so DEM-derived `O` and `error_rate_vec` are supplied as defaults when not explicitly provided by the user.

The DEM-to-detector_error_model parser is lossy: it extracts detector flips, observable flips, and per-error probabilities, but drops detector coordinates and separator-encoded correlation structure. DEM-native decoders should consume the raw string alternative in decoder_init.

Python: `cudaq_qec.get_decoder(...)` now accepts Stim DEM strings. Python-registered decoders currently receive parsed `H` plus DEM-derived `O` / `error_rate_vec` defaults, not raw DEM text.

### Dependency / build

QEC now provides its own Stim dependency via `FetchContent` when `libstim` is not already available, so standalone QEC builds do not depend on parent CUDA-Q interim build artifacts.

### Tests

C++ gtests and Python tests cover:

- DEM string construction through `get_decoder(...)`
- PCM construction still working
- user-provided options overriding DEM-derived defaults
- DEM parse edge cases
- DEMs without observables
- `stim::DemTarget` category assumptions

### Out of scope / follow-ups

Chromobius plugin integration itself; detector-coordinate storage on `detector_error_model`; optional PyMatching-specific improvements; user-facing Sphinx/RST docs.

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -427,4 +430,38 @@ inline void convert_vec_hard_to_soft(const std::vector<std::vector<t_hard>> &in,
 std::unique_ptr<decoder>
 get_decoder(const std::string &name, const cudaqx::tensor<uint8_t> &H,
             const cudaqx::heterogeneous_map options = {});
+
+struct detector_error_model;
+
+/// @brief DEM-derived defaults; pointers alias into the source `dem`.
+struct dem_default_values {
+  const cudaqx::tensor<uint8_t> *O = nullptr;
```
bmhowe23, line 524: I'm open to feedback, but this makes me think that we should merge #550 first and then update these changes to use the new type before merging this PR. Otherwise, we might forget parts like this that need to change. Thoughts?

libs/qec/lib/CMakeLists.txt
```diff
@@ -60,6 +61,16 @@ add_library(${LIBRARY_NAME} SHARED ${QEC_SOURCES})
 add_subdirectory(decoders/plugins/example)
 add_subdirectory(decoders/plugins/pymatching)
 
+# libstim comes from the parent build (CUDA-Q).
```
bmhowe23: Perhaps I'm missing something, but I'm a bit confused how this is working because I didn't think that CUDA-Q was redistributing `libstim`. Is it possible `libstim` is coming from somewhere other than the parent CUDA-Q build?

vedika-saravanan: libstim is not redistributed as a standalone CUDA-Q library. CUDA-Q vendors Stim as a submodule and adds it to the same CMake build via `add_subdirectory(tpls/Stim ...)`, which creates the upstream libstim target. This QEC target is just consuming that in-tree target during the combined build and linking it privately/static into libcudaq-qec.

I’ll update the comment to make that clearer, e.g. “libstim is defined by the vendored Stim submodule added by the parent CUDA-Q build.”

libs/qec/include/cudaq/qec/decoder.h
vedika-saravanan: I agree #550 should land first because it builds on the dense tensor<uint8_t> decoder signature that #550 swaps out for sparse_binary_matrix. Probably cleanest if #550 goes in first and I rebase this on top.

libs/qec/lib/CMakeLists.txt
bmhowe23: OK, that makes sense, but I think we should be able to build in a system or container without requiring interim build artifacts from CUDA-Q. (For example, we should be able to build using the nightly CUDA-Q image, which does not have the interim build artifacts in it.) Could you please just add our own Stim via FetchContent in our CMakeLists.txt files?

libs/qec/include/cudaq/qec/decoder.h
```diff
@@ -427,4 +430,38 @@ inline void convert_vec_hard_to_soft(const std::vector<std::vector<t_hard>> &in,
...
+/// User-supplied values for either key win over the DEM-derived ones.
+///
+/// @note Lossy: detector annotations, decomposition separators, and
+/// `error_ids` are dropped. Sufficient for matching-style / H-based
+/// decoders (LUT, NV, sliding_window, TRT, PyMatching). Decoders that
+/// need full DEM metadata (e.g. Chromobius detector color/basis) require
+/// the planned \c detector_coords extension on
+/// \c cudaq::qec::detector_error_model; tracked as a follow-up.
+std::unique_ptr<decoder>
+get_decoder_from_stim_dem(const std::string &name,
```
bmhowe23: Per our offline conversation, I think this function could be removed if we change the decoder extension point / decoder constructors to accept `std::variant<tensor, const char *>` or `std::variant<tensor, string_view>`.

libs/qec/lib/CMakeLists.txt
```diff
@@ -58,6 +60,25 @@ list(APPEND QEC_SOURCES
...
+
+if(NOT TARGET libstim)
+  message(FATAL_ERROR
+    "Stim FetchContent did not provide the libstim target.")
+endif()
+target_link_libraries(${LIBRARY_NAME} PRIVATE libstim)
+target_link_options(${LIBRARY_NAME} PRIVATE
+  $<$<OR:$<CXX_COMPILER_ID:GNU>,$<CXX_COMPILER_ID:Clang>>:-Wl,--exclude-libs,libstim.a>
+)
+
```
bmhowe23, line 84: I downloaded this PR and built locally (using 92d3dc977b46035cfd228ebf71e1b6cb524e9951), and I see the following, which could be problematic:

```bash
$ nm -D lib/libcudaq-qec.so | grep stim | c++filt
00000000000907d0 W stim::DetectorErrorModel::~DetectorErrorModel()
00000000000907d0 W stim::DetectorErrorModel::~DetectorErrorModel()
```

The above two items are the only problematic items, so it's possible that your solution here covers most of the issue, but I don't think we should be leaking those symbol out.

In order to prevent something like https://github.com/NVIDIA/cuda-quantum/issues/3045, I think something like https://github.com/NVIDIA/cuda-quantum/pull/3076 is required.

libs/qec/unittests/test_decoders.cpp
```diff
@@ -743,63 +745,136 @@ TEST(DecoderTest, GetBlockSizeAndSyndromeSize) {
   EXPECT_EQ(decoder2->get_syndrome_size(), new_syndrome_size);
 }
 
-TEST(DecoderRegistryTest, SingleParameterRegistryDirect) {
```
bmhowe23, line 746: Just checking - does constructing a decoder w/o heterogenous map no longer supported? I don't think this is very common, so it's probably ok if we deprecate it; I just wanted to double check because if it still works then I'd want to keep this test.

bmhowe23 (commented):
This is looking good...thanks, @vedika-saravanan. Just a few follow-up questions below.

libs/qec/unittests/test_decoders.cpp
vedika-saravanan: Yes, constructing through the public API without an explicit heterogeneous_map is still supported; the overloads default to an empty map. The removed test was exercising the old single-argument extension_point registry directly, which no decoder registered into and which passed by expecting a runtime_error. I renamed the existing public-path test to make this coverage explicit: `DecoderTest.GetWithoutOptionsSetsBlockAndSyndromeSize` ([7b30f08](https://github.com/NVIDIA/cudaqx/pull/571/commits/7b30f089f03a08b27678816de16c7ad35b245e03)).

libs/qec/lib/CMakeLists.txt
vedika-saravanan: Thanks for the feedback @bmhowe23. I added `VISIBILITY_INLINES_HIDDEN ON` for `cudaq-qec` and rebuilt locally. The symbol check now returns no Stim exports:. Fixed in [7b30f08](https://github.com/NVIDIA/cudaqx/pull/571/commits/7b30f089f03a08b27678816de16c7ad35b245e03)

bmhowe23 (approved):
Thanks! Could you please create a follow-up PR/issue to add docs for this capability? We should add a public example for this.

vedika-saravanan:
GHI for adding docs and example https://github.com/NVIDIA/cudaqx/issues/595

bmhowe23:
@vedika-saravanan can you please do `ninja zip_installed_files` and inspect the zip file that it produces? I am seeing many stim binaries and directories that should not be "installed", and I suspect this PR is the one that added them.

PR #576, opened 2026-05-31
Fix Python build issues due to upstream CUDA-Q changes

bmhowe23:


.github/workflows/scripts/build_cudaq.sh
```diff
@@ -132,6 +132,19 @@ diff --git a/python/runtime/cudaq/domains/plugins/CMakeLists.txt b/python/runtim
 +      nanobind-static Python3::Module
        cudaq-chemistry cudaq-operator cudaq cudaq-py-utils cudaq-platform-default)
  endif()
+
```
bmhowe23, line 135: We won't need this patch once https://github.com/NVIDIA/cuda-quantum/pull/4643 lands.

PR #577, opened 2026-06-01
fix: To add clamp for torch.log(...) to cope with TF32

kaiqiy-nv:
This PR tries to fix [[B] 6250652](https://nvbugspro.nvidia.com/bug/6250652) .

Thank you for looking at this PR.


bmhowe23:
@vedika-saravanan can you please review?

vedika-saravanan (commented):
Thanks @kaiqiy-nv, LGTM. Before we merge the PR can we add a self-contained regression test inside the test body, so CI catches this without depending on which other files are collected?

kaiqiy-nv:
Hello @vedika-saravanan , I have added one new test case that can expose this issue on its own.

vedika-saravanan (approved):

PR #580, opened 2026-06-03
Bumping CUDA-Q version

sacpis:
Bumping CUDA-Q version

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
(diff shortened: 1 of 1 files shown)


bmhowe23 (commented):
Thanks, @sacpis! I kicked off https://github.com/NVIDIA/cudaqx/actions/runs/26891106188 (our Build Wheels job) just to be more thorough with the testing. I'd like to see that pass before approving. (We really need to make that check automatic in the CI, at least for PR's that modify the .cudaq_version file.)

bmhowe23 (approved):

PR #589, opened 2026-06-07
Fix dense --> sparse conversion in `get_decoder` to avoid redundant copies

bmhowe23:
## Description

Replace the `pcmToTensor`-based `make_sparse_from_dense` helper with a direct scan of the raw numpy buffer. This makes use of the new `sparse_binary_matrix` type. This avoids an intermediate dense tensor copy, so peak memory is just the caller's dense array plus the sparse output. The inner loop always runs in the contiguous memory direction, so both layouts are cache-friendly.

Update the Fortran-order test to reflect that F-order input is now handled correctly rather than rejected. Therefore, I removed the ` test_decoder_initialization_with_error()` test and replaced it with a new test to verify it works correctly.

If the user is providing dense PCMs, this allows them to use larger dense PCMs before running out of memory on their system by avoiding unnecessary allocations and copies

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size
      (if so, an issue has been raised). **See note above**

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/libs/qec/python/bindings/py_decoder.cpp
+++ b/libs/qec/python/bindings/py_decoder.cpp
@@ -90,6 +90,54 @@ sparse_binary_matrix_from_py_dict(const nb::dict &d) {
       "Sparse H dict layout must be \"nested_csc\" or \"nested_csr\".");
 }
 
+/// Convert a dense 2-D NumPy uint8 array to sparse_binary_matrix without
+/// any intermediate dense tensor allocation.  Strides are read directly so
+/// both C-contiguous (row-major) and Fortran-contiguous (column-major) arrays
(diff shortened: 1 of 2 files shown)


vedika-saravanan:
Thanks @bmhowe23 , this looks like the right direction. In #550 we moved this path through the existing `pcmToTensor` helper to centralize validation and stride handling, but that helper is not a good fit for the sparse bridge because it constructs a dense `cudaqx::tensor` before borrowing the NumPy buffer. For large dense PCMs that can add a multi-GB transient allocation before decoder construction.

vedika-saravanan (approved):

PR #590, opened 2026-06-07
Adopt scipy.sparse as optional interop

bmhowe23:
## Description

- Remove the custom dict-based sparse H input from `get_decoder`; scipy.sparse matrices (any format) are now the sole sparse input path, detected via duck typing so scipy remains an interoperability target, not a true dependency. (Dense numpy arrays can still be used.)
- Since I want to avoid the custom dict type as a public interface, remove `generate_random_pcm_sparse` Python binding. This function was returning the custom dict type that I don't think we want to maintain going forward. Users can wrap the dense `generate_random_pcm` output in scipy.sparse themselves if they want. We will also probably want to update the interface for `generate_random_pcm` in the future, so I don't want to have a separate function for something that does the same thing. In reality, we will be moving towards more realistic PCM construction techniques anyway, so generating random ones will be less useful.
- Replace PyDecoder(dict) constructor with PyDecoder(scipy_sparse) so that BYOD decoder classes can call qec.Decoder.__init__(self, H) with a scipy sparse matrix
- Update all affected tests: remove dict-helper functions, replace dict-based sparse tests with scipy.sparse equivalents, remove duplicate-sparse test (scipy always deduplicates on tocsr())

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size
      (if so, an issue has been raised). **See above**

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/libs/qec/lib/pcm_utils.cpp
+++ b/libs/qec/lib/pcm_utils.cpp
@@ -511,9 +511,32 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
         std::to_string(weight) +
         ", n_syndromes_per_round=" + std::to_string(n_syndromes_per_round));
   }
-
+  constexpr auto kSizeMax = std::numeric_limits<std::size_t>::max();
+  // Overflow-safe cap check in (a > limit / b) form; otherwise size_t wrap
+  // could bypass the cap.
+  auto exceeds_cap = [](std::size_t a, std::size_t b) {
+    return a != 0 && b > kSizeMax / a;
+  };
+  if (exceeds_cap(n_rounds, n_errs_per_round) ||
+      exceeds_cap(n_rounds, n_syndromes_per_round)) {
+    throw std::invalid_argument(
+        "generate_random_pcm: n_rounds * n_errs_per_round and "
+        "n_rounds * n_syndromes_per_round must not overflow size_t; got "
+        "n_rounds=" +
+        std::to_string(n_rounds) +
+        ", n_errs_per_round=" + std::to_string(n_errs_per_round) +
+        ", n_syndromes_per_round=" + std::to_string(n_syndromes_per_round) +
+        ". Use cudaq::qec::generate_random_pcm_sparse(...) instead.");
+  }
   std::size_t n_cols = n_rounds * n_errs_per_round;
   std::size_t n_rows = n_rounds * n_syndromes_per_round;
+  if (exceeds_cap(n_rows, n_cols)) {
+    throw std::invalid_argument(
+        std::string("generate_random_pcm: total dense size n_rows * n_cols (") +
+        std::to_string(n_rows) + " × " + std::to_string(n_cols) +
+        ") must not overflow size_t; use "
+        "cudaq::qec::generate_random_pcm_sparse(...) instead.");
+  }
 
   cudaqx::tensor<uint8_t> pcm(std::vector<std::size_t>{n_rows, n_cols});
 
--- a/libs/qec/lib/pcm_utils.cpp
+++ b/libs/qec/lib/pcm_utils.cpp
@@ -511,34 +511,9 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
         std::to_string(weight) +
         ", n_syndromes_per_round=" + std::to_string(n_syndromes_per_round));
   }
-  // Overflow-safe cap check in (a > limit / b) form; otherwise size_t wrap
-  // could bypass the cap.
-  auto exceeds_cap = [](std::size_t a, std::size_t b) {
-    return a != 0 && b > k_max_dense_pcm_elements / a;
-  };
-  if (exceeds_cap(n_rounds, n_errs_per_round) ||
-      exceeds_cap(n_rounds, n_syndromes_per_round)) {
-    throw std::invalid_argument(
-        std::string("generate_random_pcm: n_rounds * n_errs_per_round and "
-                    "n_rounds * n_syndromes_per_round must each fit within "
-                    "the dense allocation limit (") +
-        std::to_string(k_max_dense_pcm_elements) +
-        "); got n_rounds=" + std::to_string(n_rounds) +
-        ", n_errs_per_round=" + std::to_string(n_errs_per_round) +
-        ", n_syndromes_per_round=" + std::to_string(n_syndromes_per_round) +
-        ". Use cudaq::qec::generate_random_pcm_sparse(...) instead.");
-  }
+
   std::size_t n_cols = n_rounds * n_errs_per_round;
   std::size_t n_rows = n_rounds * n_syndromes_per_round;
-  if (exceeds_cap(n_rows, n_cols)) {
-    throw std::invalid_argument(
-        std::string("generate_random_pcm: PCM size (") +
-        std::to_string(n_rows) + " × " + std::to_string(n_cols) +
-        " elements) exceeds the dense allocation limit (" +
-        std::to_string(k_max_dense_pcm_elements) +
-        ") for this API; use cudaq::qec::generate_random_pcm_sparse(...) "
-        "instead.");
-  }
 
   cudaqx::tensor<uint8_t> pcm(std::vector<std::size_t>{n_rows, n_cols});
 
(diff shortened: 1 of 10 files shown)


libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -772,8 +785,8 @@ void bindDecoder(nb::module_ &mod) {
 
         cudaq::qec::sparse_binary_matrix H_sparse;
 
-        if (nb::isinstance<nb::dict>(H))
-          H_sparse = sparse_binary_matrix_from_py_dict(nb::cast<nb::dict>(H));
+        if (nb::hasattr(H, "indptr") && nb::hasattr(H, "indices"))
```
vedika-saravanan: This only detects CSR/CSC-like scipy matrices. COO and some other scipy sparse formats do not expose `indptr`, even though `sparse_binary_matrix_from_scipy()` normalizes with `.tocsr()`. Could we detect sparse input via `hasattr(H, "tocsr")` instead and add a COO test?

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -54,40 +55,51 @@ checked_narrow_to_index_type(std::size_t value, const char *field_name) {
...
-        "Sparse H dict must have keys: layout, num_rows, num_cols, "
-        "nested. Use layout \"nested_csc\": nested has one list per COLUMN "
-        "with ROW indices where that column has a one; \"nested_csr\": one "
-        "list per ROW with COLUMN indices for that row.");
-  std::string layout = nb::cast<std::string>(d["layout"]);
+/// Build sparse_binary_matrix directly from a scipy sparse matrix.
+/// Any scipy sparse format is accepted; it is normalized to CSR internally.
+static sparse_binary_matrix sparse_binary_matrix_from_scipy(nb::object mat) {
+  // Normalize to CSR so that indptr == row_offsets, indices == col_indices.
+  nb::object csr = mat.attr("tocsr")();
```
vedika-saravanan, line 62: Should we call `eliminate_zeros()` on this CSR copy? Since we copy only `indptr` and `indices`, explicit stored zeros in a scipy sparse matrix are treated as nonzero PCM entries.

libs/qec/lib/pcm_utils.cpp
```diff
@@ -511,34 +511,9 @@ cudaqx::tensor<uint8_t> generate_random_pcm(std::size_t n_rounds,
...
-                    "n_rounds * n_syndromes_per_round must each fit within "
-                    "the dense allocation limit (") +
-        std::to_string(k_max_dense_pcm_elements) +
-        "); got n_rounds=" + std::to_string(n_rounds) +
-        ", n_errs_per_round=" + std::to_string(n_errs_per_round) +
-        ", n_syndromes_per_round=" + std::to_string(n_syndromes_per_round) +
-        ". Use cudaq::qec::generate_random_pcm_sparse(...) instead.");
-  }
+
   std::size_t n_cols = n_rounds * n_errs_per_round;
```
vedika-saravanan, line 531: Removing the arbitrary dense cap makes sense, but we should keep overflow-safe multiplication checks here. `n_rounds * n_errs_per_round` and `n_rounds * n_syndromes_per_round` can wrap before allocation/indexing.

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -142,20 +154,22 @@ class PyDecoder : public decoder {
...
-      : decoder([&H]() {
+  /// @brief Construct from a scipy sparse matrix (CSR, CSC, COO, ...) or a
+  ///        dense numpy array of any numeric dtype.
+  PyDecoder(nb::object mat)
+      : decoder([&mat]() -> cudaq::qec::sparse_binary_matrix {
+          if (nb::hasattr(mat, "indptr") && nb::hasattr(mat, "indices"))
+            return sparse_binary_matrix_from_scipy(mat);
+          // Dense numpy array of any dtype
+          auto H = nb::cast<nb::ndarray<nb::numpy, uint8_t>>(
+              mat.attr("astype")("uint8"));
```
vedika-saravanan: For dense NumPy BYOD inputs this still materializes dense storage via `astype`, `toTensor`, and `owned.copy`. Should this reuse `make_sparse_from_dense()` so `qec.Decoder.__init__(self, H)` has the same memory behavior as native `get_decoder(..., H)`?

libs/qec/python/bindings/py_decoder.cpp
bmhowe23: Yes, good point. Addressed in 0b9354ee9a1453b93fa8ff70ed97f4b42cbef151.

libs/qec/python/bindings/py_decoder.cpp
bmhowe23: Good point. Addressed in 0b9354ee9a1453b93fa8ff70ed97f4b42cbef151.

libs/qec/lib/pcm_utils.cpp
bmhowe23: Updated in 0b9354ee9a1453b93fa8ff70ed97f4b42cbef151

libs/qec/python/bindings/py_decoder.cpp
bmhowe23: Updated in 0b9354ee9a1453b93fa8ff70ed97f4b42cbef151

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -60,6 +60,12 @@ checked_narrow_to_index_type(std::size_t value, const char *field_name) {
 static sparse_binary_matrix sparse_binary_matrix_from_scipy(nb::object mat) {
   // Normalize to CSR so that indptr == row_offsets, indices == col_indices.
   nb::object csr = mat.attr("tocsr")();
+  // Drop explicitly-stored zeros: we only copy indptr/indices below, so a
+  // structural zero would otherwise be treated as a nonzero PCM entry.
+  // tocsr() returns the same object when mat is already CSR, so copy first to
+  // avoid mutating the caller's matrix.
+  csr = csr.attr("copy")();
+  csr.attr("eliminate_zeros")();
```
vedika-saravanan, line 66: Can we also canonicalize the scipy CSR structure here with `sum_duplicates()` and `sort_indices()` before copying `indptr`/`indices`? `eliminate_zeros()` handles explicit stored zeros, but scipy CSR can still carry duplicate or unsorted indices. Since we ignore `data` after this point, duplicates/unsorted entries can make sparse input behave differently from the equivalent dense matrix or get rejected later by consumers such as pymatching that require sorted-unique indices.

libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -158,16 +164,16 @@ class PyDecoder : public decoder {
...
-          cudaqx::tensor<uint8_t> owned(borrow.shape());
-          owned.copy(borrow.data(), borrow.shape());
-          return cudaq::qec::sparse_binary_matrix(
-              owned, cudaq::qec::sparse_binary_matrix_layout::csc);
+          // Dense numpy array of any dtype: build sparse storage directly so
+          // qec.Decoder.__init__(self, H) has the same memory behavior as
+          // native get_decoder(..., H) (no intermediate dense tensor copy).
+          return make_sparse_from_dense(
+              nb::cast<nb::ndarray<nb::numpy, uint8_t>>(
+                  mat.attr("astype")("uint8")));
```
vedika-saravanan: Follow-up, non-blocking: the dense BYOD path still calls `astype("uint8")`, which copies even when the input is already uint8. It would be good to switch that to a no-copy cast path later so `qec.Decoder.__init__(self, H)` avoids duplicating dense inputs when possible.

vedika-saravanan:
Thanks for the updates @bmhowe23.

LGTM. I left one comment on the scipy sparse import path that I think should be handled in this PR, since it affects sparse-input correctness for valid scipy CSR matrices. I also noted one dense BYOD memory optimization as a non-blocking follow-up.

libs/qec/python/bindings/py_decoder.cpp
bmhowe23: Addressed in efa416d

libs/qec/python/bindings/py_decoder.cpp
bmhowe23: Addressed in efa416d

vedika-saravanan (approved):
Thanks @bmhowe23, LGTM.

PR #592, opened 2026-06-07
Fix PySCF startup timeout

fallintoplace:
## Summary
- Require PySCF health checks to return `status: available` before the driver reports ready.
- Keep startup polling through REST connection failures until the timeout expires.
- Drop unused local variables in the startup path.

## Why
When `RestClient::get()` throws while the PySCF server is still starting or unreachable, the previous loop continued before incrementing or checking the timeout counter. That could hang forever on connection failures. The availability checks also treated malformed or non-ready responses as available.

## Validation
- `git diff --check`

Not run locally: full solvers build, because this checkout does not have a configured CUDA-Q/LLVM build environment.

### DIFF
--- a/libs/solvers/lib/operators/molecule/drivers/pyscf_driver.cpp
+++ b/libs/solvers/lib/operators/molecule/drivers/pyscf_driver.cpp
@@ -23,6 +23,13 @@
 
 using namespace cudaqx;
 
+namespace {
+bool hasAvailableStatus(const nlohmann::json &metadata) {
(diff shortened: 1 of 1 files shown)


bmhowe23:
Hi @fallintoplace - sorry for the slow response. CUDA-Q Solvers has been deprecated in favor of [CUDA-Q Algorithms](https://github.com/NVIDIA/cudaq-algorithms). CUDA-Q Algorithms does not use this PySCF driver, so barring any additional follow-ups, I'd recommend closing this PR.

PR #600, opened 2026-06-09
Add PyMatching HOST_CALL decoder-server RPC path

cketcham2333:
## Summary
This adds a public, CPU-only reference path for the decoder-server runtime message trio using PyMatching and `CUDAQ_DISPATCH_HOST_CALL`.
The new path introduces shared QEC realtime RPC definitions, a small in-process HOST_CALL session, and an RPC producer that writes `enqueue_syndromes`, `get_corrections`, and `reset_decoder` messages through a same-process RX/TX ring. The HOST_CALL handlers reuse the existing `decoder` base accumulator methods, so PyMatching does not need decoder-specific realtime overrides.
This intentionally does not add CUDA dispatch files, surface-code integration, env-var production routing, or any private nv-qldpc pieces.
## What changed
- Adds decoder-server runtime RPC IDs and wire payload helpers.
- Adds a host-only realtime session backed by `cudaq_host_dispatcher_loop`.
- Adds an RPC producer for the three-message flow.
- Adds PyMatching realtime tests that mirror existing PyMatching behavioral cases over the RPC path and verify reset clears accumulated corrections.
## Test plan
```bash
ctest -R "PyMatchingDecoder|PyMatchingRealtime"

### DIFF
--- a/libs/qec/lib/realtime/CMakeLists.txt
+++ b/libs/qec/lib/realtime/CMakeLists.txt
@@ -53,6 +53,7 @@ if(CMAKE_CUDA_COMPILER)
     )
     if(CUDAQ_REALTIME_LIBRARY)
       message(STATUS "Found cuda-quantum realtime library at ${CUDAQ_REALTIME_LIBRARY}")
+      get_filename_component(_CUDAQ_REALTIME_LIB_DIR "${CUDAQ_REALTIME_LIBRARY}" DIRECTORY)
     endif()
     
     # Create static device library for device code
@@ -140,6 +141,13 @@ target_link_libraries(cudaq-qec-realtime-decoding
 
 set_target_properties(cudaq-qec-realtime-decoding PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+
+if(_CUDAQ_REALTIME_LIB_DIR)
+  set_target_properties(cudaq-qec-realtime-decoding PROPERTIES
+    BUILD_RPATH "${_CUDAQ_REALTIME_LIB_DIR};${CMAKE_BINARY_DIR}/lib"
+    INSTALL_RPATH "${_CUDAQ_REALTIME_LIB_DIR};${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}"
+  )
+endif()
 
 
 install(TARGETS cudaq-qec-realtime-decoding
--- a/libs/qec/lib/realtime/CMakeLists.txt
+++ b/libs/qec/lib/realtime/CMakeLists.txt
@@ -55,6 +55,19 @@ if(CMAKE_CUDA_COMPILER)
       message(STATUS "Found cuda-quantum realtime library at ${CUDAQ_REALTIME_LIBRARY}")
       get_filename_component(_CUDAQ_REALTIME_LIB_DIR "${CUDAQ_REALTIME_LIBRARY}" DIRECTORY)
     endif()
+
+    # The host-side dispatch loop (cudaq_host_dispatcher_loop), used by
+    # qec_realtime_session, lives in libcudaq-realtime-host-dispatch -- separate
+    # from libcudaq-realtime.  Locate it so cudaq-qec-realtime-decoding can link
+    # it (mirrors the cudaq-realtime-pipeline target below).
+    find_library(CUDAQ_REALTIME_HOST_DISPATCH_LIBRARY
+      NAMES cudaq-realtime-host-dispatch
+      PATHS ${_cudaq_realtime_prefixes}
+      PATH_SUFFIXES lib
+    )
+    if(CUDAQ_REALTIME_HOST_DISPATCH_LIBRARY)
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: libs/qec/lib/realtime/CMakeLists.txt [hunk 1]
```diff
@@ -117,11 +136,19 @@ target_link_libraries(cudaq-qec-realtime-decoding
   PRIVATE
     cudaq::cudaq-common
     LLVMSupport
+    $<$<BOOL:${CUDAQ_REALTIME_LIBRARY}>:${CUDAQ_REALTIME_LIBRARY}>
```
vedika-saravanan, line 152: This target now calls `cudaq_host_dispatcher_loop`, but only links `cudaq-realtime`. Should this also find/link `cudaq-realtime-host-dispatch`, like the `cudaq-realtime-pipeline` target below? Otherwise realtime-enabled builds may fail to link or load.

libs/qec/lib/realtime/qec_realtime_session.cpp
```diff
@@ -0,0 +1,283 @@
...
+      return;
+    }
+
+    auto *decoder = get_decoder_or_throw(body->decoder_id);
+    const std::uint8_t *bits = reinterpret_cast<const std::uint8_t *>(body + 1);
+    std::vector<std::uint8_t> syndromes(num_syndromes, 0);
+    for (std::uint64_t bit = 0; bit < num_syndromes; ++bit)
+      syndromes[bit] = (bits[bit >> 3] >> (bit & 7)) & 0x1u;
+
+    (void)decoder->enqueue_syndrome(syndromes.data(), syndromes.size());
```
vedika-saravanan, line 106: This ACKs success even when `enqueue_syndrome` returns false, and there is no check that `num_syndromes` is <= `decoder->get_num_msyn_per_decode()`. Oversized requests can be accepted silently if they fit in the slot. Could we please reject invalid syndrome lengths and add a negative realtime test?

libs/qec/lib/realtime/rpc_producer.cpp
```diff
@@ -0,0 +1,270 @@
...
+  volatile std::uint64_t *rx = session.rx_flags_host();
+  volatile std::uint64_t *tx = session.tx_flags_host();
+  for (int waited = 0; waited < timeout_ms; ++waited) {
+    const std::uint32_t start =
+        g_next_slot_hint.load(std::memory_order_relaxed) %
+        static_cast<std::uint32_t>(session.num_slots());
+    for (std::uint32_t s = 0; s < session.num_slots(); ++s) {
+      const std::uint32_t slot =
+          static_cast<std::uint32_t>((start + s) % session.num_slots());
+      if (rx[slot] == 0 && tx[slot] == 0) {
```
vedika-saravanan: This is more fragile because `skip_tx_markers = 1` means `tx_flags` is not used as a busy marker; the producer only clears it in `release_slot`. So `tx[slot] == 0` does not protect the response buffer lifetime, and a slot can appear free once `rx_flags` is cleared even if the previous producer has not finished reading `tx_data`.

libs/qec/lib/realtime/qec_realtime_session.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+qec_realtime_session::~qec_realtime_session() { finalize(); }
+
+void qec_realtime_session::initialize() {
+  if (initialized_)
+    return;
+  try {
+    allocate_ring_buffer();
+    populate_function_table();
+    g_active_decoders = &decoders_;
```
vedika-saravanan: This makes decoder lookup process-global rather than session-scoped. If two `qec_realtime_session`s are initialized concurrently, the second one overwrites `g_active_decoders` and the first session's host loop will resolve decoder IDs against the wrong decoder vector. Can we either reject a second active session or pass the decoder vector/session context through the host handler path instead of using a global?

libs/qec/lib/realtime/CMakeLists.txt
cketcham2333: Resolved in commit SHA c82f0da0096b5d22d2b9dd0f6b989fe869e21ef9

libs/qec/lib/realtime/qec_realtime_session.cpp
cketcham2333: Resolved in commit SHA c82f0da0096b5d22d2b9dd0f6b989fe869e21ef9

libs/qec/lib/realtime/rpc_producer.cpp
cketcham2333: Resolved in commit SHA c82f0da0096b5d22d2b9dd0f6b989fe869e21ef9

libs/qec/lib/realtime/qec_realtime_session.cpp
cketcham2333: Resolved in commit SHA c82f0da0096b5d22d2b9dd0f6b989fe869e21ef9

vedika-saravanan (approved):

vedika-saravanan:
Thanks @cketcham2333 LGTM

PR #601, opened 2026-06-10
Prevent vendored dependency artifacts leakage

vedika-saravanan:
## Description

Cleans up CUDA-QX install/package artifacts so vendored dependency install rules do not leak into `installed_files.zip`.

This prevents third-party artifacts such as:

- `bin/pymatching`
- `lib/libpymatching.a`
- `include/pymatching/*`
- `lib/libfmt.a`
- `include/fmt/*`
- `lib/cmake/fmt/*`

from being included in the package overlay generated by `ninja zip_installed_files`.

Changes:
- Disables `fmt` install rules for the vendored `fmt` dependency.
- Marks vendored PyMatching `FetchContent` as `EXCLUDE_FROM_ALL` so its upstream install rules do not enter the top-level install manifest.
- Installs CUDA-QX core headers under the CUDA-QX install include directory instead of the CUDA-Q install include path.

This does not change runtime code or decoder behavior; it only affects install/package contents.

## Validation

Validated with:

```bash
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
ninja -C build install
ninja -C build zip_installed_files
unzip -l build/installed_files.zip | tee build/zip_contents_after.txt
```

Validated by regenerating `installed_files.zip` and checking both the install manifest and zip listing.

The leakage check now produces no output:

```bash
grep -iE 'bin/pymatching|lib/libpymatching\.a|include/pymatching|lib/libfmt\.a|include/fmt|lib/cmake/fmt|lib/pkgconfig/fmt\.pc|usr/local/cudaq' \
  build/install_manifest.txt build/zip_contents_after.txt
```

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      CMake/package behavior was validated by regenerating `installed_files.zip` and inspecting its contents.
- [x] Tests fail if the new functionality is broken.
      The package-content check fails if vendored artifacts remain in `installed_files.zip`.
      N/A, no exception behavior changed.
- [x] Truth data added where simple `EXPECT_*` / `assert` checks are insufficient.
      N/A, no algorithmic behavior changed.

### Documentation
      N/A, no public API changes.
- [x] User-visible behavior changes have public docs, or a follow-up is tracked.
      N/A, package cleanup only.

### Code style
- [x] Naming follows the existing convention (`snake_case` vs `camelCase`) for the area being modified.

### Dependencies

### DIFF
--- a/libs/core/CMakeLists.txt
+++ b/libs/core/CMakeLists.txt
@@ -37,6 +37,7 @@ option(CUDAQX_CORE_INCLUDE_TESTS
 # ==============================================================================
 
 include(FetchContent)
+include(GNUInstallDirs)
 
 FetchContent_Declare(
   xtl
@@ -70,4 +71,3 @@ if (CUDAQX_CORE_INCLUDE_TESTS)
   add_custom_target(CUDAQXCoreUnitTests)
   add_subdirectory(unittests)
 endif()
-
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,10 @@ project(CUDAQX
   VERSION 0.0.0
   LANGUAGES C CXX)
 
+# Do not let vendored dependencies add their own install rules to CUDA-QX
+# packages. CUDA-QX installs only its public artifacts.
+set(FMT_INSTALL OFF CACHE BOOL "Generate the fmt install target." FORCE)
+
 set(CUDAQX_ALL_LIBS "qec;solvers")
 set(CUDAQX_ENABLE_LIBS "all" CACHE STRING
   "Semicolon-separated list of libs to build (${CUDAQX_ALL_LIBS}), or \"all\".")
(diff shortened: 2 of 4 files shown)


libs/core/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ option(CUDAQX_CORE_INCLUDE_TESTS
 # ==============================================================================
 
 include(FetchContent)
+include(GNUInstallDirs)
```
bmhowe23, line 40: What effect does this change have?

vedika-saravanan: It brings in CMake’s standard install directory variables, especially `CMAKE_INSTALL_INCLUDEDIR`.

In this PR I changed the CUDA-QX core header install destination to use `CMAKE_INSTALL_INCLUDEDIR` instead of `CUDAQ_INCLUDE_DIR`. That makes the headers install under the CUDA-QX package prefix, e.g. `include/cuda-qx/...`, instead of leaking into the CUDA-Q install path like `usr/local/cudaq/include/cuda-qx/...`.

It should not affect build targets or compilation; it only provides the standard install-dir variable and respects any existing override.

bmhowe23 (approved):

PR #605, opened 2026-06-12
tests: Add unit tests for QEC

kaiqiy-nv:
This PR is tries to add more QEC unit tests across
- decoder configuration,
- realtime decoding paths,
- TensorRT decoder behaviour,
- sparse PCM utilities,
- DEM sampling,
- and Python decoder argument handling.

Tests have passed with logs available at [here](https://scdvsdsk.nvidia.com/jobs/24350/24350356/test_results.log) .

NOTE: This PR is for code coverage improvement only;  no bug is related.

Thank you for reviewing it.


kaiqiy-nv:
The newly added test `TRTDecoderTest.ParsePrecisionAcceptsSupportedAndLegacyValues` creates a TensorRT builder, which initialises CUDA. This CI test branch has no GPU, so the init fails.
CUDA 12.6 and 13.0 react differently on this:
CUDA 12.6 (TRT 10.7): createInferBuilder throws APIUsageError. With no try/catch, it went uncaught → std::terminate → abort.
CUDA 13.0 (TRT 10.14): createInferBuilder returns nullptr, handled by the existing if (!builder) GTEST_SKIP().


I have:
- added the standard if (!gpu_available()) GTEST_SKIP() guard,
- and wrapped builder/config creation in try/catch to handle the throwing variant.
So, the tests now skips on both 12.6 and 13.0 when there is no GPU, and runs normally where a GPU is present.

Please have another round of CI test on this, thanks!

bmhowe23 (approved):

PR #609, opened 2026-06-12
Add host-side in-process-RPC path for real-time QEC decoding

cketcham2333:
## Description
**What this adds.** A public, host-side path for **in-process RPC ("inproc_rpc")**
real-time QEC decoding. In this mode the host produces syndrome-decode requests into a
shared host/device ring buffer; they are serviced through cuda-quantum's on-GPU
`DEVICE_LOOP` dispatcher by a GPU-resident decoder (e.g. `nv-qldpc-decoder`), keeping
decode on the device with no CPU round-trip on the hot path. Selected at runtime via
`CUDAQ_QEC_REALTIME_MODE=inproc_rpc`; the existing host-call path is unchanged when it
is unset.
No proprietary device code lives in this repo. The GPU device-dispatch archive is
consumed **optionally** as an `IMPORTED` target via
`CUDAQ_QEC_REALTIME_CUDEVICE_PROPRIETARY_ARCHIVE` (whole-archive link); the realtime
graph/surface tests register only when it is supplied. (That archive is built
separately; see the companion GitLab MR.)
**Host components (new):**
- `qec_realtime_session.{h,cpp}` — shared host/device ring-buffer session.
- `rpc_producer.{h,cpp}` — host-side RPC request/response producer.
- `realtime_decoding.{h,cpp}` routes through the session/producer when
  `inproc_rpc` is selected; `config.cpp` / `decoding_config.h` plumbing.
**Build / linking:**
- `IMPORTED` proprietary-archive wiring in `lib/realtime/CMakeLists.txt` +
  `unittests/CMakeLists.txt` (whole-archive, test gating on target existence).
- Scope `--exclude-libs,ALL` to `PRIVATE` on the simulation/quantinuum realtime libs and
  add `--export-dynamic` for app examples, so the dispatch-API symbols stay
  `dlsym`-resolvable in test/app executables.
**Kernels:** `postprocess_*` XOR-accumulate corrections (`^=`) so multi-window shots
(`num_rounds > decoder_window`) reduce correctly; `test_gpu_kernels` zeroes the
accumulator and skips `ROUND_START` markers.
**Tests:** `test_realtime_qldpc_graph_decoding` (+ `qldpc_config_loader`) and a
`surface_code-1` inproc-rpc variant (`--use-relay-bp`); mock/hololink syndrome loaders
skip `ROUND_START`; refreshed `*_nv_qldpc_relay` fixtures.
**Release/build infra:** `all_libs_release.yaml` fails fast if the `nv-qldpc-decoder`
plugin or the proprietary archive is missing from the release asset; `build_all.sh` /
`build_qec.sh` build the realtime path.


libs/qec/lib/realtime/realtime_decoding.cpp
```diff
@@ -21,6 +34,116 @@ SyndromeCaptureCallback g_syndrome_capture_callback = nullptr;
...
+namespace {
+
+void maybe_init_realtime_session() {
+  if (!realtime_mode_inproc_rpc_requested()) {
+    CUDAQ_INFO("CUDAQ_QEC_REALTIME_MODE not set to inproc_rpc; using "
+               "legacy direct-call decoding path.");
+    return;
+  }
+
+  if (!any_decoder_supports_graph_dispatch())
```
melody-ren: I think this will make HOST mode unreachable. This will throw for CPU-orchestrated decoders (trt+pymatching composite decoder or vanilla pymatching). If I understand it correctly, HOST mode is now supported?

libs/qec/lib/realtime/realtime_decoding.cpp
```diff
@@ -126,11 +253,14 @@ int configure_decoders(
     CUDAQ_WARN("Error initializing decoders: {}", e.what());
     return 4;
   }
+
+  maybe_init_realtime_session();
```
vedika-saravanan, line 291: This initializes the realtime session after new_decoder->decode(syndrome), which can already create the CUDA context for GPU decoders. That makes the later `cudaSetDeviceFlags(cudaDeviceMapHost)` in `qec_realtime_session::initialize()` too late. Can we move mapped-host flag setup before the decoder dry run when `CUDAQ_QEC_REALTIME_MODE=inproc_rpc` is requested?

libs/qec/lib/realtime/qec_realtime_session.cpp
```diff
@@ -148,166 +207,797 @@ void get_corrections_host(void *slot_host, std::size_t slot_size) {
...
   std::memset(&host_ctx_, 0, sizeof(host_ctx_));
   host_ctx_.ringbuffer = ringbuffer_;
   host_ctx_.config.num_slots = static_cast<std::uint32_t>(num_slots_);
   host_ctx_.config.slot_size = static_cast<std::uint32_t>(slot_size_);
-  host_ctx_.config.dispatch_path = CUDAQ_DISPATCH_PATH_HOST;
-  host_ctx_.config.dispatch_mode = CUDAQ_DISPATCH_HOST_CALL;
+  host_ctx_.config.shared_ring_mode = 1;
   host_ctx_.config.skip_tx_markers = 1;
-  host_ctx_.function_table.entries = function_table_.data();
+  host_ctx_.function_table.entries = function_table_dev_;
```
vedika-saravanan: Should this be `function_table_host_`? This context is consumed by the CPU host dispatcher, while `function_table_dev_` is for the device dispatcher. If mapped pinned memory has distinct host/device addresses, the host loop may dereference a device pointer.

libs/qec/lib/realtime/rpc_producer.cpp
```diff
@@ -12,77 +12,83 @@
...
-      if (__atomic_compare_exchange_n(&tx[slot], &expected, kSlotBusyMarker,
-                                      false, __ATOMIC_ACQ_REL,
-                                      __ATOMIC_RELAXED)) {
-        g_next_slot_hint.store(
-            static_cast<std::uint32_t>((slot + 1) % session.num_slots()),
-            std::memory_order_relaxed);
-        return slot;
-      }
+    for (std::uint32_t s = 0; s < n; ++s) {
+      if (rx[s] == 0 && tx[s] == 0)
```
vedika-saravanan, line 95: I see the single-producer assumption is documented in rpc_producer.cpp / .h, but it is not enforced. Can we confirm this is a hard runtime/API contract? If concurrent producers can call this path later, this would need CAS claiming or producer-side serialization.

libs/qec/lib/realtime/gpu_kernels.cu
```diff
@@ -52,7 +52,7 @@ __device__ void postprocess_observable(
     }
   }
 
-  corrections[observable_idx] = result;
+  corrections[observable_idx] ^= result;
```
vedika-saravanan, line 55: The XOR accumulation looks intentional. Can you point me to the coverage that verifies correction accumulation/reset across multiple decode windows or shots? I see reset-related coverage, but wanted to confirm this path is covered end-to-end.

libs/qec/lib/realtime/realtime_decoding.cpp
cketcham2333: Resolved in commit b919b9dda428fe51c66e4395cac84abaf74a2a0f

libs/qec/lib/realtime/realtime_decoding.cpp
cketcham2333: Resolved in commit b919b9dda428fe51c66e4395cac84abaf74a2a0f

libs/qec/lib/realtime/qec_realtime_session.cpp
cketcham2333: Resolved in commit b919b9dda428fe51c66e4395cac84abaf74a2a0f

libs/qec/lib/realtime/rpc_producer.cpp
cketcham2333: Confirmed — this is a hard single-producer contract, not just a doc note. The only caller is the single-threaded QEC decode loop (realtime_decoding.cpp's enqueue_syndromes / get_corrections / reset_decoder), with one RPC in flight at a time. The latest commit (b919b9dda428fe51c66e4395cac84abaf74a2a0f) (also enforces it rather than leaving it implicit: each producer entry point now holds a `single_producer_guard` that CAS-flips an atomic on entry and throws if a second producer is already active, so a future violation fails loudly instead of racing on slot selection. It's always-on (a real throw, not assert(), which release builds compile out), so it's active in production too.

libs/qec/lib/realtime/gpu_kernels.cu
cketcham2333: The `^=` is intentional: `postprocess_observable` XOR-accumulates each device decode's observable flips into the running correction buffer (so an even number of flips over the accumulation cancels), and the reset path clears it. End-to-end coverage is in `test_realtime_qldpc_graph_decoding` (libs/qec/unittests/realtime/qec_graph_decode_test/), which drives the full shared-ring path — host-loop `enqueue_syndromes` + device-loop `get_corrections`/`reset_decoder` — against a recorded relay-BP fixture:

vedika-saravanan:
Thanks @cketcham2333 for addressing the review comments. I’m approving the PR since all of my comments have been addressed.

vedika-saravanan (approved):

PR #615, opened 2026-06-15
Dem from stim text can use error decompositions

eliotheinrich:
## Description
In the current implementation of `dem_from_stim_text`, the suggested error decomposition `^` characters are ignored, but it is sometimes useful to output graphlike edge check matrices. For example,

```
import numpy as np
import stim
import cudaq_qec as qec
import beliefmatching

dem_text = "error(0.1) D0 ^ D1\n"

check_matrices = beliefmatching.detector_error_model_to_check_matrices(stim.DetectorErrorModel(dem_text))

# Current behavior: D0 and D1 merged into one column
dem_old = qec.dem_from_stim_text(dem_text)
print("dem_from_stim_text:")
print(np.array(dem_old.detector_error_matrix))          # [[1], [1]]

# Sometimes desirable: accept the error decomposition indicated by ^ and split into two columns
print("beliefmatching's edge_check_matrix:")
print(check_matrices.edge_check_matrix.toarray())                   # [[1, 0], [0, 1]
```

This PR adds a flag to `dem_from_stim_text` which allows users to toggle if the `^` characters should be ignored (default) or should be used to generate an error-decomposed DEM. This is useful if the DEM will eventually be used for, i.e., pymatching. 

This change does not perform error decomposition; it only modifies the parsing behavior or `dem_from_stim_text` to optionally accept suggested error decompositions from `stim`.

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/libs/qec/include/cudaq/qec/detector_error_model.h
+++ b/libs/qec/include/cudaq/qec/detector_error_model.h
@@ -23,8 +23,10 @@ namespace cudaq::qec {
 ///
 /// Shared size parameters among the matrix types.
 /// - `detector_error_matrix`: num_detectors x num_error_mechanisms [d, e]
-/// - `error_rates`: num_error_mechanisms
 /// - `observables_flips_matrix`: num_observables x num_error_mechanisms [k, e]
+/// - `error_rates`: num_physical_error_mechanisms (≤ num_error_mechanisms)
+/// - `error_ids` (optional): num_error_mechanisms — maps column i to its
+///   corresponding physical error mechanismin `error_rates`
 ///
 /// @note The C++ API for this class may change in the future. The Python API is
 /// more likely to be backwards compatible.
@@ -35,9 +37,10 @@ struct detector_error_model {
   /// i is triggered by error mechanism j, and 0 otherwise.
   cudaqx::tensor<uint8_t> detector_error_matrix;
 
-  /// The list of weights has length equal to the number of columns of
-  /// `detector_error_matrix`, which assigns a likelihood to each error
-  /// mechanism.
+  /// One entry per distinct physical error mechanism. When `error_ids` is not
+  /// set, this equals one entry per column of `detector_error_matrix`. When
+  /// `error_ids` is set, this equals one entry per distinct physical error
+  /// mechanism. Use `error_rates[error_ids[i]]` for the rate of column i.
   std::vector<double> error_rates;
 
   /// The observables flips matrix is a specific kind of circuit-level parity-
@@ -82,13 +85,16 @@ struct detector_error_model {
                                bool remove_zero_syndrome_errors = false);
 };
 
-/// Parse Stim DEM text into detector/observable flip matrices and error rates.
-/// DEM-native decoders should consume raw DEM text instead.
-/// If @p decompose_errors is true, error mechanisms that carry an explicit
-/// graphlike decomposition (components separated by '^' in the DEM text) are
-/// expanded into one column per component; otherwise the '^' separators are
-/// ignored and each error instruction produces a single column.
+/// Parse the Stim DEM string @p dem_text into detector/observable flip
+/// matrices and error rates. DEM-native decoders should consume raw DEM text
+/// instead. By default (@p use_decomp_suggestions = false) the '^' separators
+/// are ignored and each error instruction produces a single column. If
+/// @p use_decomp_suggestions is true, error mechanisms that carry an explicit
+/// graphlike decomposition (components separated by '^') are expanded into one
+/// column per component; @p error_ids is then populated so that columns sharing
+/// an ID originate from the same instruction and carry correlated (not
+/// independent) error rates.
 detector_error_model dem_from_stim_text(const std::string &dem_text,
-                                        bool decompose_errors = false);
+                                        bool use_decomp_suggestions = false);
 
 } // namespace cudaq::qec
--- a/libs/qec/include/cudaq/qec/detector_error_model.h
+++ b/libs/qec/include/cudaq/qec/detector_error_model.h
@@ -84,6 +84,11 @@ struct detector_error_model {
 
 /// Parse Stim DEM text into detector/observable flip matrices and error rates.
 /// DEM-native decoders should consume raw DEM text instead.
-detector_error_model dem_from_stim_text(const std::string &dem_text);
+/// If @p decompose_errors is true, error mechanisms that carry an explicit
+/// graphlike decomposition (components separated by '^' in the DEM text) are
+/// expanded into one column per component; otherwise the '^' separators are
+/// ignored and each error instruction produces a single column.
+detector_error_model dem_from_stim_text(const std::string &dem_text,
+                                        bool decompose_errors = false);
 
 } // namespace cudaq::qec
--- a/libs/qec/include/cudaq/qec/detector_error_model.h
+++ b/libs/qec/include/cudaq/qec/detector_error_model.h
@@ -23,10 +23,8 @@ namespace cudaq::qec {
 ///
 /// Shared size parameters among the matrix types.
 /// - `detector_error_matrix`: num_detectors x num_error_mechanisms [d, e]
+/// - `error_rates`: num_error_mechanisms
 /// - `observables_flips_matrix`: num_observables x num_error_mechanisms [k, e]
-/// - `error_rates`: num_physical_error_mechanisms (≤ num_error_mechanisms)
-/// - `error_ids` (optional): num_error_mechanisms — maps column i to its
-///   corresponding physical error mechanismin `error_rates`
 ///
 /// @note The C++ API for this class may change in the future. The Python API is
 /// more likely to be backwards compatible.
@@ -37,10 +35,9 @@ struct detector_error_model {
   /// i is triggered by error mechanism j, and 0 otherwise.
   cudaqx::tensor<uint8_t> detector_error_matrix;
 
-  /// One entry per distinct physical error mechanism. When `error_ids` is not
-  /// set, this equals one entry per column of `detector_error_matrix`. When
-  /// `error_ids` is set, this equals one entry per distinct physical error
-  /// mechanism. Use `error_rates[error_ids[i]]` for the rate of column i.
+  /// The list of weights has length equal to the number of columns of
+  /// `detector_error_matrix`, which assigns a likelihood to each error
+  /// mechanism.
   std::vector<double> error_rates;
 
   /// The observables flips matrix is a specific kind of circuit-level parity-
@@ -91,9 +88,9 @@ struct detector_error_model {
 /// are ignored and each error instruction produces a single column. If
 /// @p use_decomp_suggestions is true, error mechanisms that carry an explicit
 /// graphlike decomposition (components separated by '^') are expanded into one
-/// column per component; @p error_ids is then populated so that columns sharing
-/// an ID originate from the same instruction and carry correlated (not
-/// independent) error rates.
+/// column per component, each inheriting the probability of the parent
+/// instruction. @p error_ids is always left as nullopt. Note that this is a
+/// lossy approximation of the original DEM.
 detector_error_model dem_from_stim_text(const std::string &dem_text,
                                         bool use_decomp_suggestions = false);
 
--- a/libs/qec/lib/detector_error_model.cpp
+++ b/libs/qec/lib/detector_error_model.cpp
@@ -21,7 +21,8 @@
 
 namespace cudaq::qec {
 
-detector_error_model dem_from_stim_text(const std::string &dem_text) {
+detector_error_model dem_from_stim_text(const std::string &dem_text,
+                                        bool decompose_errors) {
   auto dem = [&dem_text]() {
     try {
       return stim::DetectorErrorModel(dem_text);
@@ -51,11 +52,10 @@ detector_error_model dem_from_stim_text(const std::string &dem_text) {
                                std::to_string(prob) +
                                " out of range [0, 1] at instruction index " +
                                std::to_string(instruction_index));
+
     std::vector<std::size_t> dets;
     std::vector<std::size_t> obs;
-    for (const auto &target : inst.target_data) {
-      if (target.is_separator())
-        continue;
+    auto push_target = [&](const stim::DemTarget &target) {
       if (target.is_relative_detector_id()) {
         dets.push_back(static_cast<std::size_t>(target.val()));
       } else if (target.is_observable_id()) {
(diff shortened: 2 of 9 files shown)


libs/qec/lib/detector_error_model.cpp
```diff
@@ -66,10 +66,38 @@ detector_error_model dem_from_stim_text(const std::string &dem_text) {
...
       }
+    };
+
+    if (decompose_errors) {
+      // Each segment delimited by '^' in the DEM text becomes its own column.
+      auto flush = [&]() {
+        if (!dets.empty() || !obs.empty()) {
+          detector_hits.push_back(dets);
+          observable_hits.push_back(obs);
+          rates.push_back(prob);
```
vedika-saravanan: This changes the probability semantics for decomposed Stim errors. In Stim, `error(p) A ^ B` is still one correlated error mechanism with a suggested decomposition, but this creates multiple CUDA-Q columns each with probability `p`. Existing CUDA-Q consumers treat `error_rates` / `error_rate_vec` as per-column independent probabilities, so this would make the split components independently sampleable/weighted.

Can we avoid exposing decomposed columns as independent error mechanisms, or preserve the correlation explicitly, e.g. with shared `error_ids` plus clear consumer support? Otherwise `decompose_errors=True` should probably not populate ordinary `error_rates` in a way that looks probability-preserving.

libs/qec/lib/detector_error_model.cpp
```diff
@@ -66,10 +66,38 @@ detector_error_model dem_from_stim_text(const std::string &dem_text) {
             ") contains an unsupported target kind; only D* (detector) and "
             "L* (observable) targets are supported by the fallback parser");
       }
+    };
+
+    if (decompose_errors) {
+      // Each segment delimited by '^' in the DEM text becomes its own column.
+      auto flush = [&]() {
+        if (!dets.empty() || !obs.empty()) {
+          detector_hits.push_back(dets);
```
vedika-saravanan: This stores the raw component before duplicate detector targets are XOR-cancelled. For example, `D0 D0 ^ D1` later becomes an all-zero detector column with a nonzero rate. That is not graphlike and will break consumers such as the PyMatching plugin, which requires each column to have one or two detector endpoints.

Can we canonicalize/XOR-cancel each component before appending, then reject decomposed components with 0 or more than 2 detector endpoints instead of emitting them?

libs/qec/include/cudaq/qec/detector_error_model.h
```diff
@@ -71,6 +71,11 @@ struct detector_error_model {
 
 /// Parse Stim DEM text into detector/observable flip matrices and error rates.
 /// DEM-native decoders should consume raw DEM text instead.
-detector_error_model dem_from_stim_text(const std::string &dem_text);
+/// If @p decompose_errors is true, error mechanisms that carry an explicit
+/// graphlike decomposition (components separated by '^' in the DEM text) are
+/// expanded into one column per component; otherwise the '^' separators are
+/// ignored and each error instruction produces a single column.
+detector_error_model dem_from_stim_text(const std::string &dem_text,
```
vedika-saravanan, line 94: Changing the only declaration from `dem_from_stim_text(const std::string&)` to `dem_from_stim_text(const std::string&, bool)` preserves source compatibility via the default argument, but removes the old exported C++ symbol. Any already-built downstream code/plugin linked against the one-argument symbol would fail to load.

Can we keep the one-argument overload and add a separate two-argument overload? The one-argument overload can simply delegate to `dem_from_stim_text(dem_text, false)`.

vedika-saravanan:
Thanks @eliotheinrich. I left a few more comments below.

One testing gap that I think should be fixed in this PR: this adds new behavior to the public C++ `dem_from_stim_text` API, but the new `decompose_errors=true` path is only covered through Python tests. Could we add C++ gtest coverage as well?

At minimum, I’d like a small test in `libs/qec/unittests/test_decoders.cpp` covering:
- default / `false` behavior still ignores `^` and produces one column per Stim error instruction
- `true` behavior splits `^` components into separate columns
- edge behavior for duplicate detector targets after XOR cancellation, especially if we validate/reject empty decomposed components

Potential follow-up: docs/example clarity. The examples now show `dem_from_stim_text(..., decompose_errors=True)`, but the decoder is still constructed from `dem_text`, which uses the default non-decomposed path. That can make readers think `get_decoder(..., dem_text)` is using the decomposed DEM when it is not.

Could we either restructure the example to actually use the decomposed matrix where relevant, or track a follow-up to explicitly separate this as parser inspection only and state that direct `get_decoder(..., dem_text)` uses the default non-decomposed parsing?

libs/qec/include/cudaq/qec/detector_error_model.h
melody-ren: I agree that this is somewhat of a breaking change. However since we don't have a hard downstream dependency (we typically expect downstream to rebuild against the latest main/release), I would prefer that we don't add this overload just to preserve an old mangled symbol.

vedika-saravanan: That makes sense. If we expect downstream consumers to rebuild against the latest main/release, I’m fine not preserving the old one-arg mangled symbol here. Thanks for clarifying.

libs/qec/lib/detector_error_model.cpp
eliotheinrich: I agree that it probably makes the most sense to remove all-zero columns, but it looks like there is an existing test case that expects it as an output: https://github.com/NVIDIA/cudaqx/blob/4bb93c65cbf175e47445a7ac9dc4d7a627334a93/libs/qec/unittests/test_decoders.cpp#L806-L819

Does it make sense for these all-zero columns to be removed only when decompose_errors=true, or to change this test?

libs/qec/include/cudaq/qec/detector_error_model.h
```diff
@@ -71,6 +71,11 @@ struct detector_error_model {
...
+                                        bool decompose_errors = false);
```
melody-ren: I want a refund on my high hopes due this name `decompose_errors` ;). I thought it meant that we're (cudaqx) doing the decomposition. Can we use something like `use_stim_err_decomp` or some other concise name to indicate that this flags toggles whether we'll ignore or use the stim decomposition? Feel free to suggest a better name. I don't quite like what I suggested but can't think of a better one

libs/qec/lib/detector_error_model.cpp
melody-ren: I agree with Vedika's concern that this changes the semantics of the error rate vec. I do think that we need to have both meanings. Since `error_rates` has an established meaning in this codebase (physical error rates), I suggest that we add another way rather than overloading the existing meaning of `error_rates`

melody-ren (commented):
Thanks Eliot! Could you please update the description to clarify that this PR doesn't decompose errors (already done by stim)?

libs/qec/lib/detector_error_model.cpp
bmhowe23: Note: the all-zero's case is one that is specifically called out and addressed in #610.

libs/qec/include/cudaq/qec/detector_error_model.h
bmhowe23: Agree, we do not need to maintain ABI compatibility across release version. It is ok for expecting C++ users to rebuild. (CUDA-Q does this same thing.) I'll mark this one as resolved.

libs/qec/lib/detector_error_model.cpp
bmhowe23: > This changes the probability semantics for decomposed Stim errors. In Stim, error(p) A ^ B is still one correlated error mechanism with a suggested decomposition, but this creates multiple CUDA-Q columns each with probability p.

For ease of use with the rest of the community, we should follow whatever Stim does here. What does Stim do in this case?

libs/qec/lib/detector_error_model.cpp
eliotheinrich: 6303753 now tracks the parity of each detector/observable in each error mechanism and only includes them if the parity is odd. This is consistent with the behavior of `DetectorErrorModelTest.CanonicalizeZeroSyndromeObservable` in #610 when `remove_zero_syndrome_errors=true`.

libs/qec/include/cudaq/qec/detector_error_model.h
eliotheinrich: Sure; I've changed `decompose_errors` to `use_decomp_suggestions` in 6303753. Hopefully this is clearer!

eliotheinrich:
> One testing gap that I think should be fixed in this PR: this adds new behavior to the public C++ `dem_from_stim_text` API, but the new `decompose_errors=true` path is only covered through Python tests. Could we add C++ gtest coverage as well?
> 
> At minimum, I’d like a small test in `libs/qec/unittests/test_decoders.cpp` covering:
> 
> * default / `false` behavior still ignores `^` and produces one column per Stim error instruction
> * `true` behavior splits `^` components into separate columns
> * edge behavior for duplicate detector targets after XOR cancellation, especially if we validate/reject empty decomposed components
> 
> Potential follow-up: docs/example clarity. The examples now show `dem_from_stim_text(..., decompose_errors=True)`, but the decoder is still constructed from `dem_text`, which uses the default non-decomposed path. That can make readers think `get_decoder(..., dem_text)` is using the decomposed DEM when it is not.
> 
> Could we either restructure the example to actually use the decomposed matrix where relevant, or track a follow-up to explicitly separate this as parser inspection only and state that direct `get_decoder(..., dem_text)` uses the default non-decomposed parsing?

Thanks for the review; commit 6303753 addresses these issues. I've added new gtests to address these testing gaps: `DecomposeErrorsFalseIgnoresCaret`, `DecomposeErrorsTrueSplitsCaretComponents`, and `DecomposeErrorsXorCancelled`. For the documentation examples, I've made the fact that the decomposed DEM is not directly utilized by the decoder more explicit.

libs/qec/lib/detector_error_model.cpp
eliotheinrich: The closest analogue we can get to Stim is to use the original semantics, I believe, wherein `error_rates` represent physical error rates and `error_ids` encode correlations. This is what I've done in the most recent commit 6303753.

libs/qec/lib/detector_error_model.cpp
```diff
@@ -50,41 +52,74 @@ detector_error_model dem_from_stim_text(const std::string &dem_text) {
...
+    auto toggle = [](std::set<std::size_t> &s, std::size_t v) {
+      if (!s.erase(v)) {
+        s.insert(v);
+      }
+    };
+
+    auto push_target = [&](const stim::DemTarget &target) {
       if (target.is_relative_detector_id()) {
-        dets.push_back(static_cast<std::size_t>(target.val()));
+        toggle(dets_parity, static_cast<std::size_t>(target.val()));
```
bmhowe23, line 67: Is the purpose of this `toggle` (vice the previous dets.push_back) because we are concerned about detectors being listed twice within a single `error` mechanism in a Stim DEM? I don't think I've ever seen that before; I don't think it's possible.

libs/qec/python/tests/test_decoder.py
```diff
@@ -852,6 +852,105 @@ def test_dem_from_stim_text_explicit_parse_then_get_decoder():
...
+    dem = qec.dem_from_stim_text(dem_text, use_decomp_suggestions=True)
+    assert dem.num_error_mechanisms() == 2
+    np.testing.assert_array_equal(A(dem.detector_error_matrix),
+                                  A([[1, 0], [0, 1]]))
+    np.testing.assert_array_equal(A(dem.observables_flips_matrix),
+                                  A([[1, 0], [0, 1]]))
+
+    # 3. Repeated detector within one component XOR-cancels to 0.
+    # error(0.1) D0 D0 ^ D1  →  component 0 has no net detectors after
+    # cancellation so it is dropped entirely; only D1 remains as one column.
```
bmhowe23, line 988: Can you find any examples where Stim actually did this?

libs/qec/unittests/test_decoders.cpp
```diff
@@ -804,18 +804,20 @@ TEST(StimDemGetDecoder, StillAcceptsParityCheckMatrix) {
 }
 
 TEST(StimDemGetDecoder, RepeatedDetectorOrObservableTargetsXorFold) {
-  const std::string dem_text = R"(error(0.1) D0 D0
-error(0.1) L0 L0
+  const std::string dem_text = R"(error(0.1) D0 D0 D1
+error(0.1) L0 L0 D2
```
bmhowe23, line 928: I don't think repeated detectors or observables are ever produced by Stim. Let me know if you know otherwise.

libs/qec/lib/detector_error_model.cpp
eliotheinrich: I don't believe Stim ever emits duplicate detectors like in the examples below, but it does parse them in a way that respects the parity. Dem strings can in principle be produced from lots of sources, so it could make sense to have some policy for these cases even if stim doesn't output them. And if we are parsing the output of stim, this code won't add much extra complexity at runtime.

Open to suggestions if there's a good reason to simplify, but imo it makes sense to leave it, even if Stim will protect us in most cases.

bmhowe23: OK, thanks for the clarification. No changes needed here.

eliotheinrich:
The latest commit [5bb8117](https://github.com/NVIDIA/cudaqx/pull/615/commits/5bb811731a5e3a96ab4849d7ad64e9d6e2e8bc49) represents the decomposed DEMs as an approximation of the original DEM where the decomposed mechanisms are independently distributed with the same probability as the original mechanism. That is,

`error(0.1) D0 D1 ^ D2 D3`
becomes
`error(0.1) D0 D1`
`error(0.1) D2 D3`

This seems to be consistent with the way, i.e., pymatching interprets decomposition hints from stim:

```
import stim
import pymatching

dem_text = "error(0.1) D0 ^ D1\n"
dem = stim.DetectorErrorModel(dem_text)
m = pymatching.Matching(dem)

print("Edges pymatching derives from the stim DEM:")
for u, v, attr in m.edges():
    node_u = f"D{u}"
    node_v = f"D{v}" if v is not None else "boundary"
    print(f"  {node_u} -- {node_v}  p={attr['error_probability']}")
    
    # Edges pymatching derives from the stim DEM:
    # D0 -- D1  p=0.1
    # D2 -- D3  p=0.1
```

vedika-saravanan:
thanks Eliot. The changes LGTM to me :).

melody-ren (approved):

PR #619, opened 2026-06-18
Enable the composite decoder in surface-code example. Depends on #536

melody-ren:
## Summary

This wires the TRT + PyMatching composite decoder into the merged HOST_CALL realtime path and adds coverage for both the direct composite RPC path and the surface-code example.

- Add a `TrtDecoderHostCallRpc` unit test that builds the composite decoder through the standard decoder config path, routes it through the in-process HOST_CALL session, and validates observable corrections.
- Extend `surface_code-1` to support `trt_decoder` configs with a TensorRT predecoder and nested PyMatching global decoder.
- Use the Z-sector DEM for the TRT/PyMatching surface-code path because the full X+Z DEM generated by this example is not graphlike for PyMatching.
- Add a small ONNX generator for split predecoder outputs `[pre_L | residual]`.
- Add TRT surface-code test variants, including `inproc_rpc`, and make CI install `onnx`/`pyyaml` before configure so the tests are actually registered.
- Preserve readable syndrome dumps with `ROUND_START` markers while truncating packed syndrome bits to the real per-round syndrome width.

## Merge sequence
Depends on #536 . Should go in after 536. 

## Testing

Ran in `cudaqx-public-pr615-cu13-dev`:

```bash
cmake --build /tmp/cudaqx-public-pr615-pr536-trt-build-mainrtlibs --target surface_code-1-local
ctest --output-on-failure -R "TrtDecoderHostCallRpc|app_examples.surface_code-1-local-test-distance-3-trt"
ctest --output-on-failure -R "app_examples.surface_code-1-local-test-distance-3($|-trt)"


melody-ren:
Closing since the trt+pymatching coverage is covered in #647

PR #623, opened 2026-06-23
fix: To fix sliding decoders

kaiqiy-nv:
This PR tries to fix [[B] 6360245](https://nvbugspro.nvidia.com/bug/6360245).

It also adds a test to expose this bug.

### DIFF
--- a/libs/qec/unittests/test_decoders.cpp
+++ b/libs/qec/unittests/test_decoders.cpp
@@ -7,6 +7,7 @@
  ******************************************************************************/
 
 #include "stim.h"
+#include "../lib/decoders/sliding_window.h"
 #include "cudaq/qec/decoder.h"
 #include "cudaq/qec/detector_error_model.h"
 #include "cudaq/qec/pcm_utils.h"
@@ -37,6 +38,38 @@ private:
   std::string name;
   std::optional<std::string> oldValue;
 };
+
+class recording_sliding_window_decoder : public cudaq::qec::sliding_window {
+public:
+  using cudaq::qec::sliding_window::sliding_window;
+
+  std::vector<std::vector<cudaq::qec::float_t>> detector_history;
+
+  cudaq::qec::decoder_result
+  decode(const std::vector<cudaq::qec::float_t> &syndrome) override {
+    detector_history.push_back(syndrome);
+    return {true,
+            std::vector<cudaq::qec::float_t>(this->get_block_size(), 0.0)};
+  }
+};
+
+std::unique_ptr<recording_sliding_window_decoder>
+make_recording_sliding_window_decoder(std::size_t num_syndromes_per_round) {
+  cudaqx::tensor<uint8_t> H({num_syndromes_per_round, 1});
+  H.at({0, 0}) = 1;
+
+  cudaqx::heterogeneous_map params;
+  params.insert("window_size", std::size_t{1});
+  params.insert("step_size", std::size_t{1});
+  params.insert("num_syndromes_per_round", num_syndromes_per_round);
+  params.insert("error_rate_vec", std::vector<cudaq::qec::float_t>{0.1});
+  params.insert("inner_decoder_name", std::string{"sample_decoder"});
+  params.insert("inner_decoder_params", cudaqx::heterogeneous_map{});
+
+  return std::make_unique<recording_sliding_window_decoder>(
+      cudaq::qec::sparse_binary_matrix(H), params);
+}
+
 } // namespace
 
 TEST(DecoderUtils, CovertHardToSoft) {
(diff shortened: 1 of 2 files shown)


libs/qec/unittests/test_decoders.cpp
```diff
@@ -159,6 +178,37 @@ TEST(SampleDecoder, checkAPI) {
...
+  ASSERT_EQ(decoder.detector_history.size(), 1);
+  ASSERT_EQ(decoder.detector_history[0].size(), 1);
+  EXPECT_EQ(decoder.detector_history[0][0], 1.0);
+
+  EXPECT_TRUE(decoder.enqueue_syndrome(std::vector<uint8_t>{1}));
+  ASSERT_EQ(decoder.detector_history.size(), 2);
+  ASSERT_EQ(decoder.detector_history[1].size(), 1);
+  EXPECT_EQ(decoder.detector_history[1][0], 0.0);
+}
+
```
cketcham2333, line 304: {{0, 1}} only covers has_first_round_detectors == false, so the if (pimpl->current_round >= 2) guard is always true and its false (skip-copy) side stays untested. Could you add a set_D_sparse({{0}, {0, 1}}) case asserting d1 = s1, d2 = s1^s2, d3 = s2^s3?

Maybe something like this?

TEST(SlidingWindowDecoder, EnqueueHandlesFirstRoundDetector) {
  cudaqx::tensor<uint8_t> H({1, 1});
  H.at({0, 0}) = 1;
  cudaqx::heterogeneous_map params;
  params.insert("window_size", std::size_t{1});
  params.insert("step_size", std::size_t{1});
  params.insert("num_syndromes_per_round", std::size_t{1});
  params.insert("error_rate_vec", std::vector<cudaq::qec::float_t>{0.1});
  params.insert("inner_decoder_name", std::string{"sample_decoder"});
  params.insert("inner_decoder_params", cudaqx::heterogeneous_map{});
  recording_sliding_window_decoder decoder(cudaq::qec::sparse_binary_matrix(H),
                                           params);
  // First detector row has a single index, so has_first_round_detectors == true
  // and the first decode happens on round 1. That exercises the FALSE side of
  // the `if (pimpl->current_round >= 2)` guard (the skipped copy), which the
  // EnqueueKeepsRollingState test (D_sparse {{0, 1}}) never reaches.
  decoder.set_D_sparse(std::vector<std::vector<uint32_t>>{{0}, {0, 1}});
  decoder.set_O_sparse(std::vector<std::vector<uint32_t>>{});
  // Feeding rounds 1, 1, 0 should produce:
  //   round 1 (first-round detector): d0 = s0      = 1
  //   round 2 (timelike):             d1 = s0 ^ s1 = 1 ^ 1 = 0
  //   round 3 (timelike):             d2 = s1 ^ s2 = 1 ^ 0 = 1
  // With a first-round detector every enqueue triggers a decode (including the
  // first). On the buggy reset-to-zero code each round is mis-treated as a fresh
  // first-round detector, yielding the raw syndromes 1, 1, 0 instead.
  EXPECT_TRUE(decoder.enqueue_syndrome(std::vector<uint8_t>{1}));
  ASSERT_EQ(decoder.detector_history.size(), 1);
  ASSERT_EQ(decoder.detector_history[0].size(), 1);
  EXPECT_EQ(decoder.detector_history[0][0], 1.0);
  EXPECT_TRUE(decoder.enqueue_syndrome(std::vector<uint8_t>{1}));
  ASSERT_EQ(decoder.detector_history.size(), 2);
  ASSERT_EQ(decoder.detector_history[1].size(), 1);
  EXPECT_EQ(decoder.detector_history[1][0], 0.0);
  EXPECT_TRUE(decoder.enqueue_syndrome(std::vector<uint8_t>{0}));
  ASSERT_EQ(decoder.detector_history.size(), 3);
  ASSERT_EQ(decoder.detector_history[2].size(), 1);
  EXPECT_EQ(decoder.detector_history[2][0], 1.0);
}

cketcham2333 (changes requested):

libs/qec/unittests/test_decoders.cpp
kaiqiy-nv: Hi @cketcham2333 , we have added the "EnqueueHandlesFirstRoundDetector" test suggested by you.
We have also enhanced the tests a bit by changing the synthetic decoder setup from 1 syndrome bit per round to 2 syndrome bits per round, so it now verifies that the rolling-state fix preserves and shifts an entire syndrome round, not only a scalar value.
Cheers.

cketcham2333 (approved):

bmhowe23:
@kaiqiy-nv it looks like the CI is failing one test. Could you please take a look?

kaiqiy-nv:
Hi @bmhowe23 ,

This case seems to be a little bit complicated...

We thought there was a functional bug because we noticed that the realtime `enqueue_syndrome()` path for the `sliding_window` decoder seemed to use a stationary-window style buffer reset instead of a continuous-stream rolling buffer when converting raw syndrome rounds into timelike detector rounds.

However, the original stationary-window approach keeps the raw syndrome buffer alive until the real `sliding_window::decode()` returns a non-empty result.
Intermediate decode calls return an empty result and return before the reset, so the original code still computes adjacent detector pairs within each finite window: `r0^r1`, `r1^r2`, `r2^r3`, etc..
The reset only happens after the whole window is committed, and therefore the overall detector computation is still adjacent and correct for the existing finite-window contract.

We tried to replace it with a continuous-stream approach, which keeps the latest syndrome round after every enqueue and tries to make the base layer truly rolling/streaming: i.e. each physical syndrome round is fed once, and the decoder internally forms `prev_round ^ current_round`.
This can be a valid design, but it conflicts with the existing `surface_code-1` example because the example currently re-feeds the last syndrome of the previous window as the first syndrome of the next window. That re-feed is needed by the stationary-window approach to give each window its reference snapshot, but it is not needed for continuous streaming.
With our change, the re-fed boundary creates a spurious `r_boundary ^ r_boundary` detector and shifts/drops one real detector in the next window, which explains the CI failure.

So now we have two options maybe:

[1] Keep the existing stationary-window approach: to close the PR as 'Not a Bug' for now.
This is the low-risk way.

[2] Switch completely to the continuous-stream approach.
This may more naturally match the name `sliding_window`: each physical syndrome round is fed exactly once.
This may also be a better long-term design for true low-latency streaming.
However, this would need extra work beyond `decoder.cpp`: update the C++/Python examples to stop boundary re-feeding, update syndrome save/load replay assumptions, update DEM/window documentation, clarify the `enqueue_syndrome()` contract in the API docs, and add end-to-end tests proving no boundary `r^r` detector is generated.

We would like your opinions on this. Thanks!

bmhowe23:
> So now we have two options maybe:
> 
> [1] Keep the existing stationary-window approach: to close the PR as 'Not a Bug' for now. This is the low-risk way.
> 
> [2] Switch completely to the continuous-stream approach. This may more naturally match the name `sliding_window`: each physical syndrome round is fed exactly once. This may also be a better long-term design for true low-latency streaming. However, this would need extra work beyond `decoder.cpp`: update the C++/Python examples to stop boundary re-feeding, update syndrome save/load replay assumptions, update DEM/window documentation, clarify the `enqueue_syndrome()` contract in the API docs, and add end-to-end tests proving no boundary `r^r` detector is generated.
> 
> We would like your opinions on this. Thanks!

I would favor [1] for now. Despite the name "sliding window", it is not really supporting an infinitely streaming sliding window, and this is a known limitation. We are looking to improve this in the future, but it will require updates to how decoders are initialized, so we are planning that for future work (after 0.7). Given your analysis, I would recommend closing this as "not a bug" for now.

PR #628, opened 2026-06-25
Add QEC device_call host_dispatch E2E coverage

vedika-saravanan:
## Description

Adds focused end-to-end coverage for invoking QEC realtime decoding through a lowered CUDA-Q QPU `device_call` over the `host_dispatch` channel.

This includes the QEC CQR host-dispatch refactor cherry-picked from [Ben’s branch](https://github.com/vedika-saravanan/cudaqx/pull/1), plus focused coverage for the PyMatching smoke-test path.

The test exercises:

`QPU device_call -> host_dispatch -> QEC DeviceCallService -> QEC realtime host decoder API -> PyMatching`

This is intentionally narrow. It uses a small graphlike PyMatching configuration as a smoke-test endpoint for the realtime device-call bridge. It does **not** claim general PyMatching support or full `surface_code-1 + PyMatching` support.

## Known CUDA-Q follow-ups

- `host_dispatch` should not require a visible CUDA device. Current repro:
  `CUDA_VISIBLE_DEVICES="" ctest -R PyMatchingDev`, which fails with `NVQIR.cpp:881`.
- Bit-packed bool vector handling: NVIDIA/cuda-quantum#4816.
- Internal update for the `DeviceCallService` refactor: NVIDIA/cuda-quantum#4806.

We are holding off on bumping CUDA-Q in cudaqx until NVIDIA/cuda-quantum#4831 lands, unless we decide to merge this PR at risk and track these separately.

No user-visible runtime or performance impact. The device-call bridge is compiled only when the required CUDA-Q realtime/device-call artifacts are available.

### Before requesting review
      Not run locally; validated through CI because this requires the realtime CUDA-Q CI build.
- [x] Full CI has been run.

### Scope and size

### Tests
- [x] Tests fail if the covered `device_call -> host_dispatch -> PyMatching` path is broken, including crashes in that path.
      N/A; this PR adds positive e2e coverage for the realtime device-call path.
- [x] Truth data added where simple `EXPECT_*` / `assert` checks are insufficient.  
      N/A; the focused PyMatching e2e test checks the expected correction directly.

### Documentation
      N/A; no public API surface added.
- [x] User-visible behavior changes have public docs, or a follow-up is tracked.  
      N/A; test/CI coverage only.

### Code style

### Dependencies


libs/qec/lib/realtime/decoding-server-cqr/decoding_server_cqr.cpp
```diff
@@ -0,0 +1,212 @@
+/*******************************************************************************
+ * Copyright (c) 2026 NVIDIA Corporation & Affiliates.                         *
+ * All rights reserved.                                                        *
+ *                                                                             *
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+
+#include "cudaq_internal/device_call/DeviceCallService.h"
```
bmhowe23: I don't think we should have to #include "cudaq_internal" directories. If the QEC library needs them, they should not "internal" to cudaq-realtime. (I.e. we don't want our library to depend on non-standard, backdoor hijacks into the cudaq-realtime product.) Additionally (and I know this latter part is not part of your QEC PR), mixing PascalCase with snake_case within the same identifier is just confusing. I don't know if it's too late to change this or not in cudaq-realtime, but I thought all the public APIs were supposed to be snake_case.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include <vector>
+
+namespace {
+
+using cudaq_internal::device_call::DeviceCallHostDispatchTable;
+using cudaq_internal::device_call::DeviceCallService;
+using cudaq_internal::device_call::DeviceCallServicePluginInfo;
+
+constexpr std::uint32_t kEnqueueSyndromesUi64FnId =
+    cudaq::realtime::fnv1a_hash("enqueue_syndromes_ui64");
```
bmhowe23: Let me think about this. I would've expected the "_ui64" suffixes to not be needed for this change. I thought those were unique to the Quantinuum approach and not in alignment with the generic Decoding Server over-the-wire format.

libs/qec/lib/realtime/CMakeLists.txt
```diff
@@ -68,6 +71,52 @@ if(CMAKE_CUDA_COMPILER)
     if(CUDAQ_REALTIME_HOST_DISPATCH_LIBRARY)
       message(STATUS "Found cuda-quantum realtime host dispatch library at ${CUDAQ_REALTIME_HOST_DISPATCH_LIBRARY}")
     endif()
+
+    # CUDA-Q lowered `cudaq::device_call` host-dispatch integration needs the
+    # internal DeviceCallService plugin API, which is available only from a
+    # realtime/device-call-enabled CUDA-Q build.
+    find_path(CUDAQ_DEVICE_CALL_INCLUDE_DIR
+      NAMES cudaq_internal/device_call/DeviceCallService.h
```
bmhowe23, line 101: Just noting: here are more "internal" references that we need to work towards eliminating.

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
```diff
@@ -51,3 +51,49 @@ if(CUDAQ_REALTIME_ROOT AND CUDAQ_REALTIME_INCLUDE_DIR)
...
+      -frealtime-lowering
+  )
+
+  target_link_directories(test_pymatching_device_call_realtime
+    PRIVATE ${CUDAQ_INSTALL_DIR}/lib)
+
+  target_link_libraries(test_pymatching_device_call_realtime PRIVATE
+    cudaq-qec
+    cudaq-qec-realtime-decoding
+    cudaq-qec-realtime-decoding-quantinuum
```
bmhowe23: We should not be linking with `cudaq-qec-realtime-decoding-quantinuum` for this example. Perhaps `cudaq-qec-realtime-decoding-simulation`?

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+                   __ATOMIC_RELEASE);
+}
+
+void enqueue_syndromes_ui64_host(const void *rx_slot, void *tx_slot,
+                                 std::size_t slot_size) {
+  try {
+    std::array<std::uint64_t, 4> args{};
+    if (!tx_slot || !read_u64_args(rx_slot, slot_size, args)) {
+      if (tx_slot && rx_slot)
+        write_response(tx_slot, rx_slot, -1);
```
bmhowe23: This magic number should be some sort of `constexpr`.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+      return;
+    }
+
+    const auto decoder_id = static_cast<std::size_t>(args[0]);
+    const auto syndrome_size = args[1];
+    if (syndrome_size > 64) {
+      write_response(tx_slot, rx_slot, -3);
```
bmhowe23: This magic number should be some sort of `constexpr`.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+    std::vector<std::uint8_t> syndrome(syndrome_size);
+    for (std::uint64_t i = 0; i < syndrome_size; ++i)
+      syndrome[i] = (args[2] >> i) & 1;
+
+    cudaq::qec::decoding::host::enqueue_syndromes(decoder_id, syndrome.data(),
+                                                  syndrome.size(), args[3]);
+    write_response(tx_slot, rx_slot, 0);
+  } catch (...) {
+    if (tx_slot && rx_slot)
+      write_response(tx_slot, rx_slot, -2);
```
bmhowe23: This magic number should be some sort of `constexpr`.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+  }
+}
+
+void get_corrections_ui64_host(const void *rx_slot, void *tx_slot,
+                               std::size_t slot_size) {
+  try {
+    std::array<std::uint64_t, 3> args{};
+    if (!tx_slot || !read_u64_args(rx_slot, slot_size, args)) {
+      if (tx_slot && rx_slot)
+        write_response(tx_slot, rx_slot, -1);
```
bmhowe23: This magic number should be some sort of `constexpr`.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+      return;
+    }
+
+    constexpr std::uint32_t result_len = sizeof(std::uint64_t);
+    const auto return_size = args[1];
+    if (return_size > 64) {
+      write_response(tx_slot, rx_slot, -3);
```
bmhowe23: This magic number should be some sort of `constexpr`.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+      return;
+    }
+
+    if (sizeof(cudaq::realtime::RPCResponse) + result_len > slot_size) {
+      write_response(tx_slot, rx_slot, -5);
```
bmhowe23: This magic number should be some sort of `constexpr`.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+    std::uint64_t correction = 0;
+    for (std::uint64_t i = 0; i < return_size; ++i)
+      correction |= static_cast<std::uint64_t>(corrections[i]) << i;
+    auto *result = static_cast<std::uint8_t *>(tx_slot) +
+                   sizeof(cudaq::realtime::RPCResponse);
+    std::memcpy(result, &correction, result_len);
+    write_response(tx_slot, rx_slot, 0, result_len);
+  } catch (...) {
+    if (tx_slot && rx_slot)
+      write_response(tx_slot, rx_slot, -2);
```
bmhowe23: This magic number should be some sort of `constexpr`. There are many more of these. Rather than making comments for each one, I'll let this be my last comment like it, but please address all of them.

bmhowe23 (commented):
Some initial comments...

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
```diff
@@ -51,3 +51,49 @@ if(CUDAQ_REALTIME_ROOT AND CUDAQ_REALTIME_INCLUDE_DIR)
...
+  )
+
+  set_target_properties(test_pymatching_device_call_realtime PROPERTIES
+    BUILD_RPATH
+      "${CMAKE_BINARY_DIR}/lib;${CMAKE_BINARY_DIR}/lib/decoder-plugins;${CUDAQ_INSTALL_DIR}/lib"
+  )
+
+  add_dependencies(test_pymatching_device_call_realtime cudaq-qec-pymatching)
+  add_dependencies(CUDAQXQECUnitTests test_pymatching_device_call_realtime)
+  add_test(
```
melody-ren: Is there any reason why we use `add_test` here instead of `gtest_discover_tests` like the other tests? Generally our unit tests meant for CI get routed and discovered through gtests, and it also means we don't need to maintain a hand rolled `main()` for each test. But if this test doesn't fit into that frame, could you please document that?

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
melody-ren: I would also like to protest on behalf of number `4` - "Why hast thou forsaken me?"

libs/qec/lib/realtime/CMakeLists.txt
```diff
@@ -68,6 +71,52 @@ if(CMAKE_CUDA_COMPILER)
...
+      )
+    endif()
+    if(CUDAQ_DEVICE_CALL_RUNTIME_LIBRARY)
+      message(STATUS "Found cuda-quantum device_call runtime library at ${CUDAQ_DEVICE_CALL_RUNTIME_LIBRARY}")
+      get_filename_component(_CUDAQ_DEVICE_CALL_RUNTIME_LIB_DIR
+        "${CUDAQ_DEVICE_CALL_RUNTIME_LIBRARY}" DIRECTORY)
+    else()
+      message(STATUS "cuda-quantum device_call runtime library not found. "
+                     "QEC lowered device_call host-dispatch service will not be built.")
+    endif()
```
melody-ren: This build logic is more or less the same starting from line 78 and repeated 3 times. I think we can use a macro here

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
```diff
@@ -0,0 +1,212 @@
...
+      return;
+    }
+
+    if (sizeof(cudaq::realtime::RPCResponse) + result_len > slot_size) {
+      write_response(tx_slot, rx_slot, -5);
+      return;
+    }
+
+    const auto decoder_id = static_cast<std::size_t>(args[0]);
+    std::vector<std::uint8_t> corrections(return_size);
```
melody-ren: We should use something like `std::array<std::uint8_t, 64> corrections{}` on the realtime path

libs/qec/lib/realtime/device_call/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    "${_CUDAQ_DEVICE_CALL_RUNTIME_LIB_DIR}")
+endif()
+
+set_target_properties(cudaq-qec-realtime-device-call PROPERTIES
+  LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+  BUILD_RPATH "${_qec_device_call_build_rpath}"
+  INSTALL_RPATH "${_qec_device_call_install_rpath}"
+)
+
+install(TARGETS cudaq-qec-realtime-device-call
```
melody-ren: The PR description says that this PR is meant for CI only, but here we are installing a .so. It looks like the added device-call-service target is needed in production for the realtime path though. In that case, can you please update the description? 

Additionally, since this target is basically needed on the realtime path and built from a single file, I don't think we need a separate `.so` for it. I think this can be combined into the cudaq-qec-realtime-decoding target. The source files can be conditionally added when the needed headers are present. 

That said, I agree that we need a separated .so if we want to make the device_call functionality as an opt-in for the end users

melody-ren (commented):
Thanks Vedika! I have some questions about packaging.

libs/qec/lib/realtime/decoding-server-cqr/decoding_server_cqr.cpp
vedika-saravanan: Agreed. I'm leaving the current include as a temporary bridge for this PR based on our slack conversation. the clean fix needs CUDA-Q to expose the DeviceCallService provider API at a supported non-internal location. Once Thien's CUDA-Q change lands, I'll update this PR to use the new include path.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
vedika-saravanan: I removed the _ui64 naming entirely and redesigned the wire format. The protocol changed from 4 opaque uint64 args to 3 typed args (decoder_id, stdvec<i1> syndromes, tag). The service now registers simulation_enqueue_syndromes, simulation_get_corrections, and the mangled simulation reset callee, and handles the stdvec<i1> request/response payloads directly. https://github.com/NVIDIA/cudaqx/pull/628/commits/676861275166f05294746521cae90106625694b1

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
vedika-saravanan: Fixed in https://github.com/NVIDIA/cudaqx/commit/676861275166f05294746521cae90106625694b1. The test now links cudaq-qec-realtime-decoding-simulation instead of cudaq-qec-realtime-decoding-quantinuum.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
vedika-saravanan: Fixed in https://github.com/NVIDIA/cudaqx/pull/628/commits/676861275166f05294746521cae90106625694b1. Replaced all reviewer-visible literals with named constexprs covering status codes, function entry indices/counts, schema arg indices, scalar sizes, and the realtime vector length cap.

libs/qec/unittests/decoders/pymatching/test_pymatching_device_call_realtime.cpp
```diff
@@ -0,0 +1,136 @@
...
+  constexpr std::uint64_t kKernelDecoderId = 0;
+  constexpr std::uint64_t kKernelBlockSize = 3;
+  constexpr std::uint64_t kKernelSyndromeSize = 3;
+  constexpr std::uint64_t kKernelSyndromeTag = 1;
+  constexpr std::size_t kKernelActiveSyndromeIndex = 1;
+
+  cudaq::qec::decoding::reset_decoder(/*decoder_id=*/kKernelDecoderId);
+
+  std::vector<bool> syndrome(kKernelSyndromeSize);
+  syndrome[kKernelActiveSyndromeIndex] = true;
```
bmhowe23, line 104: Unfortunately, vectors inside `__qpu__` kernels do not follow standard C++ STL semantics. That is - I don't think it guarantees default-value initialization, so I think it is required to add a loop to initialize the vector to false prior to setting the one `syndrome[kKernelActiveSyndromeIndex] = true`.

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
```diff
@@ -51,3 +51,45 @@ if(CUDAQ_REALTIME_ROOT AND CUDAQ_REALTIME_INCLUDE_DIR)
...
+  target_include_directories(test_pymatching_device_call_realtime PRIVATE
+    ${CUDAQ_REALTIME_INCLUDE_DIR}
+    ${CUDAToolkit_INCLUDE_DIRS}
+  )
+
+  cudaqx_add_device_code(test_pymatching_device_call_realtime
+    SOURCES
+      test_pymatching_device_call_realtime.cpp
+    COMPILER_FLAGS
+      --target qpp-cpu
```
bmhowe23:
```suggestion
      --target stim
```

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
```diff
@@ -51,3 +51,45 @@ if(CUDAQ_REALTIME_ROOT AND CUDAQ_REALTIME_INCLUDE_DIR)
...
+  target_link_directories(test_pymatching_device_call_realtime
+    PRIVATE ${CUDAQ_INSTALL_DIR}/lib)
+
+  target_link_libraries(test_pymatching_device_call_realtime PRIVATE
+    GTest::gtest_main
+    cudaq-qec
+    cudaq-qec-realtime-decoding
+    cudaq-qec-realtime-decoding-simulation
+    cudaq::cudaq cudaq::cudaq-common cudaq-mlir-runtime
+    nvqir nvqir-qpp
```
bmhowe23:
```suggestion
    nvqir nvqir-stim
```

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
```diff
@@ -51,3 +51,45 @@ if(CUDAQ_REALTIME_ROOT AND CUDAQ_REALTIME_INCLUDE_DIR)
...
+  set_target_properties(test_pymatching_device_call_realtime PROPERTIES
+    BUILD_RPATH
+      "${CMAKE_BINARY_DIR}/lib;${CMAKE_BINARY_DIR}/lib/decoder-plugins;${CUDAQ_INSTALL_DIR}/lib"
+  )
+
+  add_dependencies(test_pymatching_device_call_realtime cudaq-qec-pymatching)
+  add_dependencies(CUDAQXQECUnitTests test_pymatching_device_call_realtime)
+  gtest_discover_tests(test_pymatching_device_call_realtime
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    PROPERTIES ENVIRONMENT "CUDAQ_DEVICE_CALL_CHANNEL=host_dispatch")
```
bmhowe23, line 108: I am having trouble keeping this test scenario (CUDAQ_DEVICE_CALL_CHANNEL=host_dispatch) vs Chuck's previous test scenario (CUDAQ_QEC_REALTIME_MODE=inproc_rpc) straight in my head. As I understand it right now, the one in this PR is more thorough. It still runs entirely in one process, but it uses 2 different threads, and I suspect his inproc_rpc from #609 just uses 1 thread. Is that right?

bmhowe23 (commented):
Thanks for the quick updates so far, Vedika. I will continue reviewing a bit more tomorrow, but I figured I'd stop and post what I have right now. Part of this is that I need to be downloading, building, and testing to make sure I understand what is being done in this PR and what work remains.

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
vedika-saravanan: Yes, this fits the normal GTest pattern. I had originally used `add_test` only because I needed to attach the `CUDAQ_DEVICE_CALL_CHANNEL=host_dispatch` environment to this one test, but `gtest_discover_tests` supports that too.

Fixed in [afa9284f](https://github.com/NVIDIA/cudaqx/pull/628/commits/afa9284ff0fd919295cf9cbaeca5abe623486716). converted this to `GTest::gtest_main` + `gtest_discover_tests`, with the host-dispatch environment preserved on the discovered test.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
vedika-saravanan: Fair protest. That `4` came from the old `_ui64` helper shape: decoder id, syndrome size, packed syndrome bits, and tag. That made sense only for the first quick smoke test, but it was too opaque and also no longer matched the simulation ABI.

Fixed in 67686127. removed the `_ui64` path and switched to the typed simulation wrapper ABI, so the old 4-argument shape is gone. Remaining indices/counts/status values are named constants.

libs/qec/lib/realtime/CMakeLists.txt
vedika-saravanan: Agreed. The repeated blocks were from looking for the same realtime/device-call artifacts in multiple possible CUDA-Q install layouts, but the logic itself does not need to be duplicated.

Fixed in afa9284f. I added helper macros for the realtime/device-call `find_path` and `find_library` calls and reused them across the realtime headers/libs, host-dispatch library, device-call headers/runtime, and realtime pipeline lookup.

libs/qec/lib/realtime/device_call/qec_device_call_service.cpp
vedika-saravanan: The fixed-size raw buffer was just meant to keep the realtime handler bounded, but `std::array` makes that cap explicit and safer.

Fixed in afa9284f. the service now uses bounded `std::array` buffers for syndrome and correction payloads, with explicit size checks before copying. Oversized requests are rejected before touching the buffers.

libs/qec/lib/realtime/device_call/CMakeLists.txt
vedika-saravanan: Good point. I don’t think this needs to be a separately installed `.so` for this PR. The service is not a user-facing opt-in component here; it is part of the QEC realtime implementation when CUDA-Q’s device-call provider API/runtime are available.

Fixed in afa9284f. I folded `qec_device_call_service.cpp` into `cudaq-qec-realtime-decoding` under the device-call availability gate, removed the separate installed shared library, and kept `cudaq-qec-realtime-device-call` only as an interface marker so the test can be gated cleanly.

With that change, the PR remains CI/test coverage focused and does not add a new installed device-call-service library.

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
vedika-saravanan: Fixed in [0a5f2e0](https://github.com/NVIDIA/cudaqx/pull/628/commits/0a5f2e0d9059266d7d9cf1bfe7555d2bd62bcd25)

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
vedika-saravanan: Fixed in [0a5f2e0](https://github.com/NVIDIA/cudaqx/pull/628/commits/0a5f2e0d9059266d7d9cf1bfe7555d2bd62bcd25)

libs/qec/unittests/decoders/pymatching/CMakeLists.txt
vedika-saravanan: I checked the source to make sure I’m describing this correctly.

You’re right that this PR’s `host_dispatch` path runs in a single process and uses two host-side threads. The main/test thread runs `cudaq::run(...)`, and CUDA-Q’s host-dispatch thread runs `cudaq_host_dispatcher_loop(...)`.

The nuance is that QEC’s `CUDAQ_QEC_REALTIME_MODE=inproc_rpc` HOST mode also starts a QEC host-loop thread through `qec_realtime_session::start_host_loop`, so I don’t think the main distinction from #609 is just thread count.

The main distinction is the entry point being exercised. #609 exercises QEC’s in-process RPC path from host code through `rpc_producer`. This PR exercises calls that originate inside a `__qpu__` kernel, get lowered by CUDA-Q to `device_call`, go through CUDA-Q `host_dispatch`, and then enter QEC through the `DeviceCallService` bridge.

So the new coverage here is specifically the compiler-lowered `device_call -> host_dispatch -> QEC handler` path, which #609 does not exercise.

vedika-saravanan:
> Thanks for the quick updates so far, Vedika. I will continue reviewing a bit more tomorrow, but I figured I'd stop and post what I have right now. Part of this is that I need to be downloading, building, and testing to make sure I understand what is being done in this PR and what work remains.

Thanks Ben, that sounds good.

My current understanding is that this PR is scoped to the single-process lowered-kernel path: `__qpu__ device_call -> CUDA-Q host_dispatch -> QEC DeviceCallService -> QEC host decoder API`, with PyMatching as a small graphlike smoke-test endpoint.

The broader remaining work is the decoder-server path beyond this single-process host-dispatch setup. Full surface-code plus configurable decoder coverage should come through Melody’s new app work.

bmhowe23:
@vedika-saravanan - I posted https://github.com/vedika-saravanan/cudaqx/pull/1 with some suggested changes to this PR. Let me know what you think.

bmhowe23 (approved):
Approved with the assumption that these will be follow-up PRs (copied from the current PR description):

* host_dispatch should not require a visible CUDA device. Current repro: CUDA_VISIBLE_DEVICES="" ctest -R PyMatchingDev, which fails with NVQIR.cpp:881.
* Bit-packed bool vector handling: https://github.com/NVIDIA/cuda-quantum/pull/4816.
* Internal update for the https://github.com/NVIDIA/cuda-quantum/pull/4806.

PR #630, opened 2026-06-25
Create CUDA QEC Logger

tlshannon:
## Description

This PR creates new logger functionality for CUDA-Q QEC, based on the desire to decouple the decoders from CUDA-Q.
- Migrates QEC logging to a dedicated `cudaq::qec` namespace to avoid collisions with the runtime logger.
- Introduces QEC-specific macros `CUDA_QEC_*` and migrates QEC callsites to them; keeps `CUDAQ_*` aliases as compatibility shims with deprecation notes.
- Allow creation of an async forwarder (shared worker thread, default `setForwarder()`), where a user can specify a callback function.  The default is to not create the forwarder.  If a zero-arg setForwarder() is called, then the default callback outputs to stdout/stderr.

Example output: `[2026-06-25 17:24:33.618681] [info] [test_logger.cpp:247] no-forwarder`

Note: This PR is ~1400 lines, due to new implementation of the logger forwarder thread and new tests.

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size
- [x] PR is under ~1000 lines, or an exception is justified in the description (noted in description).

### Tests
- [x] New functionality has new tests (Logger tests).
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


docs/sphinx/api/qec/cpp_api.rst
```diff
@@ -94,6 +94,41 @@ Parity Check Matrix Utilities
...
 .. doxygenfunction:: cudaq::qec::sort_pcm_columns(const cudaqx::tensor<uint8_t> &, std::uint32_t);
 
+Logger
+=============
+
+The QEC logger API currently lives in ``cudaq::qec::detail`` and is used by
+the ``CUDA_QEC_*`` macros exposed in ``cudaq/qec/logger.h``.
+
+.. doxygenenum:: cudaq::qec::detail::LogLevel
+.. doxygenenum:: cudaq::qec::detail::ForwardDropPolicy
```
bmhowe23: Could you please update these to use a consistent naming convention, likely snake_case for all public APIs?

libs/qec/include/cudaq/qec/logger.h
```diff
@@ -0,0 +1,304 @@
...
+  std::thread::id threadId;
+};
+
+/// @brief Queue backpressure policy for forwarded records.
+enum class ForwardDropPolicy { dropNewest, dropOldest };
+
+/// @brief Runtime configuration for asynchronous forwarding.
+struct ForwarderConfig {
+  /// @brief Callback executed on the forwarder worker thread.
+  std::function<void(const ForwardedLogRecord &)> callback;
```
bmhowe23: Would it make sense to change it to this so that one could `move` it (like when being `push_back`-ed onto a queue?
```suggestion
  std::function<void(ForwardedLogRecord &&)> callback;
```

libs/qec/lib/logger.cpp
bmhowe23, line 1: General question ... is the new format compatible enough with the old one such that no changes would be needed for `libs/qec/utils/replay_decoder_logs.py`?

bmhowe23 (commented):
Thanks, Tracy. Overall, I think it looks good. Just a few comments/questions/nits below.

libs/qec/include/cudaq/qec/logger.h
tlshannon: Yes, updated in [9b473af](https://github.com/NVIDIA/cudaqx/pull/630/commits/9b473af9675cba2b7ade3b790f3c44f7cdb78c98)

docs/sphinx/api/qec/cpp_api.rst
tlshannon: I see that now. Updated in [4f2616b](https://github.com/NVIDIA/cudaqx/pull/630/commits/4f2616b0823d64f7b92d7a8dab62e0131df56252)

libs/qec/lib/logger.cpp
tlshannon: Yes, the logger updates are compatible with replay_decoder_logs.py as long as messages aren't truncated (only possible when using a forwarder thread).  I updated the replay script to flag if this happens in [4df30ac](https://github.com/NVIDIA/cudaqx/pull/630/commits/4df30ac393bbf9ca209ccf308ff018b78e958787)

libs/qec/include/cudaq/qec/logger.h
```diff
@@ -0,0 +1,317 @@
...
+#else
+// Remove cudaq::debug log messages from Release binaries.
+template <typename... Args>
+void debug(const std::string_view, Args &&...) {}
+#endif
+
+// Internal helper macro; undefine to avoid leaking it (and colliding with the
+// CUDA-Q runtime logger's identically named macro) into translation units that
+// include both headers.
+#undef CUDA_QEC_LOGGER_DEDUCTION_STRUCT
```
bmhowe23: Is this comment right? I don't think CUDA-Q runtime logger should share any macro names with ours. The one `#undef` this is referring to has "QEC" in it, so I don't think CUDA-Q has that in it.

Ideally, we wouldn't have any macro name overlap with CUDA-Q, and I think (?) it should be possible to avoid that. Let me know if you disagree.

libs/qec/include/cudaq/qec/logger.h
```diff
@@ -0,0 +1,317 @@
...
+#define CUDA_QEC_WARN(...) CUDA_QEC_LOG_IMPL(warn, __VA_ARGS__)
+#define CUDA_QEC_INFO(...) CUDA_QEC_LOG_IMPL(info, __VA_ARGS__)
+
+#ifdef CUDAQ_DEBUG
+#define CUDA_QEC_DBG(...) CUDA_QEC_LOG_IMPL(debug, __VA_ARGS__)
+#else
+#define CUDA_QEC_DBG(...)
+#endif
+
+// Backward-compatible aliases for existing QEC callsites. If runtime logger
```
bmhowe23: I think these backwards-compatible aliases should probably be removed. They might hide cases that we actually want to change, right?

tlshannon: Will do.  I wasn't sure if we should leave them for 3rd party users. [60573ec](https://github.com/NVIDIA/cudaqx/pull/630/commits/60573ec02cb1d2a6a71fb1f2d2dab50ca0a781b5)

libs/qec/include/cudaq/qec/logger.h
tlshannon: You are correct.  It is a leftover from before a rename. Fixed in [60573ec](https://github.com/NVIDIA/cudaqx/pull/630/commits/60573ec02cb1d2a6a71fb1f2d2dab50ca0a781b5)

bmhowe23 (approved):
Thanks, Tracy.

PR #631, opened 2026-06-25
qec/realtime: device-graph scheduler for per-round Hololink QLDPC decoding

cketcham2333:
## Description

Migrates the realtime QEC `DEVICE` dispatch path and the Hololink QLDPC
decoder bridge from the old one-shot-per-syndrome cooperative dispatch to a
**per-round RPC protocol driven by a self-relaunching device-graph scheduler**.

Per-round semantics match how the relay-BP decoder actually consumes data: the
client sends one `enqueue_syndromes` RPC per measurement round (DEVICE_CALL),
then a `get_corrections` RPC per shot; `reset_decoder` clears decoder state.
The decode is a device-launchable cooperative CUDA graph that the `enqueue`
handler triggers fire-and-forget; the scheduler graph tail-relaunches itself so
the device-side launch budget (capped at 120 fire-and-forget launches) is reset
each cycle, allowing an unbounded number of decodes.

Changes in this PR (cudaqx side):

- **Session (`qec_realtime_session`)**: `DEVICE` mode now builds a 3-entry
  DEVICE_CALL function table (enqueue / get_corrections / reset) and launches
  the self-relaunching device-graph scheduler. The `HOST` path is unchanged and
  still serves host-call decoders (e.g. PyMatching).
- **Strict-FIFO ring discipline**: monotonic host producer cursor + disable
  shared-ring (lowest-free-slot) scanning, so the single device consumer walks
  the ring in lockstep with the producer. Pairs with the persistent RX cursor
  in the companion cuda-quantum change.
- **Hololink bridge** (`hololink_qldpc_graph_decoder_bridge.cpp`): rewritten to
  drive the scheduler directly on the DOCA RoCE ring (RX kernel → decode graph →
  TX kernel), reserving SMs for RX/TX, with a guard that clamps `num_pages` to
  the HSB WQE depth (64).
- **Playback** (`hololink_fpga_syndrome_playback.cpp`): adds `--per-round` mode
  (N enqueue frames + 1 get_corrections per shot), per-frame PTP latency CSV,
  and an accurate per-round verification summary; the shot-based default path is
  preserved.
- **Orchestration script**: `--per-round` wiring, real-FPGA bring-up (RoCE v2
  GID wait + permanent FPGA ARP seed), and removal of the `num_pages` knob
  (HSB races above WQE_NUM=64).
- **Docs**: `realtime_relay_bp.rst` updated for the scheduler architecture,
  proprietary components, the per-round flow, and the surface-code relay-BP test.

**Cross-repo dependencies** (not in this PR):
- cuda-quantum realtime: the self-relaunching device-graph scheduler
  (`CUDAQ_DISPATCH_STATUS_TRIGGER_GRAPH`, persistent RX cursor, graph API).
  Temporarily pinned via `.cudaq_version` →
  `cketcham2333/cuda-quantum@8b9d3dc950` (see the do-not-merge note above).
- cuda-qx: the proprietary device-launchable decode graph + `enqueue_syndromes`
  append DEVICE_CALL; the bridge/session link its cudevice archive.

### Before requesting review
- [ ] PR is in Draft if it is not yet ready for review. (Keep as Draft until the
      cuda-quantum pin is resolved.)
- [x] Temporary / debugging changes have been removed. (The per-round
      verification diagnostics are intentional, permanent verifier output.)
- [ ] Full CI has been run.

### Scope and size
- [ ] PR is under ~1000 lines — **exception**: ~1,730 lines; justified above
      (cohesive feature, majority is test tooling + docs).
- [x] Refactoring-only changes are isolated in their own PR(s). (Session refactor
      is required by the new dispatch path, not standalone.)
      (`test_realtime_qldpc_graph_decoding` updated to the strict-FIFO ring mode,
      which is the behavior under test.)

### Tests
- [x] New functionality has new tests. (`GraphDecodeTest.DecodesAllSyndromes`
      exercises the device-graph scheduler; surface-code relay-BP app runs
      >120 decodes without hitting the launch cap.)
      just when it is missing.
- [ ] Negative tests added where exceptions are expected. (N/A.)
- [x] Truth data added where simple checks are insufficient. (Corrections are
      compared against expected observables per shot.)
- [x] CI runtime impact considered; team notified if significant. (`.cudaq_version`
      pin change forces a one-time cuda-quantum rebuild from the fork.)

### Documentation
- [x] User-visible behavior changes have public docs. (`realtime_relay_bp.rst`.)

### Code style

### Dependencies
- [x] No new third-party dependencies. (Depends on internal cuda-quantum /
      cuda-qx realtime components noted above; Hololink/DOCA tooling was already
      gated behind `CUDAQX_QEC_ENABLE_HOLOLINK_TOOLS`. `.cudaq_version` is
      temporarily pinned to a fork — must revert before merge.)

### DIFF
--- a/libs/qec/lib/realtime/rpc_producer.cpp
+++ b/libs/qec/lib/realtime/rpc_producer.cpp
@@ -94,7 +94,8 @@ std::uint32_t acquire_slot(cudaq::qec::realtime::qec_realtime_session &session,
   const std::size_t n = session.num_slots();
   if (rx == nullptr || tx == nullptr || n == 0)
     return UINT32_MAX;
-  const std::uint32_t s = static_cast<std::uint32_t>(session.producer_cursor() % n);
+  const std::uint32_t s =
+      static_cast<std::uint32_t>(session.producer_cursor() % n);
   for (int waited = 0; waited < timeout_ms; ++waited) {
     if (rx[s] == 0 && tx[s] == 0) {
       session.set_producer_cursor((s + 1u) % n);
--- a/libs/qec/lib/realtime/rpc_producer.cpp
+++ b/libs/qec/lib/realtime/rpc_producer.cpp
@@ -142,24 +142,23 @@ void write_and_signal(cudaq::qec::realtime::qec_realtime_session &session,
       session.rx_data_dev() + slot * session.slot_size());
 }
 
-// Bounded spin for `RPCResponse::magic == RPC_MAGIC_RESPONSE`.  Returns
-// false on timeout.  The wire protocol guarantees the consumer writes the
-// header AFTER the payload + flag, so once the magic is visible the rest of
-// the slot is too.
+// Bounded spin for response publication.  Returns false on timeout.  The
+// response is considered complete only after the writer has produced an
+// RPCResponse header and published the matching tx_flags entry.
 bool wait_for_response(cudaq::qec::realtime::qec_realtime_session &session,
                        std::uint32_t slot, int timeout_ms) {
   // Two-ring wire format: the response lives in the TX slot.  The
   // dispatcher's writer (the captured graph for GRAPH_LAUNCH; the
-  // DEVICE_LOOP kernel for DEVICE_CALL) writes RPCResponse and signals
-  // tx_flags[slot]; we spin on the magic word as a structural cue then
-  // confirm via tx_flags before reading the body.
+  // DEVICE_LOOP kernel for DEVICE_CALL) writes RPCResponse and then signals
+  // tx_flags[slot].  Wait for both before reading or releasing the slot.
   std::uint8_t *tx_slot_host =
       session.tx_data_host() + slot * session.slot_size();
   auto *resp =
       reinterpret_cast<const cudaq::realtime::RPCResponse *>(tx_slot_host);
   for (int waited = 0; waited < timeout_ms; ++waited) {
     __sync_synchronize();
-    if (resp->magic == cudaq::realtime::RPC_MAGIC_RESPONSE)
+    if (resp->magic == cudaq::realtime::RPC_MAGIC_RESPONSE &&
+        session.tx_flags_host()[slot] != 0)
       return true;
     // 200us granularity matches the test.  Shorter than acquire_slot's
     // sleep because get_corrections / reset round-trips are sub-ms on
--- a/libs/qec/lib/realtime/rpc_producer.cpp
+++ b/libs/qec/lib/realtime/rpc_producer.cpp
@@ -66,18 +66,26 @@ struct single_producer_guard {
   single_producer_guard &operator=(const single_producer_guard &) = delete;
 };
 
-// Bounded spin for a free slot.  Slot is free when both rx_flags[i] and
-// tx_flags[i] are zero.  Returns UINT32_MAX on timeout.
+// Bounded spin for the NEXT slot in monotonic ring order.  Returns
+// UINT32_MAX on timeout.
+//
+// Ring discipline: the producer walks slots monotonically
+// (session.producer_cursor(), advanced mod num_slots each acquire) rather
+// than picking the lowest free slot.  This keeps it in lockstep with the
+// strict-FIFO consumer (the device-graph scheduler / host loop, both with
+// shared-ring scanning OFF), which waits at exactly its own monotonically-
+// advancing cursor.  The cursor is reset to 0 by
+// qec_realtime_session::initialize(), matching the consumer's reset, so both
+// start at slot 0.  Back-pressure is preserved: we wait until the chosen slot
+// is free (rx_flags[s]==0 AND tx_flags[s]==0 -- request consumed by the
+// dispatcher and response consumed by release_slot()).
 //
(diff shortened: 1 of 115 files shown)


libs/qec/unittests/utils/hololink_qldpc_graph_decoder_bridge.cpp
```diff
@@ -136,63 +228,244 @@ int main(int argc, char *argv[]) {
...
+    decoder->release_decode_graph(raw_res);
+    return 1;
+  }
+  std::cout << "  Scheduler launched (3 DEVICE_CALL entries, triggered decode "
+               "graph)"
+            << std::endl;
+
+  // -- Run the Hololink RX/TX kernels on a worker thread ----------------------
+  std::signal(SIGINT, handle_sigint);
+  std::thread monitor([&]() { hololink_blocking_monitor(transceiver); });
```
vedika-saravanan, line 430: Could there be a startup race here? My understanding is that HSB’s blocking_monitor() launches the RX/TX kernels, so the new thread may not have completed those launches when Bridge Ready is printed. Since playback starts as soon as it sees that message, would it be safer to wait for an explicit readiness signal or retain a short startup grace period before advertising readiness?

libs/qec/unittests/utils/hololink_fpga_syndrome_playback.cpp
```diff
@@ -912,32 +1111,98 @@ int main(int argc, char **argv) {
...
+  // `rounds` enqueue_syndromes frames + 1 get_corrections frame.  rounds is the
+  // number of ROUND_START slices in the syndrome file (1 if unstructured).
+  const std::size_t rounds =
+      options.per_round
+          ? std::max<std::size_t>(1, syndromes.front().per_round.size())
+          : 0;
+  const std::size_t frames_per_shot = options.per_round ? rounds + 1 : 1;
+  // Single logical observable for the surface-code fixture; the verifier checks
+  // observable bit 0 against the expected correction.
+  const std::int64_t num_observables = 1;
```
vedika-saravanan: Should num_observables be derived from O_sparse here, as it is in the bridge? With the value fixed at 1, could a multi-observable configuration request and verify only the first correction while still reporting success? If this tool intentionally supports only one observable, would it be worth validating that explicitly?

libs/qec/unittests/utils/hololink_fpga_syndrome_playback.cpp
```diff
@@ -780,59 +898,140 @@ VerifyResult verify_captured_responses(
...
+  if (per_round && frames_per_shot) {
+    const std::uint32_t total_frames =
+        static_cast<std::uint32_t>(num_expected * frames_per_shot);
+    auto describe = [&](std::uint32_t rid) {
+      const std::uint32_t local = rid % frames_per_shot;
+      const std::uint32_t shot = rid / frames_per_shot;
+      const char *kind =
+          (local + 1 == frames_per_shot) ? "get_corrections" : "enqueue";
+      std::cout << "    rid=" << rid << " (shot=" << shot << " local=" << local
+                << " " << kind << " slot%128=" << (rid % 128) << ")";
```
vedika-saravanan: Since the ring depth is now fixed at 64, should this diagnostic use rid % 64 instead of rid % 128 or use the configured ring depth directly? Would that make the reported alias slot consistent with the new ring configuration?

libs/qec/lib/realtime/rpc_producer.cpp
melody-ren, line 162: I think this is an existing issue predates this PR. Here we return as soon as we see magic. I don't think we actually confirm the state of tx_flags as stated in the comment. On cudaq side, the handler returns before tx_flags is set 
[Link](https://github.com/NVIDIA/cuda-quantum/blob/644b7628245572e154f10ba9d7e3b2dc6c14dd38/realtime/lib/daemon/dispatcher/host_dispatcher.cu#L130). We potentially have a race condition here.

libs/qec/unittests/utils/hololink_qldpc_graph_decoder_bridge.cpp
cketcham2333: Resolved in commit 65d6e273a4b508a1cadfd76366fb05193a9e82c2

libs/qec/unittests/utils/hololink_fpga_syndrome_playback.cpp
cketcham2333: Resolved in commit 65d6e273a4b508a1cadfd76366fb05193a9e82c2

libs/qec/unittests/utils/hololink_fpga_syndrome_playback.cpp
cketcham2333: Resolved in commit 65d6e273a4b508a1cadfd76366fb05193a9e82c2

libs/qec/lib/realtime/rpc_producer.cpp
cketcham2333: Resolved in commit 65d6e273a4b508a1cadfd76366fb05193a9e82c2

melody-ren (approved):

PR #633, opened 2026-06-26
Add YAML and Python config support for Chromobius TRT global decoder

melody-ren:
## Summary

Adds `chromobius_config` support for TRT global decoder configuration across C++ config maps, YAML serialization/deserialization, realtime parameter preparation, and Python bindings.

## Changes

- Adds `chromobius_config` with Chromobius boolean options.
- Extends `global_decoder_config` to support both `pymatching_config` and `chromobius_config`.
- Dispatches TRT `global_decoder_params` by `global_decoder` name for map and YAML paths.
- Defaults missing params for known TRT global decoders while preserving unknown decoder names as extension points when params are omitted.
- Ensures realtime TRT configs attach named global decoders even when `global_decoder_params` is omitted.
- Adds Python binding/export support for `chromobius_config`.
- Keeps Chromobius config scoped to TRT global decoder params rather than top-level decoder YAML custom args.
- Adds C++ and Python coverage for Chromobius params, defaults, unknown decoder behavior, and TRT YAML round trips.

### DIFF
--- a/libs/qec/include/cudaq/qec/realtime/decoding_config.h
+++ b/libs/qec/include/cudaq/qec/realtime/decoding_config.h
@@ -181,7 +181,7 @@ struct decoder_config {
   std::vector<std::int64_t> D_sparse;
   std::variant<single_error_lut_config, multi_error_lut_config,
                nv_qldpc_decoder_config, sliding_window_config,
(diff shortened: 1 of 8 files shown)


bmhowe23 (approved):
LGTM

PR #634, opened 2026-06-26
Add hardware pinning for QEC decoders

kvmto:
## Summary

Allows any registered QEC decoder to be assigned to specific hardware via decoder kwargs and the realtime YAML configuration:

- `cuda_device_id` — pins GPU decoders to a specific device
- `numa_node_id` — binds CPU decoder allocations to a NUMA node

Affinity is owned entirely by the base `decoder` class — no plugin code required. Private RAII guards (`CudaDeviceGuard`, `NumaGuard`) in `decoder.cpp` apply automatically at construction (`decoder::get`), batch decode (`decode_batch`), and async decode (`decode_async`). Raw Linux syscalls, no libnuma. Negative/absent ids are a no-op; out-of-range ids throw at construction.

`device_affinity.h` is now a CUDA-free param-reader header only (`read_cuda_device_id`, `read_numa_node_id`). `scoped_cuda_device.h` is removed — it was in the wrong location (public include tree). The public `decoder.h` interface remains free of CUDA and NUMA dependencies.

### Why the base class owns hardware affinity

The v1 approach put `ScopedCudaDevice` in the public include tree and required
each decoder plugin to opt in manually. That has two problems. First, a
third-party plugin author who forgets to add the guard gets silent wrong-device
behaviour with no error. Second, `scoped_cuda_device.h` pulled `<cuda_runtime.h>`
into the public API, which breaks the rule that `decoder.h` must be CUDA-free.

The v2 approach puts both RAII guards (`CudaDeviceGuard` and `NumaGuard`) in an
anonymous namespace inside `decoder.cpp`. They are invisible outside that
translation unit. The factory `decoder::get()` applies the guards during
construction and stores the ids in two protected members (`cuda_device_id_`,
`numa_node_id_`). Every decoder that goes through the factory gets correct
hardware affinity for free, including third-party plugins.

### Where guards are applied and why

| Call site | Guard applied | Reason |
|---|---|---|
| `decoder::get()` | Both guards | GPU allocations and NUMA first-touch happen here |
| `decoder::decode_batch()` | Both guards | One syscall for the whole batch, not per syndrome |
| `decoder::decode_async()` | Both guards, value-captured | Async thread starts with no inherited affinity |
| `decoder::decode()` | None | Sync callers in realtime loops manage their own thread state |

The single-syndrome `decode()` has no guard by design. Adding a CUDA API call
per syndrome in a tight realtime loop would add syscall overhead on every call.
Callers that need affinity on the sync path set it themselves before entering
the loop.

### The trt_decoder override

`trt_decoder` overrides `decode_batch()` entirely, which means the base class
version with its guard is never reached through virtual dispatch. A small inline
RAII struct at the top of `trt_decoder::decode_batch()` handles this. It reads
`cuda_device_id_` from the protected base-class member and uses
`cudaGetDevice`/`cudaSetDevice` directly, since the private `CudaDeviceGuard`
type is not visible outside `decoder.cpp`.

### No libnuma

NUMA binding uses raw Linux syscalls (`SYS_set_mempolicy`, `sched_setaffinity`)
behind `#if defined(__linux__)` with a one-time stderr warning on other
platforms. This avoids a new link dependency. The CPU affinity restore is
conditional on whether `sched_getaffinity` succeeded first, which avoids
permanently pinning a thread in containers with locked cpusets.

### What this does not cover

Decode-time affinity for the sync `decode()` path is left to the caller.
The `nv-qldpc-decoder` and realtime-session NUMA IPC pinning are out of scope
and tracked separately.

The single-syndrome `decode()` has no guard by design, adding a CUDA API call
and a NUMA syscall per syndrome in a realtime loop would add per-call overhead.
Callers on that path are expected to set affinity once before entering the loop.
All other paths (`decode_batch`, `decode_async`, construction) are fully covered.

## Tests

- `device_affinity` param readers (absent, int, size\_t, negative-throws)
- `decoder::get()` rejects out-of-range `cuda_device_id` for any decoder
- `decode_async` does not corrupt the calling thread's current device
- `trt_decoder` rejects invalid device ids and correctly runs `decode_async` on the assigned GPU end-to-end

## Out of scope

Does not address `nv-qldpc-decoder` runtime device handling or realtime-session NUMA IPC pinning.


tlshannon:
I think this works, but I'm thinking we might have to do it differently for the realtime api.  I think we either need to assume that the thread that creates the decoder is the thread to set affinity on (and just set it persistently during construction) or allow the user to set it outside of the decode/enqueue calls.  Perhaps we could set some flags to at least notify the user that they forgot to set the affinity, instead of silently ignoring it.  I'm not quite sure what all the user interfaces should be.  It seems like, in a realtime system, we should launch a thread that has the decoder on it and just set the affinity once for that thread.

libs/qec/lib/decoder.cpp
```diff
@@ -9,13 +9,143 @@
...
+#include <sstream>
+#include <sys/syscall.h>
+#include <unistd.h>
+#endif
+
+namespace {
+
+// RAII: sets the calling thread's CUDA current device, restores on destruction.
+// target < 0 = no-op. Throws std::runtime_error if target is out of range.
+struct CudaDeviceGuard {
```
bmhowe23: It's not clear to me that this approach works across all decoders. For example, I think the following example where the user is trying to run on different GPUs wouldn't actually run on different GPUs

```cpp
// Suppose "my_gpu_decoder" is a typical GPU decoder that:
//   (a) overrides only the single-syndrome decode(), not decode_batch(), and
//   (b) allocates its device buffers lazily on the first decode() call.
//
// Machine has >= 3 visible GPUs. Process starts with current device = 0.

auto d1 = cudaq::qec::decoder::get("my_gpu_decoder", H, {{"cuda_device_id", 1}});
auto d2 = cudaq::qec::decoder::get("my_gpu_decoder", H, {{"cuda_device_id", 2}});
// At this point decoder::get() pinned the device *around each constructor*,
// then restored it. But neither constructor allocated anything, so the guard
// protected nothing. Current device is back to 0.

auto r1 = d1->decode(syndrome1);   // <-- runs with current device == 0
auto r2 = d2->decode(syndrome2);   // <-- still current device == 0

// What actually happened:
//   * d1->decode(): base class does NOT wrap the single-syndrome decode().
//     d1 lazily cudaMalloc's its buffers NOW, on device 0 (the current
//     device), and launches its kernels on device 0. It never touched
//     device 1. cuda_device_id_ == 1 was stored but never consulted on this
//     path.
//   * d2->decode(): same -- allocates and runs on device 0, not device 2.
//
// Net result: both decoders run entirely on GPU 0. The cuda_device_id pins
// are silently ignored. (If d1 had instead eagerly allocated on device 1 in
// its ctor, you'd get the worse failure: device-1 memory accessed while
// device 0 is current -> cudaErrorIllegalAddress.)
```

kvmto: I was convinced that I covered also this case today, I need to test it better and think more about the API.
I am noticing there might be a necessary API change.

kvmto: I will also see how this fits with Sinter integrations.

kvmto:
I will split the PR in two or three.
Still finishing some things, but I should have tackled the majority of the use cases and comments now.
@bmhowe23 @tlshannon

libs/qec/lib/decoder.cpp
```diff
@@ -138,7 +393,29 @@ decoder::get(const std::string &name, const decoder_init &init,
...
-  return iter->second(init, param_map);
+  // Guards during construction so allocations land on the right hardware.
+  // Restored before this function returns; decode-time affinity is re-applied
+  // per call in decode_batch() and decode_async().
+  const int dev = cudaq::qec::read_cuda_device_id(param_map);
+  int node = cudaq::qec::read_numa_node_id(param_map);
+  if (node < 0 && dev >= 0)
+    node = numa_node_for_cuda_device(dev);
+  CudaDeviceGuard ctor_dev(dev);
+  NumaGuard ctor_numa(node); // preferred mode at construction
```
tlshannon: Should mempolicy be passed into this constructor as well?

libs/qec/lib/decoder_pool.cpp
```diff
@@ -0,0 +1,115 @@
...
+      jobs.push(std::move(j));
+    }
+    cv.notify_one();
+    return fut;
+  }
+
+  void run() {
+    // Construct + pin on this worker thread so resources land on the target
+    // GPU.
+    auto dec = decoder::get(spec.name, spec.H, spec.options);
```
tlshannon: Suggest putting this in a try/catch since CudaDeviceGuard can throw an exception.

kvmto: addressed in [dca1f0e](https://github.com/NVIDIA/cudaqx/pull/634/commits/dca1f0e4901bdb608a8d2cab5cd683e55bbe2286)

libs/qec/lib/decoder.cpp
kvmto: addressed in [dca1f0e](https://github.com/NVIDIA/cudaqx/pull/634/commits/dca1f0e4901bdb608a8d2cab5cd683e55bbe2286)

libs/qec/lib/decoder.cpp
kvmto: addressed in [dca1f0e](https://github.com/NVIDIA/cudaqx/pull/634/commits/dca1f0e4901bdb608a8d2cab5cd683e55bbe2286)

kvmto:
We will need to wait for new CI :) @bmhowe23 @anjbur

libs/qec/lib/decoders/plugins/trt_decoder/trt_decoder.cpp
```diff
@@ -888,6 +892,86 @@ decoder_result trt_decoder::decode(const std::vector<float_t> &syndrome) {
...
+    if (_prev_dev != cuda_device_id_) {
+      cudaError_t _set_err = cudaSetDevice(cuda_device_id_);
+      if (_set_err != cudaSuccess)
+        throw std::runtime_error("trt_decoder::decode_batch: cudaSetDevice(" +
+                                 std::to_string(cuda_device_id_) +
+                                 ") failed: " + cudaGetErrorString(_set_err));
+      _dev_switched = true;
+    }
+  }
+  struct RestoreDevice {
```
tlshannon: Suggest consolidating the copies of the device/NUMA set-restore logic.

The same scoped set-then-restore contract now lives in multiple places: CudaDeviceGuard/NumaGuard (decoder.cpp), RestoreDevice/ScopedNuma (trt_decoder.cpp), and inline in bind_current_thread() (decoder.cpp). Each hand-maintains the same subtle invariants, so a fix to any one has to be mirrored in the others or they drift.

hardware_affinity.h is already lib-private, header-only, and shared across the plugin .so boundary (the reason trt_decoder hand-rolls its own), so a single reusable RAII guard (with helper functions) there could replace them. Not a blocker, but it would remove the divergence risk.

kvmto: Thanks a lot, Tracy!

Good catch. I fixed it.

The guards now live in a single `lib-private/hardware_guards.h` header. Since it is header-only, it works across the plugin `.so` boundary. The base entry points, the realtime session, and `trt_decoder::decode_batch` all use the same pair of guards. The inline `RestoreDevice` and `ScopedNuma` implementations here have been removed.

The only exception is `bind_current_thread()`. It intentionally keeps its inline logic because it restores the previous state only on failure. On success, the new binding is meant to persist, so a scoped guard is not a good fit. It still reuses the shared capture and restore helpers.

This also makes error handling consistent. If the current device cannot be read, every entry point now logs a warning and skips restoration. Previously, the TRT path threw an exception. Out of range device IDs now produce the same error everywhere.

tlshannon (approved):
LGTM

bmhowe23 (changes requested):
Thanks, Kevin ... this looks really thorough. As we discussed offline, there may be some use cases that we have not fully thought through for the theading and NUMA pinning. Therefore, I think it would be best to break this down into smaller PRs.

1. Decoder-specific changes to support "cuda_device_id". This one is the near-term priority and is likely the only one needed for this release. While the main goal is YAML support and C++ use cases, I think that this should likely automagically work for Python if we just implement the decoder-specific kwargs like all of our other kwargs.
2. Handle NUMA memory pinning and better threading understanding/support. It would be great to capture what you have in this PR into another draft PR. But it may be next release before we work through all the use cases there.
3. Python interfaces for 2

PR #636, opened 2026-06-26
Add detector support to memory circuits

### LINKED ISSUE #556: Decoded logical error rate does not decrease with code distance on surface code
## Summary

When decoding a `stim`-generated `surface_code:rotated_memory_z` circuit with `cudaq_qec.get_decoder("nv-qldpc-decoder", ...)`, the script runs to completion without any error or warning, but the **decoded** logical error rate (LER) does **not** decrease as the surface code distance `d` increases at sub-threshold physical error rates. The decoded curve sits essentially on top of the raw / no-QEC curve, instead of separating from it and dropping exponentially with `d`.

## Reproduction

Full standalone script (also available as a gist:
<https://gist.github.com/bmhowe23/21ed66ceecc0972344f7b61b44e58186>):

```python
import numpy as np
import cudaq_qec as qec
import stim
from beliefmatching.belief_matching import detector_error_model_to_check_matrices


def run_cudaq_simulation(distance, physical_error_rate):
    nShots = 100_000
    nRounds = distance

    stim_surface_code = stim.Circuit.generated(
        "surface_code:rotated_memory_z",
        rounds=nRounds,
        distance=distance,
        after_clifford_depolarization=physical_error_rate,
        after_reset_flip_probability=physical_error_rate,
        before_measure_flip_probability=physical_error_rate,
        before_round_data_depolarization=physical_error_rate,
    )

    dem = stim_surface_code.detector_error_model(decompose_errors=True)
    matrices = detector_error_model_to_check_matrices(dem)
    H, O, priors = matrices.check_matrix, matrices.observables_matrix, matrices.priors

    decoder = qec.get_decoder(
        "nv-qldpc-decoder",
        H.todense(order="C").astype(np.uint8),
        error_rate_vec=priors,
        use_sparsity=True,
        bp_batch_size=1000,
        # OSD-related parameters left at defaults
    )

    syndromes, obs = stim_surface_code.compile_detector_sampler().sample(
        shots=nShots, separate_observables=True
    )
    logical_measurements = obs.flatten()

    dr = decoder.decode_batch(syndromes)
    error_predictions = np.array([e.result for e in dr], dtype=np.uint8)
    data_predictions = (O @ error_predictions.T) % 2

    ler_decoded = np.sum(data_predictions ^ logical_measurements) / nShots
    ler_raw     = np.sum(logical_measurements) / nShots
    return ler_decoded, ler_raw


for p in [1e-4, 1e-3]:
    for d in [3, 5, 7, 11, 13, 15, 17]:
        ler_decoded, ler_raw = run_cudaq_simulation(d, p)
        print(f"d={d:2d}, p={p:.0e}: decoded={ler_decoded:.4e}, raw={ler_raw:.4e}")
```

### Observed output

```
Running simulation for distance=3, p=0.0001...
LER decoded: 0.00019, LER raw: 0.00236
Running simulation for distance=5, p=0.0001...
LER decoded: 7e-05, LER raw: 0.00584
Running simulation for distance=7, p=0.0001...
LER decoded: 7e-05, LER raw: 0.01169
Running simulation for distance=11, p=0.0001...
LER decoded: 0.00017, LER raw: 0.0276
Running simulation for distance=3, p=0.001...
LER decoded: 0.00097, LER raw: 0.02238
Running simulation for distance=5, p=0.001...
LER decoded: 0.00284, LER raw: 0.05789
Running simulation for distance=7, p=0.001...
LER decoded: 0.00609, LER raw: 0.10409
Running simulation for distance=11, p=0.001...
LER decoded: 0.01683, LER raw: 0.21567
```

Image

### Expected behavior

For `p` well below threshold, the decoded LER should decrease (roughly exponentially) with `d` and clearly separate from the raw curve.

## Environment

- Install method: `pip install cudaq-qec` in cudaq stable docker container.
- `cudaq_qec.__version__`: `0.6.0`
- `cudaq.__version__`: `0.14.2`
- `stim.__version__`: `1.15.0`

eliotheinrich:
## Description

The addition of detectors and logical observables https://github.com/NVIDIA/cuda-quantum/pull/4602 means we can begin adding detectors and observable annotations to our QPU kernels. This will replace the previous ad-hoc temporal structure in, i.e., `sample_memory_circuit`, which will also address https://github.com/NVIDIA/cudaqx/issues/556. 

This PR adds detector/observable annotations to the `memory_circuit` kernel, and uses them to output syndromes in `sample_memory_circuit` and `dem_from_memory_circuit`, along with some supporting changes. These changes are closely linked, and so are bundled into a single PR, although this does push the diff over 1000 lines. 

It also adds `decompose_errors` as a flag to `dem_from_memory_circuit`, allowing for the Stim error analyzer to output graphlike approximations of the DEM (see https://github.com/NVIDIA/cudaqx/pull/615). The API for `sample_memory_circuit` is unchanged, although `x_sample_memory_circuit` and `z_sample_memory_circuit` have been added as new functions to be used alongside the existing `x_dem_from_memory_circuit` and `z_dem_from_memory_circuit`. 

It also creates a breaking change by changing the output shape of the syndromes from `sample_memory_circuit` from `(nShots * nRounds) x nAncilla` to `nShots x nDetectors`. The detectors are formatted as follows: for each shot, the first nAncillax(z) bits correspond to the outcomes of measurements in the X(Z) basis, depending on if the state was prepared in the X(Z) basis. The final nAncillax(z) bits correspond to final-round stabilizer measurements in the X(Z) basis. The intermediate bits correspond to inter-round detectors (i.e. m_t \xor m_t+1). This is a more natural way to format the data (allowing the removal of several post-processsing reshapes) and is more inline with, i.e., Stim.

Two new tests are added, one for `x(z)_sample_memory_circuit`: test_dem.py::test_sample_memory_circuit_aligned_with_dem`, and one for the new detector format described in the previous paragraph: `test_dem.py::test_dem_from_memory_circuit_boundary_row_format`. Note that many of the existing tests have also had to be modified to match the new format. Please take a look.

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


libs/qec/lib/device/memory_circuit.cpp
```diff
@@ -6,61 +6,93 @@
...
-    const std::vector<std::size_t> &z_stabilizers) {
+__qpu__ void memory_circuit(const code::stabilizer_round &stabilizer_round,
+                            const code::one_qubit_encoding &statePrep,
+                            std::size_t num_data, std::size_t numAncx,
+                            std::size_t numAncz, std::size_t num_rounds,
+                            const std::vector<std::size_t> &x_stabilizers,
+                            const std::vector<std::size_t> &z_stabilizers,
+                            const std::vector<std::size_t> &obs_matrix_flat,
+                            std::size_t num_observables,
+                            bool measure_in_x_basis) {
```
justinlietz, line 21: In a memory experiment, the state prep is matched to the measurement basis. We could trim off one of the input variables by either using `statePrep` to infer the measurement basis, or by selecting the appropriate statePrep from the measure basis flag.

libs/qec/lib/device/memory_circuit.cpp
```diff
@@ -6,61 +6,93 @@
...
+  cudaq::qvector data(num_data), xstab_anc(numAncx), zstab_anc(numAncz);
+
   // Create the logical patch
   patch logical(data, xstab_anc, zstab_anc);
 
   // Prepare the initial state fault tolerantly
   statePrep({data, xstab_anc, zstab_anc});
 
+  // Z-measurement before any operations always yields 0...0, so first syndromes
+  // are partially non-deterministic
```
justinlietz:
```suggestion
  // The "off-basis" detectors will be non-deterministic after the first stabilizer round.
```

libs/qec/lib/device/memory_circuit.cpp
```diff
@@ -6,61 +6,93 @@
...
+  auto data_results = mz(data);
 
-  // Persists ancilla measures
-  __memory_circuit_stabs(data, xstab_anc, zstab_anc, stabilizer_round,
-                         statePrep, numRounds, x_stabilizers, z_stabilizers);
+  // Emit one logical_observable per row of the observable matrix.
+  for (std::size_t obs = 0; obs < num_observables; ++obs) {
+    std::size_t support_weight = 0;
+    for (std::size_t q = 0; q < num_data; ++q) {
+      if (obs_matrix_flat[obs * num_data + q] != 0)
```
justinlietz, line 58: The restriction on vectors only makes this section pretty tough with all the index gymnastics. I'm wondering if we should pass in the sparse form of the observable, a matrix of indices which are row separated by `-1`. It's not particularly clean either, but worth considering depending on your taste.

Same with detectors below.

libs/qec/lib/device/memory_circuit.cpp
```diff
@@ -6,61 +6,93 @@
...
+                            const std::vector<std::size_t> &obs_matrix_flat,
+                            std::size_t num_observables,
+                            bool measure_in_x_basis) {
+  // Allocate the data and ancilla qubits
+  cudaq::qvector data(num_data), xstab_anc(numAncx), zstab_anc(numAncz);
+
   // Create the logical patch
   patch logical(data, xstab_anc, zstab_anc);
 
   // Prepare the initial state fault tolerantly
```
justinlietz:
```suggestion
  // Prepare the initial state
```
I do not think we impose that this is fault tolerant, so perhaps we can clean this comment up while we're here.

libs/qec/lib/device/memory_circuit.cpp
```diff
@@ -6,61 +6,93 @@
...
+      if (obs_matrix_flat[obs * num_data + q] != 0)
+        obs_support[idx++] = data_results[q];
+    }
+    cudaq::logical_observable(obs_support);
+  }
 
-  auto dataResults = mz(data);
-}
+  // For each syndrome, connect output from final syndrome round to each qubit
+  // in the support of that stabilizer.
```
justinlietz:
```suggestion
  // For each stabilizer, form detectors from data qubit readout connected with final stabilizer round.
```

libs/qec/python/tests/test_code.py
```diff
@@ -53,19 +53,23 @@ def test_code_stabilizers():
 
 def test_sample_memory_circuit():
     steane = qec.get_code("steane")
+    nShots, nRounds = 10, 4
+    # Shape: (nShots, k) where k = 2*numAncz + (nRounds-1)*numCols
+    # For Steane: numAncz=3, numCols=6, nRounds=4 → k = 2*3 + 3*6 = 24
```
justinlietz, line 59: This is true, but I like to think of it as:
`num_on_basis_stabilizers + (nRounds-1)*num_total_stabilzers + num_on_basis_stabilizers =  2*num_on_basis_stabilizers + (nRounds-1)*num_total_stabilzers`

Don't have to modify anything, but it looks a bit less mysterious this way.

justinlietz (commented):
This is looking great Eliot, thanks! A great improvement for the library.

libs/qec/lib/device/memory_circuit.cpp
eliotheinrich: I tried, but the statePrep parameter is a qkernel, and I don't think we can introspect on it to determine the basis. And the statePrep kernel itself is a field of the qec::code, which does not seem to be a valid parameter type for QPU kernels at the moment, so I think we might be stuck with the extra parameter :(

libs/qec/lib/device/memory_circuit.cpp
eliotheinrich: I agree it's a little gnarly, but mocking up the sparse version doesn't look much better. It would also require more rewiring in the caller functions. IMO we can leave it as is unless there's i.e. a performance reason to change it later.

libs/qec/python/tests/test_code.py
```diff
@@ -95,24 +99,27 @@ def test_noisy_simulation():
...
                                                        noise=noise)
     assert isinstance(syndromes, np.ndarray)
-    assert syndromes.shape == (40, 6)
+    assert syndromes.shape == (nShots, nDetectors)
     print(syndromes)
     assert np.any(syndromes)
     cudaq.reset_target()
 
 
+@pytest.mark.skip
```
eliotheinrich, line 115: I need to disable this test for now; I have raised an issue here: https://github.com/NVIDIA/cudaqx/issues/646

kvmto:
I had to run the usual code review. Things that maybe your agent can automate, but please address the docs comment :) those are important since the docs were a bit skinny:

1- Weakened Stim test assertions
In checkSteaneNoiseStim, all five sub-cases used to check X and Z syndromes separately (for example, bitflip noise asserted x_sum > 0 and z_sum == 0). They now only check that the total sum is greater than zero. I understand the old half/half column split no longer applies with the new detector layout, but could we re-derive the X vs Z check using the new layout? The test comments already show where the fixed-type columns live, so it seems doable. As written, a bug that fires the wrong stabilizer type would pass.

2- numRounds=0 behavior
The new kernel runs one stabilizer_round unconditionally and then loops from round 1. So numRounds=0 still executes one round. In experiments.cpp the data offset is computed as numCols * numRounds, which is 0 in that case, so the returned data results would actually contain ancilla bits. The old kernel looped from round 0 and handled this correctly. Should we add a guard that rejects numRounds=0, or restore the old behavior?

3- Doc build: stale doxygenfunction signatures
cpp_api.rst lines 33 to 35 still pin the 4-argument signatures for dem_from_memory_circuit, x_dem_from_memory_circuit, and z_dem_from_memory_circuit, but this PR adds a fifth decompose_errors parameter to all three. Won't these directives fail to resolve in the Sphinx build? Can you update them?

4- Stale formula in the docs
The sliding-window decoder example in introduction.rst (both the Python and C++ tabs) still computes num_syndromes_per_round as detector_error_matrix.shape[0] // num_rounds. With the new boundary detector rows the row count is no longer num_rounds * syndromes_per_round, so this formula gives the wrong answer now. Can you update the example?

5- noise=None inconsistency
The new x_sample_memory_circuit and z_sample_memory_circuit Python bindings throw an error when noise=None, but sample_memory_circuit silently falls back to noiseless simulation. Since these are advertised as variants of the same function, should they behave the same way? Which behavior do you want?

6- Duplicated binding code (minor)
The x_sample_memory_circuit and z_sample_memory_circuit bindings are identical lambdas except for the function name and error string. Could they share one helper? Same for the two "count then gather" loops in memory_circuit.cpp. Also, the new #include <numeric> in experiments.cpp looks unused.

bonus 7- Asymmetric codes (design question, not a blocker)
For a code where the number of X and Z stabilizers differ, the boundary blocks have numFixed rows but canonicalize_for_rounds divides by the full per-round count, so the round grouping would misalign. No in-tree code hits this today. Is that a known limitation worth documenting?

eliotheinrich:
> I had to run the usual code review. Things that maybe your agent can automate, but please address the docs comment :) those are important since the docs were a bit skinny:

Thanks for the review! I've pushed the changes to dded61d4fbafd8da152b1320436a9285bd65594c, and the docs changes to d83ddb3f363f5dbb19cb504bcbe30af3ea3da820. 

> 1- Weakened Stim test assertions In checkSteaneNoiseStim, all five sub-cases used to check X and Z syndromes separately (for example, bitflip noise asserted x_sum > 0 and z_sum == 0). They now only check that the total sum is greater than zero. I understand the old half/half column split no longer applies with the new detector layout, but could we re-derive the X vs Z check using the new layout? The test comments already show where the fixed-type columns live, so it seems doable. As written, a bug that fires the wrong stabilizer type would pass.

Indeed; I think I originally changed them to address the fact that the first round syndrome bits were non-deterministic in the original implementation, but I guess I got overzealous. I've added them back in with X/Z slicing logic.

> 2- numRounds=0 behavior The new kernel runs one stabilizer_round unconditionally and then loops from round 1. So numRounds=0 still executes one round. In experiments.cpp the data offset is computed as numCols * numRounds, which is 0 in that case, so the returned data results would actually contain ancilla bits. The old kernel looped from round 0 and handled this correctly. Should we add a guard that rejects numRounds=0, or restore the old behavior?

Good catch. I added a runtime_error when numRounds=0; I think this makes the most sense. A zero-round memory circuit is funky for a bunch of reasons, i.e. a code with transversal state prep won't even be prepared in the logical codespace before exiting in this scenario. It would also require special-case handling in a lot of places (like the one you pointed out).

> 3- Doc build: stale doxygenfunction signatures cpp_api.rst lines 33 to 35 still pin the 4-argument signatures for dem_from_memory_circuit, x_dem_from_memory_circuit, and z_dem_from_memory_circuit, but this PR adds a fifth decompose_errors parameter to all three. Won't these directives fail to resolve in the Sphinx build? Can you update them?

> 4- Stale formula in the docs The sliding-window decoder example in introduction.rst (both the Python and C++ tabs) still computes num_syndromes_per_round as detector_error_matrix.shape[0] // num_rounds. With the new boundary detector rows the row count is no longer num_rounds * syndromes_per_round, so this formula gives the wrong answer now. Can you update the example?

Yes, good catches. Both are fixed in the new commit. For the introduction.rst examples, the number of extracted syndromes per round is equal to the number of z-stabilizers, since these examples are only correcting Z-errors. 
 
> 5- noise=None inconsistency The new x_sample_memory_circuit and z_sample_memory_circuit Python bindings throw an error when noise=None, but sample_memory_circuit silently falls back to noiseless simulation. Since these are advertised as variants of the same function, should they behave the same way? Which behavior do you want?

Thanks for catching that. It should behave like the default and fall back to noiseless; changed now.

> 6- Duplicated binding code (minor) The x_sample_memory_circuit and z_sample_memory_circuit bindings are identical lambdas except for the function name and error string. Could they share one helper? Same for the two "count then gather" loops in memory_circuit.cpp. Also, the new #include in experiments.cpp looks unused.

Indeed; they're now consolidated. Since x/z_dem_from_memory_circuit follow the same pattern, I also consolidated them with a helper lambda. 

> bonus 7- Asymmetric codes (design question, not a blocker) For a code where the number of X and Z stabilizers differ, the boundary blocks have numFixed rows but canonicalize_for_rounds divides by the full per-round count, so the round grouping would misalign. No in-tree code hits this today. Is that a known limitation worth documenting?

Good point; might be worth a follow-up. As is, I think it's a bit beyond the scope of this PR. I can post an issue if we agree.

eliotheinrich:
/ok to test [dded61d](https://github.com/NVIDIA/cudaqx/pull/636/commits/dded61d4fbafd8da152b1320436a9285bd65594c)

docs/sphinx/api/qec/sliding_window_api.rst
```diff
@@ -66,7 +66,7 @@
             opts = {
                 'error_rate_vec': np.array(dem.error_rates),
                 'window_size': 1,
-                'num_syndromes_per_round': dem.detector_error_matrix.shape[0] // num_rounds,
+                'num_syndromes_per_round': code.get_num_z_stabilizers(),
```
bmhowe23, line 69: This is kind of related to @kvmto's comment, and I think it merits a follow-up PR. With the changes done in this PR, the sliding window decoder can no longer really take a single parameter called "num_syndromes_per_round" (because of the reduced size in the first and last round).

bmhowe23 (approved):
Thanks, @eliotheinrich. We should just open a new issue for the one item I and Kevin mentioned above (canonicalize_for_rounds and sliding window decoder tweaks).

eliotheinrich:
@kvmto @bmhowe23 I've raised a follow-up issue to address the sliding window decoder and canonicalize_for_rounds format issues: https://github.com/NVIDIA/cudaqx/pull/636

PR #637, opened 2026-06-26
qec: add surface code orientation support (XV/XH/ZV/ZH)

melody-ren:
## Summary

The rotated `surface_code` previously hard-coded a single stabilizer layout.
This PR adds a first-class `sc_orientation` enum and threads it through
`stabilizer_grid`, letting callers choose which Pauli type occupies the bulk
interior and which boundary pair carries X- vs Z-type stabilizers.

The orientation names follow Ising’s `code_rotation` geometry convention. To
preserve existing CUDA-QX behavior, the default orientation is `ZH`. Existing
callers that do not set an orientation keep the same stabilizers, observables,
and DEM behavior.

## What's Added

- `sc_orientation` enum: `XV`, `XH`, `ZV`, `ZH`.
- `sc_orientation_from_str(const std::string&)`: converts canonical names and
  Ising aliases to `sc_orientation`, case-insensitively with surrounding
  whitespace ignored:
  - `XV` / `O1`
  - `XH` / `O2`
  - `ZV` / `O3`
  - `ZH` / `O4`
- `stabilizer_grid` gains an orientation-aware constructor:
  `stabilizer_grid(uint32_t distance, sc_orientation orientation = sc_orientation::ZH)`.
- `stabilizer_grid` stores orientation privately and exposes it through
  `get_orientation()`.
- `generate_grid_roles()` now derives the bulk checkerboard and boundary
  stabilizer types from the selected orientation.
- `surface_code(const heterogeneous_map&)` accepts an `"orientation"` option,
  defaulting to `"ZH"`.
- Python bindings expose `sc_orientation` and the orientation-aware
  `stabilizer_grid` constructor, accepting either the enum or a string.
  The grid’s orientation is available as a read-only Python property.

## Orientation Semantics

The names follow Ising’s `code_rotation` convention:

- first character: bulk syndrome type
- second character: `rotated_type`

| Orientation | X-type boundaries | Z-type boundaries |
| --- | --- | --- |
| `XV` | left / right | top / bottom |
| `XH` | top / bottom | left / right |
| `ZV` | top / bottom | left / right |
| `ZH` | left / right | top / bottom |

`XV`/`ZH` share boundary placement but differ in the bulk X/Z checkerboard.
Likewise, `XH`/`ZV` share boundary placement but differ in the bulk checkerboard.
All four are distinct layouts.

## Observables

`get_spin_op_observables()` is orientation-aware and returns a valid logical
pair for the selected orientation:

| Orientation | X observable | Z observable |
| --- | --- | --- |
| `XV`, `ZH` | top row | left column |
| `XH`, `ZV` | left column | top row |

The default `ZH` orientation preserves the previous observable convention:
X along the top row and Z along the left column.

## Testing

C++ (`test_qec.cpp`):

- `checkSurfaceCodeOrientationFromString`: string conversion, alias handling,
  invalid-input rejection, and construction via `get_code(..., {"orientation", ...})`.
- `checkSurfaceCodeDefaultZHLayout`: verifies the default `ZH` layout.
- `checkSurfaceCodeAllOrientationGeometries`: pins the exact distance-3 role
  grid for all four orientations, guarding against label swaps or boundary
  regressions.
- `checkSurfaceCodeOrientationAlgebra`: verifies all stabilizers commute, the
  logical pair is valid for all four orientations, and the expected observable
  supports are returned.

Python (`test_surface_code.py`):

- Orientation-aware `stabilizer_grid` construction using enum, canonical string,
  and Ising alias string forms.
- Read-only orientation inspection from Python.
- `get_code(..., orientation=...)` coverage for orientation strings.

## Backward Compatibility

Existing callers that do not specify an orientation are unaffected. The default
is now named `ZH` to match Ising geometry naming, but it is the same physical
layout CUDA-QX previously used.

This is a self-contained library/bindings change with no new runtime
dependencies.

### DIFF
--- a/libs/qec/include/cudaq/qec/codes/surface_code.h
+++ b/libs/qec/include/cudaq/qec/codes/surface_code.h
@@ -36,8 +36,10 @@ enum surface_role { amx, amz, empty };
 /// row). See get_spin_op_observables() for details.
 enum class sc_orientation { XV, XH, ZV, ZH };
 
-/// @brief Parse a surface-code orientation string (XV, XH, ZV, or ZH).
-sc_orientation parse_orientation(const std::string &s);
+/// @brief Convert a surface-code orientation string to a sc_orientation enum
+/// (XV, XH, ZV, ZH, or Ising's O1-O4 aliases, where O1/O2/O3/O4 map to
+/// XV/XH/ZV/ZH).
+sc_orientation sc_orientation_from_str(const std::string &s);
 
 /// @brief describes the 2d coordinate on the stabilizer grid
 struct vec2d {
@@ -69,9 +71,8 @@ bool operator<(const vec2d &lhs, const vec2d &rhs);
 ///
 /// Each entry on the grid can be an X stabilizer, Z stabilizer,
 /// or empty, as is needed on the edges.
-/// The diagrams below show the default ZH orientation, which preserves the
-/// original fixed surface-code layout. Other orientations assign X/Z roles
-/// according to sc_orientation.
+/// The diagrams below show the default ZH orientation. Other orientations
+/// assign X/Z roles according to sc_orientation.
 /// The grid length of 4 corresponds to a distance 3 surface code, which results
 /// in:
 /// ```
@@ -122,15 +123,15 @@ public:
   /// determines the number of data qubits per dimension
   uint32_t distance = 0;
 
-  /// @brief Orientation used to assign stabilizer roles. The default ZH is the
-  /// legacy fixed surface-code layout, named per Ising's code_rotation.
-  sc_orientation orientation = sc_orientation::ZH;
-
   /// @brief length of the stabilizer grid
   /// for distance = d data qubits,
   /// the stabilizer grid has length d+1
   uint32_t grid_length = 0;
 
+private:
+  sc_orientation orientation_ = sc_orientation::ZH;
+
+public:
   /// @brief flattened vector of the stabilizer grid sites roles'
   /// grid idx -> role
   /// stored in row major order
@@ -170,6 +171,9 @@ public:
                   sc_orientation orientation = sc_orientation::ZH);
   /// @brief Empty constructor
   stabilizer_grid();
+
+  /// @brief Get the orientation used to construct this stabilizer grid.
+  sc_orientation get_orientation() const;
 
   /// @brief Print a 2d grid of stabilizer roles
   void print_stabilizer_grid() const;
--- a/libs/qec/include/cudaq/qec/codes/surface_code.h
+++ b/libs/qec/include/cudaq/qec/codes/surface_code.h
@@ -22,13 +22,11 @@ enum surface_role { amx, amz, empty };
 /// @brief Surface-code orientation: controls which Pauli type (X or Z) occupies
 /// the bulk interior and which boundaries are X-type vs Z-type.
 ///
-/// Names follow Ising's `code_rotation` convention (first character = bulk
-/// syndrome type, second character = rotated_type), so a given name denotes the
-/// SAME physical stabilizer layout in cudaqx and Ising — an Ising-trained model
-/// can be matched by name. By geometry: XV and ZH place X-type stabilizers on
-/// the left/right boundaries (Z on top/bottom); XH and ZV place X-type on
-/// top/bottom (Z on left/right). XV vs ZH (and XH vs ZV) differ in the bulk X/Z
-/// checkerboard.
+/// Geometry names follow Ising's `code_rotation` convention (first character =
+/// bulk syndrome type, second character = rotated_type). By geometry: XV and ZH
+/// place X-type stabilizers on the left/right boundaries (Z on top/bottom); XH
+/// and ZV place X-type on top/bottom (Z on left/right). XV vs ZH (and XH vs ZV)
+/// differ in the bulk X/Z checkerboard.
 ///
 /// Observable convention: get_spin_op_observables() returns the valid logical
 /// pair for the orientation. XV/ZH use X obs along the top row and Z obs along
--- a/libs/qec/include/cudaq/qec/codes/surface_code.h
+++ b/libs/qec/include/cudaq/qec/codes/surface_code.h
@@ -67,6 +67,9 @@ bool operator<(const vec2d &lhs, const vec2d &rhs);
 ///
 /// Each entry on the grid can be an X stabilizer, Z stabilizer,
 /// or empty, as is needed on the edges.
+/// The diagrams below show the default ZH orientation, which preserves the
+/// original fixed surface-code layout. Other orientations assign X/Z roles
+/// according to sc_orientation.
 /// The grid length of 4 corresponds to a distance 3 surface code, which results
 /// in:
 /// ```
@@ -188,6 +191,9 @@ public:
   std::vector<cudaq::spin_op_term> get_spin_op_stabilizers() const;
 
   /// @brief Get the observables as a vector of cudaq::spin_op_terms
+  ///
+  /// @return The X logical observable first, followed by the Z logical
+  /// observable.
   ///
   /// @note Returns the correct logical pair for the grid's orientation. For XV
   /// and ZH the X observable runs along the top row of data qubits and the Z
--- a/libs/qec/include/cudaq/qec/codes/surface_code.h
+++ b/libs/qec/include/cudaq/qec/codes/surface_code.h
@@ -9,6 +9,7 @@
 
 #include "cudaq/qec/code.h"
 #include "cudaq/qec/patch.h"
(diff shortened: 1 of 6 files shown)


libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -18,6 +19,33 @@ namespace cudaq::qec::surface_code {
 /// grid
 enum surface_role { amx, amz, empty };
 
+/// @brief Surface-code orientation: controls which Pauli type (X or Z) occupies
+/// the bulk interior and which boundaries are X-type vs Z-type.
+///
+/// Within cudaqx: XV and ZH put X-type stabilizers on left/right boundaries;
+/// XH and ZV put them on top/bottom.
+///
+/// WARNING — do NOT map these to Ising by name string. Ising has two naming
```
bmhowe23: This sounds like we're asking for interoperability trouble. Is there a problem with following Ising's convention here? I fear that 99.9% people would miss this warning.

melody-ren: Yup agreed, and can't blame that 99.9%. This was basically a relic from an earlier version when I didn't want to change `get_spin_op_observables()` since that would require a large amount of downstream updates. But when I was doing self-review, I realized leaving that untouched would give you the wrong observables so I fixed that. With that fixed, I agree that now we should change the labels to follow Ising's convention. Relabeling commit is coming

melody-ren: Fixed this in 54fac96acb698ea0a7c276409efed973c7db1f29

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -18,6 +19,24 @@ namespace cudaq::qec::surface_code {
 /// grid
 enum surface_role { amx, amz, empty };
 
+/// @brief Surface-code orientation: controls which Pauli type (X or Z) occupies
+/// the bulk interior and which boundaries are X-type vs Z-type.
+///
+/// Geometry names follow Ising's `code_rotation` convention (first character =
```
melody-ren: Notes to self: Add Ising-decoding repo URL here

melody-ren: Fixed in 5b1fd174c2ed329668b1e43880a9c71b89a3d535

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -18,6 +19,26 @@ namespace cudaq::qec::surface_code {
 /// grid
 enum surface_role { amx, amz, empty };
 
+/// @brief Surface-code orientation: controls which Pauli type (X or Z) occupies
+/// the bulk interior and which boundaries are X-type vs Z-type.
+///
+/// Geometry names follow the NVIDIA/Ising-Decoding naming convention
+/// (https://github.com/NVIDIA/Ising-Decoding): Ising's `code_rotation`
```
melody-ren, line 26: Now that I think about it, should I also use Ising's user facing naming convention (e.g, O1, O2, O3 etc instead of XV/ZH etc)? Ising uses XV/ZH etc as internal names

bmhowe23: Yes, you can use both. It would definitely be good to support O1-O4.

libs/qec/unittests/test_qec.cpp
```diff
@@ -682,6 +763,221 @@ TEST(QECNoiseModelTester, ConstructsTwoQubitBitflipChannel) {
...
+    EXPECT_EQ(surface_role::empty, role_at(3, 3));
+  }
+}
+
+// After making get_spin_op_observables() orientation-aware, every orientation
+// must yield a VALID logical pair: the X observable commutes with all Z
+// stabilizers, the Z observable commutes with all X stabilizers, and the two
+// observables anticommute. This also pins which physical support each
+// orientation uses (top row vs left column).
+TEST(QECCodeTester, checkSurfaceCodeObservablesPerOrientation) {
```
eliotheinrich: I think this test is mostly a duplicate of `expect_surface_code_algebra`, except that this one tests `.get_spin_op_observables()`. You could probably just pull that into the algebra test case (i.e. check that `support_of(observables[0]) == uses_horizontal_x_logical(orientation) ? top_row : left_column` and remove this test?

libs/qec/lib/codes/surface_code.cpp
```diff
@@ -6,11 +6,59 @@
...
+    // character controls boundary placement only, so XV and XH share this bulk
+    // assignment.
+    return odd_parity ? surface_role::amz : surface_role::amx;
+  case sc_orientation::ZV:
+  case sc_orientation::ZH:
+    // Likewise, ZV and ZH share a bulk assignment with Z on even parity and X
+    // on odd parity.
+    return odd_parity ? surface_role::amx : surface_role::amz;
+  }
+  throw std::runtime_error("Unhandled surface-code orientation.");
```
eliotheinrich: This switch is a bit gnarly; you could probably replace it with something like
```
bool is_x_type = (orientation == XV || orientation == XH);
if (is_x_type) 
    return odd_parity ? surface_role::amz : surface_role::amx;
else
    return odd_parity ? surface_role::amx : surface_role::amz;
```
The throw is unreachable either way.

libs/qec/python/tests/test_surface_code.py
```diff
@@ -25,6 +25,20 @@ def test_basic_construction_d3():
...
+    assert g.orientation == qec.sc_orientation.XH
+    assert len(g.x_stabilizers) == 4
+    assert len(g.z_stabilizers) == 4
+
+    canonical_grid = qec.stabilizer_grid(3, "ZV")
+    assert canonical_grid.orientation == qec.sc_orientation.ZV
+    assert len(canonical_grid.x_stabilizers) == 4
+    assert len(canonical_grid.z_stabilizers) == 4
+
+
```
eliotheinrich, line 44: I think a small test to using an 'oriented' surface code in Python wouldn't go amiss, just to be sure the Python version of the code isn't misbehaving. I.e. checking that `qec.get_code("surface_code", distance=3, orientation="ZV"})` is working as expected.

libs/qec/include/cudaq/qec/codes/surface_code.h
```diff
@@ -48,6 +69,9 @@ bool operator<(const vec2d &lhs, const vec2d &rhs);
 ///
 /// Each entry on the grid can be an X stabilizer, Z stabilizer,
 /// or empty, as is needed on the edges.
+/// The diagrams below show the default ZH orientation, which preserves the
+/// original fixed surface-code layout. Other orientations assign X/Z roles
```
eliotheinrich: nitpick: Does it make sense to refer to ZH as the 'original'/'legacy' orientation if the distinction seems irrelevant after this PR is merged?

eliotheinrich (commented):
@melody-ren Looks good! Can't find anything incorrect or inconsistent about the orientations. I posted a few nitpicks for your consideration.

libs/qec/include/cudaq/qec/codes/surface_code.h
melody-ren: Added the O1-O4 naming support in 65c519ebcf3eb4e2736c697455b638c2bd7b5754

libs/qec/unittests/test_qec.cpp
melody-ren: Good catch. Fixed in 65c519ebcf3eb4e2736c697455b638c2bd7b5754

libs/qec/lib/codes/surface_code.cpp
melody-ren: oh yes, good eye. Fixed in 65c519ebcf3eb4e2736c697455b638c2bd7b5754

libs/qec/python/tests/test_surface_code.py
melody-ren: Thanks! Great catch. I lose count the number of times I forgot to add coverage for the python side. 

While I was adding the python test, I realized another "hole for the future me" issue. Namely, the API design for `parse_orientation` made it seem like you could set the orientation after construction, which was not the intended behaviour. I fixed that by making orientation a private member and changing the helper `parse_orientation` to `sc_orientation_from_str` to make the intended use case clear. Thanks for the review!

libs/qec/include/cudaq/qec/codes/surface_code.h
melody-ren: Ah yes, that wording was useful when I was working through the change. That aged like milk. Removed all mention of similar wording. Thanks!

eliotheinrich (approved):
Looks good to me!

PR #647, opened 2026-07-01
Add a surface code example to demonstrate NVQLink

melody-ren:
## Summary

This PR adds a YAML-configurable realtime surface-code example that streams CUDA-QX measurement data.

It supports:

- PyMatching.
- TensorRT CNN predecoder plus PyMatching.
- Conditionally configured NV-Q-LDPC/Relay-BP.

The TensorRT path interoperates with the Ising basis-Z, `XV` surface-code predecoder by aligning the surface-code geometry, measurement order, detector layout, and logical observable convention.

Depends on #637 for surface code orientation. 

### Ising predecoder interoperability

The Ising bundle is an external asset and is not committed. When the TensorRT+Ising path is requested without it, the example prints the exact local-generation recipe and exits.

This PR has no runtime impact.

Justification for > 1k lines PR: the example requires its own `.cpp` file, driver script and helper python script. Additionally, it also carries the unmerged content from #637 .


libs/qec/unittests/realtime/app_examples/surface_code-4-yaml-test.sh
```diff
@@ -0,0 +1,263 @@
...
+export CUDAQ_DEFAULT_SIMULATOR=stim
+export CUDAQ_QEC_REALTIME_MODE=${CUDAQ_QEC_REALTIME_MODE:-inproc_rpc}
+
+P_SPAM=0.01
+
+# Residual logical-error ceiling. A correctly-wired decoder should leave very
+# few residual logical errors; a wired-but-wrong decoder (e.g. one that loads
+# but does not actually correct) leaves many. num_shots/4 is a generous ceiling
+# that still catches a broken decoder.
+MAX_NON_ZERO=$((NUM_SHOTS / 4))
```
kvmto: At p_spam=0.01 the uncorrected logical error rate is about 4% at d=3 and 15% at d=5/T6, both under this 25% ceiling, and the check at line 238 only requires the corrections line to exist, not to be nonzero. A decoder that loads but never corrects passes every positive test. Could we assert the correction count is positive or lower the ceiling below the uncorrected rate?

libs/qec/unittests/realtime/app_examples/CMakeLists.txt
```diff
@@ -428,4 +428,211 @@ add_surface_code_test(
...
+  --yaml ${_sc4_neg_dir}/d3.yml --load_syndrome ${_sc4_neg_dir}/missing.txt
+  --distance 3 --num_rounds 3)
+set_tests_properties(app_examples.surface_code-4-yaml-negative-replay-missing-file
+  PROPERTIES FIXTURES_REQUIRED sc4_replay)
+_sc4_reject(replay-truncated "missing the CORRECTIONS_START/CORRECTIONS_END"
+  --yaml ${_sc4_neg_dir}/d3.yml --load_syndrome ${_sc4_neg_dir}/cap_trunc.txt
+  --distance 3 --num_rounds 3 --num_shots 5)
+set_tests_properties(app_examples.surface_code-4-yaml-negative-replay-truncated
+  PROPERTIES FIXTURES_REQUIRED sc4_replay)
+_sc4_reject(replay-correction-mismatch "replay correction mismatch"
```
kvmto, line 662: These negatives only check that tampered captures fail. If replay itself broke and always returned zero corrections they would still pass. Could we add one positive test that replays the untampered cap.txt and expects "All corrections match"?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+    const std::vector<std::size_t> &cnot_schedZ_flat, bool enqueue_syndromes,
+    bool do_errors_after_non_last_rounds, double p_spam,
+    int logical_qubit_idx) {
+  // Create the logical patch
+  patch logical(data, xstab_anc, zstab_anc);
+  std::vector<cudaq::measure_result> combined_syndrome(xstab_anc.size() +
+                                                       zstab_anc.size());
+
+  // Handle the stabilizer lock-in round (numRounds == 1)
+  if (numRounds == 1) {
```
kvmto, line 538: With `--num_rounds 1` the second call at line 663 re-enters this branch and enqueues an extra syndrome round, so the stream no longer matches the DEM and the decoder silently returns zeros. Should the second call be guarded with `numRounds >= 2`, or should main() reject rounds=1?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+    printf("Error: distance must be an odd integer >= 3 (got %d).\n", distance);
+    return 1;
+  }
+
+  // Set defaults if not specified
+  if (num_rounds == -1)
+    num_rounds = distance;
+  // --yaml is authoritative for the decoder; a co-passed --decoder_type would
+  // be ambiguous, so reject the combination.
+  if (yaml_mode && decoder_type_explicit) {
```
kvmto, line 1413: `--yaml` with `--save_dem` is accepted, the shared dem_filename is overwritten, and the save is silently skipped with exit 0. Should that combination be rejected like `--decoder_type` is here?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+  // Syndrome save/load options
+  bool save_syndrome = false;
+  bool load_syndrome = false;
+  std::string syndrome_filename;
+  bool use_relay_bp = false;
+
+  // Parse the command line arguments
+  for (int i = 1; i < argc; i++) {
+    std::string arg = argv[i];
+    if (arg == "--distance") {
```
kvmto, line 1332: Value-taking flags read argv[i+1] without checking i+1 < argc, so a trailing flag segfaults. Could we add a bounds check with a usage error?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+      }
+#endif
+    }
+    measure_syndrome_round(logical, cnot_schedX_flat, cnot_schedZ_flat,
+                           combined_syndrome);
+    if (enqueue_syndromes)
+      cudaq::qec::decoding::enqueue_syndromes(
+          /*decoder_id=*/logical_qubit_idx, combined_syndrome);
+#if PER_SHOT_DEBUG
+    debug_print_syndromes(syndrome_x_int, syndrome_z_int);
```
kvmto: syndrome_x_int and syndrome_z_int no longer exist here, and line 707 passes a vector to an int64_t parameter, so PER_SHOT_DEBUG=1 fails to compile. Fix or remove the debug path?

kvmto (changes requested):
Nice example and the negative-test harness is careful. Two blocking questions on what the positive tests can actually detect, plus four smaller inline comments.

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+//   - Block 0 (numAncz prep singles): one single-term detector per round-0
+//     Z-ancilla (deterministic after prep0 |0>_L).
+//   - Blocks 1+2 (paired): `pairedRounds` rounds, each pairing this round's
+//     full syndrome [X..., Z...] against the previous round's, via
+//     cudaq::detectors(prev, curr) (numAncx + numAncz detectors per round).
+//   - Block 3 (numAncz boundary): one detector per Z-stabilizer s, XOR-ing the
+//     final data measurements in that stabilizer's support with the
+//     last-round Z-ancilla measurement for that stabilizer.
+__qpu__ void
+dem_gen_circuit(const cudaq::qec::code::one_qubit_encoding &statePrep,
```
eliotheinrich, line 703: This function seems to follow the memory_circuit kernel in PR#636, which merged last week; it looks like this PR was posted after that merge, but would this be a good place to pull that change in?

That kernel and other places (i.e. https://github.com/NVIDIA/cudaqx/blob/main/libs/qec/lib/codes/surface_code_device.cpp#L75) generally format the inter-round detector syndromes as [Z][X], while this PR formats them as [X][Z]. Not sure if there's a reason for this and if so how easy it would be to circumvent. but worth noting.

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+  }
+  h(logicalQubit.ancx);
+  auto results = mz(logicalQubit.ancx);
+  for (std::size_t i = 0; i < logicalQubit.ancx.size(); i++)
+    reset(logicalQubit.ancx[i]);
+  return results;
+}
+
+__qpu__ void
+measure_syndrome_round(cudaq::qec::patch logicalQubit,
```
eliotheinrich, line 511: Why not use the already-existing `surface_code` stabilizer extraction kernels? I.e. https://github.com/NVIDIA/cudaqx/blob/main/libs/qec/lib/codes/surface_code_device.cpp#L52

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+void debug_print_applying_correction(int64_t correction) {
+  printf("Applying correction: %ld\n", correction);
+}
+
+void debug_start_shot() { printf("Starting shot\n"); }
+
+namespace cudaq::qec::qpu {
+
+// Transversal CNOT gate
+__qpu__ void logical_cnot(cudaq::qview<> ctrl_data, cudaq::qview<> tgt_data) {
```
eliotheinrich: It doesn't look like this is called anywhere. Is this dead code?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+  // Put a comma in between each pair of elements
+  for (std::size_t i = 0; i < cnot_schedX_flat.size(); i += 2)
+    printf("%lu %lu, ", cnot_schedX_flat[i], cnot_schedX_flat[i + 1]);
+  printf("\n");
+  printf("cnot_schedZ_flat: ");
+  for (std::size_t i = 0; i < cnot_schedZ_flat.size(); i += 2)
+    printf("%lu %lu, ", cnot_schedZ_flat[i], cnot_schedZ_flat[i + 1]);
+  printf("\n");
+
+  cudaq::noise_model noise;
```
eliotheinrich, line 794: Unused?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+                          logicalQubit.ancz[cnot_sched[i]]);
+  }
+  auto results = mz(logicalQubit.ancz);
+  for (std::size_t i = 0; i < logicalQubit.ancz.size(); i++)
+    reset(logicalQubit.ancz[i]);
+  return results;
+}
+
+__qpu__ std::vector<cudaq::measure_result>
+se_x_ft(cudaq::qec::patch logicalQubit,
```
eliotheinrich: I think this (and the corresponding z-syndrome extraction) name stands for 'stabilizer extraction x fault tolerant', but I'm not sure. Can this be renamed or replaced with the `surface_code` built-in kernels? And is it actually fault-tolerant?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1566 @@
...
+  // paired rounds + 49 data bits), so they must be enqueued to the decoder --
+  // as the last enqueue, completing the
+  // num_rounds*(numAncx+numAncz)+numData-bit buffer (385 at d7/T7) -- BEFORE
+  // asking for corrections. The decoder's boundary detectors XOR these data
+  // bits against the final-round Z-ancilla measurements.
+  //
+  // Note: the result packs numLogical * numData data bits plus the correction
+  // count into one 64-bit value, so num_logical * distance^2 must stay under 64
+  // (enforced in main()). That permits several independent patches at small
+  // distances but only one logical qubit at distance 7 (49 < 64 < 98).
```
eliotheinrich, line 651: Why do this bit-packing? Can we store the syndromes/corrections in a `std::vector` to remove this rather strict size constraint?

eliotheinrich (commented):
Looks nice, but it looks like there are a few spots which are hand-rolling sections that we should be able to support using existing components (and since this is a demo, it seems prudent to use our nice features). I could be missing some context, please let me know if that's the case.

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml-test.sh
melody-ren: Good call. I lowered the ceiling since checking greater than zero correction count might not be enough. Done in 0929acdd2674ab7a00f4db92037efc8c59dfb61c

libs/qec/unittests/realtime/app_examples/CMakeLists.txt
melody-ren: Good catch. Added here in 0929acdd2674ab7a00f4db92037efc8c59dfb61c

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: Ah you're right. That would have been bad to rewrite the yaml under the covers. Fixed in 0929acdd2674ab7a00f4db92037efc8c59dfb61c

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: Yea agree. Added in 0929acdd2674ab7a00f4db92037efc8c59dfb61c

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: oh good catch. These were dead code. Removed in 0929acdd2674ab7a00f4db92037efc8c59dfb61c

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: Good catch! I totally missed the "not-so-happy-path" testing. Thank you. Added in 43726b4ad95bf3888cf3791d20e89814e98dbb38

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: Oh yes, more dead than zombies. Removed in c281aac2394ee7107665dbdc55e513893c0f6cae

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: You're right. Old names were very misleading. Fixed in c281aac2394ee7107665dbdc55e513893c0f6cae

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: Unfortunately this is a restriction imposed by `cudaq::run` for being a top level kernel, since `run` marshals per-shot data as a flat byte span. That means the data type needs to be fixed sized, so using `std::vector` here won't work.

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: This PR was posted before PR636 was merged into upstream's main. I've since pulled in the updates from main. But you're right, the upstream's update does [Z][X]. The reason this PR has a hand rolled version that does [X][Z] is because we want to support the use case of dropping in pre-trained Ising predecoder as one of the options. Ising is following the [X][Z] schedule. In addition, Ising also has prep single/paired/boundary detectors. In this example, we hand craft kernels to match that of Ising's. Same reason goes for the stabilizer extraction kernel discussion below. 

In general, I agree that it would be nice if we could use CUDAQX's library to achieve this. As of now, since we cannot specify the CNOT schedule using the library, we define the schedule explicitly in `get_stab_cnot_schedule`. It would be nice if we can expand the library's feature to support more flexible scheduling settings, though it's up to debate how much flexibility we want to offer.

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: See discussion above

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
melody-ren: It is used when we inject spam errors in the kernel

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1543 @@
...
+    }
+    measure_syndrome_round(logical, cnot_schedX_flat, cnot_schedZ_flat,
+                           combined_syndrome);
+    if (enqueue_syndromes)
+      cudaq::qec::decoding::enqueue_syndromes(
+          /*decoder_id=*/logical_qubit_idx, combined_syndrome);
+  }
+}
+
+__qpu__ std::int64_t
```
tlshannon, line 579: Should this be uint64_t to match ret below?

libs/qec/unittests/realtime/app_examples/surface_code-4-yaml.cpp
```diff
@@ -0,0 +1,1543 @@
...
+  std::ifstream dem_file(dem_filename);
+  std::string dem_str((std::istreambuf_iterator<char>(dem_file)),
+                      std::istreambuf_iterator<char>());
+  auto config =
+      cudaq::qec::decoding::config::multi_decoder_config::from_yaml_str(
+          dem_str);
+  if (numLogical != config.decoders.size()) {
+    printf("ERROR: numLogical [%ld] != config.decoders.size() [%ld]\n",
+           numLogical, config.decoders.size());
+    exit(1);
```
tlshannon, line 336: Do we want to use exit(), or throw exceptions?

melody-ren:
Closing since this PR's commits were carried into #666

PR #653, opened 2026-07-06
Add QEC decoder-server core and CQR adapter

vedika-saravanan:
## Description

Introduces the QEC decoder-server core and CQR DeviceCallService adapter for routing lowered CUDA-Q `cudaq::device_call` decoder RPCs to server-managed decoder sessions.

This PR focuses on the core in-process/CQR server path. Standalone daemon tooling and two-process UDP/RoCE integration tests are intentionally handled separately, for example in the parallel decoder-server integration branch.

### Architecture

```text
CUDA-Q runtime
  └─ DeviceCallService (CQR plugin)
       └─ CqrTransceiver          ← bridges CQR callbacks → ITransceiver
            └─ DecoderServer
                 ├─ RpcDispatcher  ← routes by function_id
                 └─ SessionRegistry
                      └─ DecoderSession (one per decoder)
                           ├─ FIFO worker thread
                           ├─ RoundAccumulator
                           └─ cudaq::qec::decoder instance
```

**Key design points:**

- `DecoderServer` + `SessionRegistry` + `DecoderSession` + `RpcDispatcher` + `ITransceiver` form a core static lib (`cudaq-qec-decoder-server`) with no `CUDAQ_REALTIME` dependency, so the core is testable with `LoopbackTransceiver`.
- `CqrTransceiver` bridges synchronous CUDAQ handler callbacks `(rx_slot, tx_slot, slot_size)` to the async `ITransceiver` model.
- `enqueue_syndromes` is ACKed immediately on the CQR path; decoder execution failures are latched and surfaced on the next `get_corrections`.
- `get_corrections` and `reset_decoder` block until the `DecoderSession` worker responds.
- `SessionRegistry` validates single-transport MVP scope. Mixed transports in one server instance are rejected; concrete CpuRoce/GpuRoce adapter wiring is deferred.
- `RoundAccumulator` supports per-round syndrome assembly and the default pass-through mapping.
- `get_corrections` packs byte-per-bit observable corrections into LSB-first bit-packed wire format with `result_len = ceil(return_size / 8)` exactly.
- Invalid `num_syndromes`, `return_size`, and malformed frame sizes are rejected with `BAD_REQUEST` before unsafe casts or buffer access.
- Clean shutdown is supported through `ITransceiver::shutdown()`, `DecoderServer::stop()`, CQR shutdown export, and loopback shutdown.

**To run the CQR host-dispatch path:**

```bash
export CUDAQ_DEVICE_CALL_CHANNEL=host_dispatch
export CUDAQ_QEC_DECODER_CONFIG=/path/to/decoder_config.yaml
./your_qec_program
```

Alternatively, in-process users may rely on the most recent `configure_decoders()` config before the first decoder RPC.

This PR is over ~1000 lines; the size is justified because it introduces a new subsystem boundary: server core, CQR adapter, wire format handling, session registry, round accumulation, and shutdown/error propagation.

Adds overhead only when the CQR decoder-server path is enabled. The server uses one FIFO worker thread per configured decoder session, copies incoming frame payloads into owned `WorkItem`s, and bit-packs/unpacks syndrome/correction data at the CQR boundary.

### Before requesting review
- [ ] Local test logs reviewed; no unexplained warnings or errors.
- [ ] Full CI has been run.

### Scope and size

### Tests
- [x] New functionality has new tests. - Ben has daemon + two-process end-to-end tests covering this; unit tests tracked as optional follow-up
- [x] Tests fail if the new functionality is broken, not just when it is missing. - the two-process tests exercise the full stack
- [ ] Negative tests added where exceptions are expected.
- [ ] Truth data added where simple `EXPECT_*` / `assert` checks are insufficient.
- [ ] CI runtime impact considered; team notified if significant.

### Documentation
- [ ] User-visible behavior changes have public docs, or a follow-up is tracked.

### Code style

### Dependencies


libs/qec/lib/realtime/realtime_decoding.h
```diff
@@ -54,4 +54,10 @@ void finalize_decoders();
 __attribute__((visibility("default"))) void
 set_syndrome_capture_callback(void (*callback)(const uint8_t *, size_t));
 
+/// @brief The currently registered syndrome-capture callback (nullptr if
+/// none). Served decode paths that bypass host::enqueue_syndromes (the
+/// decoder-server service) use this to keep --save_syndrome working.
+__attribute__((visibility("default"))) void (*get_syndrome_capture_callback())(
```
bmhowe23: It seems like this is a "private" API, not intended to be used by general users. Therefore I'd recommend we prefix the name with an underscore and add that comment. Same with the `set_syndrome_capture_callback` function above (that one must've snuck through).

CMakeLists.txt
```diff
@@ -299,6 +299,7 @@ foreach(lib ${CUDAQX_ENABLE_LIBS})
 endforeach()
 
 
+
```
bmhowe23: This looks like a stray change that should be reverted, just to keep the git history on this file unpolluted.

libs/qec/lib/realtime/CMakeLists.txt
```diff
@@ -308,11 +308,11 @@ else()
   endif()
 endif()
 
+add_subdirectory(decoding-server-cqr)
```
bmhowe23, line 344: Just confirming ... we don't need `CUDAQ_QEC_DEVICE_CALL_SERVICE_AVAILABLE` to build the server? I think that makes sense but I just wanted to double check.

libs/qec/lib/realtime/simulation-cqr/simulation_cqr_device.cpp
```diff
@@ -49,22 +59,66 @@ __qpu__ void
...
+    std::uint8_t value = 0;
+    for (std::size_t bit = 0; bit < 8; ++bit) {
+      std::size_t index = byte * 8 + bit;
+      if (index < num_bits) {
+        bool is_set = syndromes[index];
+        if (is_set)
+          value = value | static_cast<std::uint8_t>(1 << bit);
+      }
+    }
+    packed[byte] = value;
```
bmhowe23, line 79: Just noting: now that https://github.com/NVIDIA/cuda-quantum/pull/4816 has landed, once we incorporate that update from CUDA-Q, we will no longer need to do the bit-packing that this function is doing right now.

libs/qec/lib/realtime/simulation-cqr/simulation_cqr_device.cpp
```diff
@@ -16,6 +16,13 @@
 // fnv1a_32(<callee name>), so these names are what the host-dispatch service
 // (decoding_server_cqr.cpp) registers.
 //
+// The wrappers also perform the spec's on-wire rewrite themselves (until the
+// device_call lowering can): syndrome/correction bits cross the wire
+// bit-packed (LSB-first) in a std::vector<uint8_t> placed last, with the
+// explicit num_syndromes / return_size scalars of the spec's
+// enqueue_syndromes / get_corrections wire layouts. The stdvec<u8>
+// serialization's uint64 byte-count prefix is the spec's ARRAY_UINT8 prefix.
```
bmhowe23, line 24: Just noting this comment (and perhaps whole file) may change for https://github.com/NVIDIA/cuda-quantum/pull/4816 (same comment as above, essentially).

libs/qec/lib/realtime/decoding-server-cqr/CqrTransceiver.h
```diff
@@ -0,0 +1,301 @@
...
+  if (slot_size < sizeof(cudaq::realtime::RPCHeader))
+    return false;
+
+  const auto *cqr_hdr =
+      static_cast<const cudaq::realtime::RPCHeader *>(rx_slot);
+  if (cqr_hdr->magic != cudaq::realtime::RPC_MAGIC_REQUEST)
+    return false;
+
+  const std::size_t arg_len = cqr_hdr->arg_len;
+  const auto *payload = static_cast<const uint8_t *>(rx_slot) +
```
tlshannon, line 233: Should we check arg_len against (slot_size - sizeof(RPCHeader))?

libs/qec/lib/realtime/decoding-server-cqr/CqrTransceiver.h
```diff
@@ -0,0 +1,301 @@
...
+  if (function_id == kEnqueueSyndromesFunctionId) {
+    // Fire-and-forget: hand the frame to the server and ACK immediately
+    // (status OK = ACCEPTED); the dispatcher thread must not park on
+    // decoder execution.
+    {
+      std::lock_guard<std::mutex> lk(mtx_);
+      inbox_.push_back(std::move(frame));
+    }
+    cv_.notify_one();
+    write_ack(tx_slot, rid, hdr->ptp_timestamp);
```
tlshannon: Suggest saving off ptp_timestamp instead of referencing after move().

libs/qec/lib/realtime/decoding-server-cqr/CqrTransceiver.h
vedika-saravanan: Good catch. Fixed in https://github.com/NVIDIA/cudaqx/pull/653/commits/e0cbfee3a48599d6b6d0aa1bd755ef01b8d3fec4.

libs/qec/lib/realtime/decoding-server-cqr/CqrTransceiver.h
vedika-saravanan: You're right. Fixed in https://github.com/NVIDIA/cudaqx/pull/653/commits/e0cbfee3a48599d6b6d0aa1bd755ef01b8d3fec4

libs/qec/lib/realtime/simulation-cqr/simulation_cqr_device.cpp
bmhowe23: I have handled this in the feature branch. Therefore, it doesn't have to be handled in this PR. I'll mark this as resolved.

libs/qec/lib/realtime/simulation-cqr/simulation_cqr_device.cpp
bmhowe23: I have handled this in the feature branch. Therefore, it doesn't have to be handled in this PR. I'll mark this as resolved.

bmhowe23 (commented):
Thanks, Vedika. It's coming together quickly ... thank you. Just a few comments below. Sorry, I thought I had already hit submit on this :(

libs/qec/lib/realtime/CMakeLists.txt
vedika-saravanan: Correct cudaq-qec-decoder-server always builds. CUDAQ_QEC_DEVICE_CALL_SERVICE_AVAILABLE still gates the CQR plugin inside the subdirectory's CMakeLists.

libs/qec/lib/realtime/realtime_decoding.h
vedika-saravanan: Good catch. Fixed in https://github.com/NVIDIA/cudaqx/pull/653/commits/0032c685da740bb0477cc5261ceb09cf9be1e19b.

CMakeLists.txt
vedika-saravanan: Fixed in https://github.com/NVIDIA/cudaqx/pull/653/commits/0032c685da740bb0477cc5261ceb09cf9be1e19b.

bmhowe23 (approved):

PR #654, opened 2026-07-07
Bump CUDA-Q commit and fix for breaking changes

bmhowe23:
Ref: https://github.com/NVIDIA/cuda-quantum/pull/4770

Migrate qec_realtime_session to the new engine API:
  - DEVICE mode builds a cudaq_graph_launch_engine_t from the GRAPH_LAUNCH entries of the function table (which already carry graph_exec / function_id / routing_key). The engine now owns the worker streams, idle mask, inflight-slot tags, and per-worker GraphIOContext array, so the hand-rolled worker/stream/mask/io-ctx setup is gone. The lone captured decoder's pinned mailbox is passed as the engine's external mailbox bank.
  - HOST mode drives cudaq_host_ring_dispatch_loop with a NULL engine (all entries are HOST_CALL), preserving the shared-ring scan behavior.
  - Teardown destroys the engine after the loop thread is joined. Semantics are preserved: DEVICE mode already used auto-sweep, and QEC always uses separate RX/TX ring buffers so the engine takes the same GraphIOContext path as before.

Disable the experimental cudaq-realtime-pipeline target by default (new CUDAQX_QEC_ENABLE_REALTIME_PIPELINE option, OFF). realtime_pipeline.cu depends on host-dispatch API surface PR4770 dropped (skip_stream_sweep, caller-owned worker streams, per-worker pre/post-launch hooks) and needs a separate port. Dependent unit tests auto-skip when the target does not exist.

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "704938b2492793c461f8141f61465db1d71a2f1a"
+    "ref": "c0a143b3116b65c771048fc3f5bb02894e347bba"
   }
 }
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "c0a143b3116b65c771048fc3f5bb02894e347bba"
+    "ref": "8b3f04beefbebd1294209b90c4a3b42b7e03ee03"
   }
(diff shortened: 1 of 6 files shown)


bmhowe23:
All of the non-wheel CI checks passed. Some are still running, and 1 failed, but they all passed when they ran earlier to today with 2981b34a2bed55a6a0d9e1afc229b8741ad09547 (earlier commit in this PR). The failure looks like some sort of dependency download issue unrelated to anything in this PR. To keep the merge train moving, I'm going to merge anyway.

PR #656, opened 2026-07-07
Add boundary-aware overloads for canonicalize_for_rounds and sliding window decoder

### LINKED ISSUE #649: Update sliding window decoder and canonicalize_for_rounds to use new syndrome format
https://github.com/NVIDIA/cudaqx/pull/636 changed the format of detector syndromes output by memory circuits. We need to update the sliding window decoder and canonicalize_for_rounds to use the new format.

eliotheinrich:
## Description

Closes https://github.com/NVIDIA/cudaqx/issues/649

The addition of boundary detectors offsets the syndromes output by memory circuit experiments by introducing `num_boundary_syndromes` at the beginning and end of the syndrome block. The means that `canonicalize_for_rounds` straddles the new boundaries between rounds, leading to error mechanisms being associated to incorrect rounds when the boundary syndromes have a different length than the inter-round syndromes. This is a cosmetic error for PCMs representing entire memory circuits, but is incorrect when followed up with `get_pcm_for_rounds` for the purpose of, i.e., the sliding window decoder. 

This PR adds round-aware overloads to `canonicalize_for_rounds` and the sliding window decoder, bringing them up-to-date with syndrome format.

`canonicalize_for_rounds` has a new overload which accepts `num_boundary_syndromes`, which appropriately treats the first and last `num_boundary_syndromes` columns as distinct rounds for the purpose of round-based topological sorting. 

The sliding window decoder now takes an additional parameter `num_boundary_syndromes` (default 0). The PCM of the decoder is zero-padded to have uniform width `num_syndromes_per_round`. Syndromes passed to the decoder are automatically padded to the uniform length. If `num_boundary_syndromes` is 0, then the boundary syndromes are presumed to be the same width as the interior syndromes (as is the case in X/Z-only experiments). 

New tests:

`test_qec.cpp:PCMUtilsTester.checkGetSortedColumnIndicesWithBoundary`: Verifies the boundary-aware column sort orders
`test_qec.cpp:DetectorErrorModelTest.CanonicalizeBoundaryRejectsWideBoundary`: Assertrs `canonicalize_for_rounds(S, B, ...)` throws when the boundary is wider than the interior (`B > S`).
`test_decoders.cpp::SlidingWindowDecoder.PreparePcmRejectsBadBoundaryLayout`: Confirms the sliding-window decoder throws during construction for inconsistent boundary layouts (`B > S`, or a row count that can't fit `[B | K·S | B]`).
`test_qec_stim.cpp::QECCodeTester.checkDemFromMemoryCircuitShor9`: Checks the Shor [[9,1,3]] full DEM has the expected non-uniform shape, consistent column counts, and correct round ordering.
`test_qec_stim.cpp::QECCodeTester.checkSlidingWindowShor9Boundary`: A single all-rounds window decoding the Shor boundary DEM must match a full decoder for every DEM column.
`test_qec_stim.cpp:QECCodeTester.checkSlidingWindowShor9Streaming`: Streams variable-width detector layers for X(Z) (6(2) at boundaries, 8 interior) through a sliding config (window=2, step=1) and requires results to match a full decoder.
`test_sliding_window.py::test_sliding_window_boundary_layout`: Parametrized over 3 codes × 3 window counts; injects errors over 300 shots and asserts sliding-window observable predictions match a single_error_lut full decoder.

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/libs/qec/include/cudaq/qec/pcm_utils.h
+++ b/libs/qec/include/cudaq/qec/pcm_utils.h
@@ -140,6 +140,18 @@ bool pcm_is_sorted(const cudaqx::tensor<uint8_t> &pcm,
 bool pcm_is_sorted(const std::vector<std::vector<std::uint32_t>> &sparse_pcm,
                    std::uint32_t num_syndromes_per_round = 0);
 
+/// @brief Boundary-aware overload of pcm_is_sorted: the first and last
+/// @p num_boundary_syndromes rows form the boundary rounds and each interior
+/// round spans @p num_syndromes_per_round rows (a [B | K*S | B] layout).
+/// @param sparse_pcm The sparse PCM to check (in the same format as
+/// dense_to_sparse())
+/// @param num_syndromes_per_round The interior-round width.
+/// @param num_boundary_syndromes The width of the first/last boundary rounds.
+/// @return True if the PCM is sorted for this boundary layout, false otherwise.
+bool pcm_is_sorted(const std::vector<std::vector<std::uint32_t>> &sparse_pcm,
+                   std::uint32_t num_syndromes_per_round,
+                   std::uint32_t num_boundary_syndromes);
+
 /// @brief Reorder the columns of a PCM according to the given column order.
 /// Note: this may return a subset of the columns in the original PCM if the
 /// \p column_order does not contain all of the columns in the original PCM.
@@ -182,6 +194,62 @@ simplify_pcm(const cudaqx::tensor<uint8_t> &pcm,
              const std::vector<double> &weights,
              std::uint32_t num_syndromes_per_round = 0);
 
+namespace details {
+
+/// @brief Internal helper (not part of the public API). Maps between detector
+/// rounds and detector rows for a (possibly non-uniform) round layout. The rows
+/// are laid out as [B | S | ... | S | B], where B == num_boundary_syndromes is
+/// the width of the first/last boundary rounds and S == num_syndromes_per_round
+/// is the interior width. B == 0 or B == S is the uniform layout (every round
+/// has width S).
+struct round_layout {
+  std::size_t S = 0;          ///< interior round width
+  std::size_t B = 0;          ///< boundary round width
+  std::size_t num_rows = 0;   ///< total number of detector rows
+  std::size_t num_rounds = 0; ///< number of rounds (detector layers)
+  bool boundary = false;      ///< true iff B is a distinct boundary width
+
+  round_layout() = default;
+  round_layout(std::size_t num_syndromes_per_round,
+               std::size_t num_boundary_syndromes, std::size_t total_rows)
+      : S(num_syndromes_per_round), B(num_boundary_syndromes),
+        num_rows(total_rows) {
+    boundary = (B != 0 && B != S);
+    num_rounds = boundary ? (num_rows - 2 * B) / S + 2 : num_rows / S;
+  }
+
+  /// Global row index at which round @p r begins; round_start(num_rounds) is
+  /// the trailing sentinel (== num_rows).
+  std::size_t round_start(std::size_t r) const {
+    if (!boundary)
+      return r * S;
+    if (r == 0)
+      return 0;
+    if (r >= num_rounds)
+      return num_rows;
+    return B + (r - 1) * S;
+  }
+
+  /// Number of detector rows in round @p r (B for the first/last round, else
+  /// S).
+  std::size_t round_width(std::size_t r) const {
+    return round_start(r + 1) - round_start(r);
+  }
+
+  /// The round that global detector row @p row belongs to.
+  std::size_t row_to_round(std::size_t row) const {
+    if (!boundary)
+      return row / S;
+    if (row < B)
+      return 0;
+    if (row >= num_rows - B)
+      return num_rounds - 1;
+    return 1 + (row - B) / S;
+  }
+};
+
+} // namespace details
+
 /// @brief Get a sub-PCM for a range of rounds. It is recommended (but not
 /// required) that you call sort_pcm_columns() before calling this function.
 /// @param pcm The PCM to get a sub-PCM for.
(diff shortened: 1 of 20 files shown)


libs/qec/include/cudaq/qec/pcm_utils.h
```diff
@@ -172,6 +194,62 @@ simplify_pcm(const cudaqx::tensor<uint8_t> &pcm,
              const std::vector<double> &weights,
              std::uint32_t num_syndromes_per_round = 0);
 
+namespace details {
```
bmhowe23: Can you please move this section to a .cpp file? I don't think it needs to be in a public header file if we are calling it an "internal helper".

eliotheinrich: I can, but I deliberately left it accessible here so that the round layout logic can be shared here and in the sliding window decoder (and perhaps other places that may need a mapping between rounds and detector indices): https://github.com/NVIDIA/cudaqx/pull/656/changes#diff-30af01af4513a257ae4040df674193bc2cce9d3ce09e81adc3493db11a0dba4dR60.

Open to moving it to a .cpp file if we'd rather not expose it.

bmhowe23: In that case, I'd recommend 2 things.
1. Pull them into a private header file (not in the `libs/qec/include` directory because that entire directly should only contain our public interfaces that we are committed to maintaining some form of backwards compatibility on). I'd recommend `libs/qec/lib`.
2. Only put the declarations in header files and put implementations in .cpp files. Implementations in header files can lead to duplicate symbols and general compilation bloat since every target compilation unit must recompile the source code. We should only do that when it is trivial code or needed by for templates.

If you do the above 2 things, I believe the sliding window decoder and other internal code can use it.

eliotheinrich: Thanks; resolved in db3cbe1cbf3a9758644e123169b3f9f8241080e7!

libs/qec/include/cudaq/qec/pcm_utils.h
```diff
@@ -182,20 +204,31 @@ simplify_pcm(const cudaqx::tensor<uint8_t> &pcm,
...
 /// @return A tuple with the new PCM with the columns in the range [start_round,
 /// end_round], the first column included, and the last column included.
 std::tuple<cudaqx::tensor<uint8_t>, std::uint32_t, std::uint32_t>
 get_pcm_for_rounds(const cudaqx::tensor<uint8_t> &pcm,
                    std::uint32_t num_syndromes_per_round,
                    std::uint32_t start_round, std::uint32_t end_round,
                    bool straddle_start_round = false,
-                   bool straddle_end_round = false);
+                   bool straddle_end_round = false,
+                   std::uint32_t num_boundary_syndromes = 0);
```
kvmto, line 217: Should the Python `get_pcm_for_rounds` binding also accept and forward `num_boundary_syndromes`? Otherwise, Python users can perform boundary-aware canonicalization, but `get_pcm_for_rounds` still uses uniform round boundaries and may return the wrong slice. Would it also be worth adding a Python test for that workflow?

libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -199,25 +124,34 @@ sliding_window::sliding_window(const cudaq::qec::sparse_binary_matrix &H,
...
   if (num_syndromes_per_round == 0)
     throw std::invalid_argument("sliding_window constructor: "
                                 "num_syndromes_per_round must be non-zero");
 
-  num_rounds = H.num_rows() / num_syndromes_per_round;
-  num_windows = (num_rounds - window_size) / step_size + 1;
-  num_syndromes_per_window = num_syndromes_per_round * window_size;
+  // Treat a 0 boundary width as the uniform layout (B == S).
+  if (num_boundary_syndromes == 0)
+    num_boundary_syndromes = num_syndromes_per_round;
```
kvmto, line 139: Could this normalization unintentionally reject one-layer uniform PCMs? When `0` is converted to `S`, the validation above requires `num_rows >= 2*S`, whereas a previously valid uniform layout may have `num_rows == S` and `window_size == 1`. Should uniform layouts retain the `num_rows % S == 0` validation?

libs/qec/lib/pcm_utils.cpp
```diff
@@ -130,18 +176,29 @@ void validate_pcm_for_rounds_params(std::uint32_t num_syndromes_per_round,
         "get_pcm_for_rounds: num_syndromes_per_round must be less than the "
         "number of rows in PCM");
   }
-  const std::size_t first_row_to_keep =
-      static_cast<std::size_t>(start_round) * num_syndromes_per_round;
-  const std::size_t last_row_to_keep =
-      static_cast<std::size_t>(end_round + 1) * num_syndromes_per_round - 1;
-  if (first_row_to_keep >= num_rows) {
+  if (num_boundary_syndromes > num_syndromes_per_round) {
```
kvmto, line 174: Could the preceding `num_syndromes_per_round > num_rows` check reject a valid boundary-only layout? For `[B | B]` with no interior layers, `num_rows == 2B`, while `S` may be greater than `2B`. Should validation rely on the `[B | K*S | B]` shape when a boundary width is provided instead of requiring `S <= num_rows`?

kvmto:
The rest seems good! :)

libs/qec/include/cudaq/qec/pcm_utils.h
eliotheinrich: Yes, good point. Addressed in [254bde9](https://github.com/NVIDIA/cudaqx/pull/656/commits/254bde9eacfe202d214f8a3aa9737c5a88f5be96) by adding the Python overload and a test of the Python workflow.

libs/qec/lib/pcm_utils.cpp
eliotheinrich: Good point; I think the subsequent checks are sufficient, and this one was too aggressive. I removed it in [254bde9](https://github.com/NVIDIA/cudaqx/pull/656/commits/254bde9eacfe202d214f8a3aa9737c5a88f5be96)

libs/qec/lib/decoders/sliding_window.cpp
eliotheinrich: That would be a problematic scenario, but I don't think it's possible for our memory circuits at the moment; we require at least a lock-in and read-out round, so the bare minimum number of detectors is 2B (=2S in this case). I've added a new check for this invariant in `validate_inputs` in [254bde9](https://github.com/NVIDIA/cudaqx/pull/656/commits/254bde9eacfe202d214f8a3aa9737c5a88f5be96) so it doesn't do anything unexpected.

bmhowe23 (approved):
Thanks, @eliotheinrich!

PR #658, opened 2026-07-07
Add triangular color code plugin with superdense memory circuit

kvmto:
## Description

Adds the triangular color code to CUDA-QX as a Python QEC code plugin. The registered `stabilizer_round` implements a superdense (paired-ancilla) syndrome-extraction schedule that measures both stabilizer bases of every plaquette in a single depth-8 CNOT pass; the record-conditioned readout byproduct inherent to this scheme is handled through the code-declared inlined-feedback mechanism of #665 rather than by an in-circuit correction gate. The plugin runs end-to-end through the standard framework API (`sample_memory_circuit`, `dem_from_memory_circuit`), and the resulting detector error model is validated mechanism-for-mechanism against an independently constructed reference DEM.

**Depends on #665** (this branch contains its commits; review the top commit here). The feedback-declaration API (`get_inlined_feedback`, `get_observable_inlined_feedback_z/_x`) is defined there.

## Design

- **`ColorCodeGeometry`** — triangular color code with a rectangular-grid embedding (data-qubit grid, algorithmic plaquette generation, syndrome-to-grid mapping, superdense plaquette labeling), ported from an internal reference implementation. Retains the pre-decoder mapping surface (CNN grid embedding, syndrome-to-grid indices) for upcoming decoder work.
- **Schedule builders** (`z_side_data`, `superdense_schedule`) — host-side construction of the per-round CNOT schedule from the geometry: 8 layers over a unified index space, with the ancilla-pair partition of each plaquette's support determined by plaquette weight and color.
- **Registered round as a capturing kernel** — the framework passes only flattened parity matrices to `stabilizer_round`, and the boundary-plaquette partition is color-dependent, i.e. not derivable from parity rows alone. The plugin therefore registers a `@cudaq.kernel` defined in a factory that captures the host-built schedule; the fixed framework signature is unchanged and the captured lists survive the sub-kernel dispatch from the C++ `memory_circuit` (multiple instances remain independent). Records are returned in `[Z][X]` order.
- **Declared inlined feedback** — the byproduct is a linear function of earlier Z-ancilla records, so it folds into detector definitions as data. With `F(S) = {a : |z_side(a) ∩ S| odd}`: the detector matrix has `fb(j,k) = 1` iff `j,k < P` and plaquette `k ∈ F(supp(plaquette j))`; the Z-basis observable row is the indicator of `F(logical-operator support)`; the X basis declares none. Diagonal entries exploit the XOR multiset semantics of `cudaq::detector` (a duplicated record cancels), which is what allows a detector to *exclude* its own previous-round record.
- **Observable representatives** — logical Z is the bottom boundary row (weight d); logical X is the all-data-qubit representative. The asymmetry matches the reference construction, whose Z observable carries herald records while its X observable does not.
- **Order invariant** — `__init__` asserts the minimum data-qubit index is strictly increasing across grid-ordered plaquettes, which guarantees the framework's stabilizer re-sorting is the identity permutation (record k ↔ plaquette k).

## Validation

- **Framework-path E2E**: all four preparations (`prep0/1/p/m`) are noiseless-deterministic with correct logical readout for d=3 at 2 and 4 rounds; DEM extraction succeeds in both bases under two-qubit depolarizing noise; a decode smoke test shows strict logical-error reduction.
- **Golden-oracle DEM equivalence (both bases, d=5, 5 rounds, p=0.01 depolarizing on every CNOT)**: the framework-produced DEM equals a checked-in reference DEM **bit-exactly** — identical mechanism sets (detector supports and observable flips, after a documented block-order permutation) and identical probabilities. Z: 2957 mechanisms, X: 2958.
- **Evidence chain, stated precisely**: the reference DEMs were generated by an independent construction (own kernel with explicit detector rules, not the framework path). That generator byte-reproduces the original reference DEMs under their native convention; the checked-in goldens are the same construction with noise applied to every round. The comparison chain therefore has one inference step: generator ≡ original reference (native convention), framework ≡ generator (all-noisy convention).
- **Convention notes** (deliberate, documented deviations from the original reference DEMs):
  1. *All rounds noisy.* The original reference DEMs use a noiseless final stabilizer round. The framework applies the user's noise model to every gate of a matching type and currently has no mechanism for a noiseless region, so the shipped goldens use the all-rounds-noisy convention (a strict superset of the reference mechanisms: with the final-round convention equalized, the diff is empty). A noiseless-region capability would make the original reference DEMs a drop-in second golden; see follow-ups.
  2. *All-data logical-X representative*, adopted from the reference construction (see Design).

## Notes for reviewers

- The capturing-kernel pattern for a registered operation has no in-tree precedent; it was feasibility-tested through the full C++ dispatch path before adoption, and the alternative (deriving the schedule in-kernel from parity rows) is information-theoretically unavailable for boundary plaquettes.
- Geometry code is a fidelity-preserving port (including some conservatively retained unreachable branches) to keep line-level correspondence with the reference implementation.
- Test suite: 81 items, including matrix truth tables (d=3 exact, d=5 spot checks), schedule/geometry invariants, framework E2E, and the two golden tests with exact-equality assertions and mechanism-count guards.

## Follow-ups (out of scope here)

- A noiseless-region capability in the noise pipeline (enables the perfect-final-round convention and direct comparison against the original reference DEMs).
- Color-decoder (Chromobius) integration; the geometry retains the required coordinate surface.
- Realtime consumers of the declared feedback layout (see #665 follow-ups).


libs/qec/python/tests/data/superdense_golden_d5_r5_allnoisy_X.dem
```diff
@@ -0,0 +1,2965 @@
+# Reference superdense color-code DEM, d=5, 5 rounds, DEPOLARIZE2(0.01)
+# per CX, ALL ROUNDS NOISY (physically stricter convention).
+# Generator: reference superdense construction (superdense_memory kernel,
+# cudaq.dem_from_kernel). The perfect-final-round originals are
+# byte-reproduced by the same generator with its two noise guards restored
+# (R > 1 and r < R - 1); here both are relaxed to p_cx > 0.0 so every
+# stabilizer round is noisy.
```
eliotheinrich, line 9: Would it be feasible to add a reproduction script for this DEM in the header comments here so that if anyone needs to rebuild this golden value down the road are able to? (ditto for the *_Z.dem)

libs/qec/python/cudaq_qec/plugins/codes/color_code.py
```diff
@@ -0,0 +1,1286 @@
...
+                        f"Missing expected {key} position {pos} for w4 L-shape plaq_idx={plaq_idx}, coords={coords}"
+                    )
+                out[key] = int(qid)
+            return out
+
+        raise ValueError(
+            f"Unsupported w4 geometry for plaq_idx={plaq_idx}: distinct_rows={len(rows)}, distinct_cols={len(cols)}, coords={coords}"
+        )
+
+        return out
```
eliotheinrich: This is unreachable; any reason for it to be here?

libs/qec/python/cudaq_qec/plugins/codes/color_code.py
```diff
@@ -0,0 +1,1286 @@
...
+
+    def __init__(self, **kwargs):
+        # The Code binding accepts no constructor arguments; kwargs from
+        # qec.get_code(...) are consumed here on the Python side.
+        qec.Code.__init__(self)
+        if 'distance' not in kwargs:
+            raise RuntimeError(
+                "[color_code] distance not provided. distance must be "
+                "provided via qec.get_code('color_code', distance=d).")
+        self.distance = kwargs['distance']
```
eliotheinrich: Should add some kind of input validation (i.e. make sure that distance is a positive integer)

libs/qec/python/cudaq_qec/plugins/codes/color_code.py
```diff
@@ -0,0 +1,1286 @@
...
+                    else:  # rr % 3 == 2
+                        for cc in range(cols):
+                            edges.append((index_ancilla, data_qubit_index))
+                            index_ancilla += 1
+                            if cc == cols - 1:
+                                data_qubit_index += 3
+                            else:
+                                data_qubit_index += 1
+
+            layers.append(sorted(edges))
```
eliotheinrich: Two comments: 
1.) Nesting conditionals inside a for-loop doesn't make much sense since this is just a sequence; just push each conditional up one layer of scope
2.) The final three layers are a mirror image of the three layers before them under exchange of target and control qubits. Can make this explicit in the code

libs/qec/python/tests/test_color_code.py
```diff
@@ -0,0 +1,846 @@
...
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("prep,basis,obs_getter,expected", [
+    (qec.operation.prep0, 'z', 'get_observables_z', 0),
+    (qec.operation.prep1, 'z', 'get_observables_z', 1),
+    (qec.operation.prepp, 'x', 'get_observables_x', 0),
+    (qec.operation.prepm, 'x', 'get_observables_x', 1),
+])
+@pytest.mark.parametrize("nRounds", [2, 4])
```
eliotheinrich: nRounds == 1 is a bit of an edge case since no inter-round detectors are formed. Easy and IMO worthwhile to add to the parametrized test here

libs/qec/python/cudaq_qec/plugins/codes/color_code.py
```diff
@@ -0,0 +1,1286 @@
...
+    X byproduct on the Z-ancilla-coupled data qubits; the correction is absorbed
+    downstream by the code's inlined-feedback declarations rather than by a
+    coherent in-round CX.
+    """
+
+    @cudaq.kernel
+    def _round(logicalQubit: patch, x_stabilizers: list[int],
+               z_stabilizers: list[int]) -> list[cudaq.measure_handle]:
+        # x_stabilizers / z_stabilizers arrive from the framework but are
+        # unused: the captured schedule alone drives the gate sequence.
```
eliotheinrich, line 1006: It might be reasonable to add some checks that the *_stabilizers passed correspond correctly to the stabilizers of the code, even if they aren't used. If bad stabilizers are passed to `_round`, then something has gone wrong, and we should probably throw.

eliotheinrich (commented):
Looks great, Kevin! I left some suggestions, the only thing I would consider blocking is the input validation on the distance and some hint for reproducing the golden values in the .dem files if possible.

jolle-ag:
Looks great! I would just recommend to test it (at least some spot tests) at `d=7` since there are lattice configurations that won't have been previously encountered at `d=3` and `d=5`. Mostly the "deep bulk" where multiple weight-6 plaquettes are neighboring

PR #661, opened 2026-07-08
Fix issue#646

### LINKED ISSUE #646: `test_code.py::test_python_code` disabled by #636
https://github.com/NVIDIA/cudaqx/pull/636/ adds changes to detectors, which revealed an issue with the Python/C++ interoperability. The bug can be recreated (after cloning that PR) with

```
import cudaq, cudaq_qec as qec
from cudaq_qec import patch


# Bell state |00>+|11> is stabilized by both XX and ZZ, so every stabilizer
# measurement is deterministic round-to-round (no error is expected).
@cudaq.kernel
def prep0(p: patch):
    h(p.data[0])
    x.ctrl(p.data[0], p.data[1])


# One stabilizer round, run as a sub-kernel of the C++ `memory_circuit`.
@cudaq.kernel
def stab(p: patch, xs: list[int], zs: list[int]) -> list[cudaq.measure_handle]:
    h(p.ancx[0]); x.ctrl(p.ancx[0], p.data[0]); x.ctrl(p.ancx[0], p.data[1]); h(p.ancx[0])
    x.ctrl(p.data[0], p.ancz[0]); x.ctrl(p.data[1], p.ancz[0])
    results = mz([*p.ancz, *p.ancx])
    reset(p.ancx[0]); reset(p.ancz[0])
    return results


@qec.code('mini')
class Mini:
    def __init__(self, **kw):
        qec.Code.__init__(self, **kw)
        self.stabilizers = [cudaq.SpinOperator.from_word(w) for w in ["XX", "ZZ"]]
        self.pauli_observables = [cudaq.SpinOperator.from_word("ZZ")]
        self.operation_encodings = {qec.operation.prep0: prep0,
                                    qec.operation.stabilizer_round: stab}
    def get_num_data_qubits(self): return 2
    def get_num_ancilla_qubits(self): return 2
    def get_num_ancilla_x_qubits(self): return 1
    def get_num_ancilla_z_qubits(self): return 1
    def get_num_x_stabilizers(self): return 1
    def get_num_z_stabilizers(self): return 1


# All stabilizers are deterministic, so this should succeed -- but it raises a  "ValueError: non-deterministic detectors".
qec.sample_memory_circuit(qec.get_code('mini'), numShots=2, numRounds=3)
```

This is not an issue if the above code is recreated in pure C++. I don't fully understand why this happens, but I suspect it will require a change to CUDA-Q or a follow-up. In the meantime, I have disabled the test., and will add it back once I understand the issue more fully.

khalatepradnya:
## Description

Re-enable test since CUDA-Q PR https://github.com/NVIDIA/cuda-quantum/pull/4861 addressed the underlying issue with the measurement handle lifetime.

N/A

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "3b20c9669c10d6e9bc0fe1758757f5240c588faa"
+    "ref": "5439aff2c851a1b06ced9f24cdec7b5bfc3273ac"
   }
 }
--- a/libs/qec/python/tests/test_code.py
+++ b/libs/qec/python/tests/test_code.py
@@ -112,14 +112,13 @@ def test_noisy_simulation():
     cudaq.reset_target()
 
 
-@pytest.mark.skip
 def test_python_code():
     steane = qec.get_code("py-steane-example")
     syndromes, dataResults = qec.sample_memory_circuit(steane,
                                                        numShots=10,
                                                        numRounds=4)
     assert isinstance(syndromes, np.ndarray)
-    assert syndromes.shape == (10, 40)
+    assert syndromes.shape == (10, 24)
     print(syndromes)
     assert not np.any(syndromes)
 


eliotheinrich (commented):
LGTM, thanks for the fix! The .cudaq-version is getting bumped by https://github.com/NVIDIA/cudaqx/pull/654, let's let that PR merge first and be responsible for the version update.

.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "3b20c9669c10d6e9bc0fe1758757f5240c588faa"
+    "ref": "5439aff2c851a1b06ced9f24cdec7b5bfc3273ac"
```
bmhowe23, line 4: I just included this CUDA-Q ref in #654, which I'd like to merge today. After it lands, you should be able back out any changes to this file. Thanks for the CUDA-Q fix and CUDA-Q QEC PR!

bmhowe23: @khalatepradnya fyi, #654 just merged to main

bmhowe23 (approved):
Thanks, @khalatepradnya!

PR #666, opened 2026-07-09
Standalone Realtime QEC Decoding Server

bmhowe23:
## Description

Builds on #653 

This PR builds out the **standalone realtime decoding server** built on cudaq-realtime: a separate decoder process reached from the QPU-side program over a CUDA-Q `device_call` transport, with both the decoder selection and the transport being *configuration, not code*. It also adds the parallel/heterogeneous decoder support underneath it and the two-process and FPGA-in-the-loop tests that exercise the whole path.

### What's new

**Decoding server (`libs/qec/tools/decoding-server/`)**
- `decoding_server`: a standalone decoding-server process — the service end of a CUDA-Q `device_call` transport, decoding on the CPU with whatever decoder(s) a YAML config selects (`--config=<yaml>`, parsed by `multi_decoder_config`). The function table (`enqueue_syndromes` / `get_corrections` / `reset_decoder`) comes from the `decoding-server-cqr` service plugin regardless of transport or decoder.
- Transport is selected at launch: `--transport=udp` (loopback ring transceiver, runs anywhere) or `--transport=cpu_roce` (RDMA, requires a ConnectX NIC). For `cpu_roce`, `--qp_config=rendezvous|hsb_fpga` selects TCP QP/rkey rendezvous vs. the Holoscan-Sensor-Bridge FPGA handshake (the server performs no FPGA control-plane traffic). (This is initial and subject to change.)
- Installed like the libraries: the binary and its example YAML configs go to `bin/` under a new `qec-tools` install component (with a proper `INSTALL_RPATH` so the installed copy finds `libcudaq-qec` and the decoder plugins). Deliberately excluded from the Python wheel, since the cudaq wheel does not ship the CUDA-Q realtime libs the server needs.

**Core layer (`decoding-server-cqr`)** (Built on #653)
- New `ITransceiver` abstraction with three implementations: the existing loopback, a new `CpuRoceTransceiver` (RDMA over the CPU HOST_CALL path), and a new `GpuRoceTransceiver` — a GPU RoCE path that wires the Hololink Sensor Bridge receiver into CUDA-Q's device-graph dispatcher (env-var configured, guarded by `CUDAQ_GPU_ROCE_AVAILABLE` CMake detection with graceful fallback when no GPU/NIC is present).
- Wire-format and RPC protocol fixes in `RpcWireFormat.h` / `decoder_rpc_ids.h` to stay byte-exact with what CUDA-Q kernel lowering actually emits.

**Parallel / heterogeneous decoders**
- `SessionRegistry` + `DecodingSession` rework so multiple decoders run concurrently on independent worker threads, including *different decoder types* served by one server; lifecycle races and hardening gaps found in the PR #653 review are fixed.
- `create_realtime_decoder()` centralizes realtime decoder construction (ID + O/D matrix wiring) so the in-process path, the server, and tests all build decoders the same way; decoder construction/initialization failures (e.g. TensorRT) now propagate instead of being swallowed.

**Tests**
- `test_decoding_server`: two-process test — this process is caller-only (no decoder configured at all), the server is spawned as a subprocess, and the server's dispatch count proves the `device_call`s crossed the process boundary. Parameterized over `udp` and `cpu_roce`/SoftRoCE transports.
- `test_decoding_server_core` / `test_decoders_yaml`: unit coverage for the server core and YAML decoder configuration.
- `concurrency_test_decoder`: an instrumented decoder plugin proving three decoder workers genuinely overlap in time.
- `app_examples`: `surface_code-1` (minimal realtime decode app + two-process CQR variant) and `surface_code-4-yaml` (multi-patch surface code, one decoder type per patch, exercising heterogeneous decoders through the server), plus `gen_dsparse_from_memory_circuit.py` for generating the sparse detector matrices.
- `hsb_fpga_decoding_server_test.sh`: end-to-end orchestration over Holoscan-Sensor-Bridge RDMA — FPGA (or emulator) plays back syndrome frames into the server's RDMA ring and verifies the RPC responses; works with any CPU decoder the YAML selects.

### Size note
This is well over the ~1000-line guideline (~8,100 insertions), but the bulk is test infrastructure (`app_examples`, the two-process tests, and the HSB orchestration script account for roughly 5,500 lines); the server, transceivers, and server-core changes are the reviewable core.

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
- [ ] ~~Public-facing APIs have Doxygen docs.~~ Deferred
- [ ] ~~User-visible behavior changes have public docs, or a follow-up is tracked.~~ Deferred

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


BRANCH_SUMMARY.txt
bmhowe23, line 1: Noting so I can't forget. This needs to be deleted before merging.

PR #669, opened 2026-07-10
Include cmake/ in cudaq realtime sparse checkout; fix test

bmhowe23:
cuda-quantum commit 604d3034 ([#4870](https://github.com/NVIDIA/cuda-quantum/pull/4870)) made realtime/unittests include cmake/modules/CUDAQGtestDiscovery.cmake from the repo root, which the realtime-only sparse checkout in CI does not fetch, breaking configure with "include could not find requested file". Fetch cmake/ alongside realtime/ in both places that clone the realtime sources.

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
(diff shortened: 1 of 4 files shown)


bmhowe23:
Checked this out private CI runs. Merging now.

PR #670, opened 2026-07-10
qec/realtime: Relay BP (nv-qldpc) gpu_roce profile for the HSB decoding-server test

cketcham2333:
## Description

Adds a **Relay BP (nv-qldpc) profile** to the HSB decoding-server test, driving the decoder
through the server's `gpu_roce` transport — the **self-relaunching device-graph scheduler**
(`GpuRoceTransceiver`): `enqueue_syndromes` / `get_corrections` / `reset_decoder` execute as
`DEVICE_CALL`s on the GPU, and the captured RelayBP decode graph fires **device-side**
(fire-and-forget + tail self-relaunch) when a window completes. This replaces
`hololink_qldpc_graph_decoder_bridge` for the decoding-server flow; the old bridge test is
untouched. The pymatching/cpu_roce profile is unchanged and re-verified.

```
hololink_fpga_syndrome_playback (UNCHANGED —      decoding_server --transport=gpu_roce
 sole FPGA control-plane writer)                    HOLOLINK_* env → GpuRoceTransceiver
  programs SIF w/ server QP/RKey/addr               DEVICE_CALL enqueue/get/reset on GPU,
  writes per-round RPC frames to FPGA BRAM          RelayBP decode graph fired device-side,
  arms ILA, verifies captured RPCResponses          dispatch graph tail-self-relaunches
       FPGA/emulator ──RDMA WRITE──► DOCA GPU rx ring ; server ──RDMA WRITE──► FPGA SIF TX
```

## Changes

**`hsb_fpga_decoding_server_test.sh`**
- `--transport cpu_roce|gpu_roce`, defaulted from the decoder profile (`pymatching` →
  `cpu_roce`, `nv-qldpc-decoder` → `gpu_roce`); new `--gpu`, `--proprietary-archive`,
  `--nv-qldpc-plugin` flags.
- Data generation (fresh each run via `surface_code-4-yaml`) adds `--use-relay-bp` for the
  nv-qldpc profile and **injects `transport: gpu_roce`** into the generated YAML —
  `DecoderServer` selects its transceiver from the per-decoder `transport:` key, the generator
  omits non-default optionals, and the default (`cpu_roce`) resolves to the not-yet-implemented
  `CpuRoceTransceiver` stub.
- gpu_roce server launch: Hololink parameters via `HOLOLINK_*` env (the server's gpu_roce mode
  ignores the cpu_roce CLI flags); remote QP converted to decimal
  (`GpuRoceConfig::from_env` parses base-10 only); `CUDA_MODULE_LOADING=EAGER` as with the old
  bridge launcher; readiness keyed on `QEC_DECODING_SERVER_READY gpu_roce` (the transceiver
  prints the QP/RKey/Buffer handshake before that line, without the bridge banner).
- Build phase: builds HSB `gpu_roce_transceiver` + cuda-quantum `cudaq-realtime-bridge-hololink`,
  wires `CUDAQ_QEC_REALTIME_CUDEVICE_PROPRIETARY_ARCHIVE` / `GPU_ROCE_TRANSCEIVER_LIB` /
  `CUDAQ_REALTIME_BRIDGE_HOLOLINK_LIBRARY` into the cudaqx configure, clears the CMake cache
  (find_library NOTFOUND staleness), and symlinks the nv-qldpc plugin into
  `build/lib/decoder-plugins` (also opportunistically on `--build`-less runs). All gated on the
  proprietary archive's presence, so pymatching-only rigs build exactly as before.

**`decoding-server-cqr/CMakeLists.txt` — bug fix (separable)**
`CUDAQ_GPU_ROCE_AVAILABLE` was set as `CACHE INTERNAL` on success, but the plain
`set(... FALSE)` at the top of the file shadows the cache under CMP0126 (NEW), so the
compile-definition gate for `cudaq-qec-decoding-server` never fired: the library silently built
**without** gpu_roce (`make_transport` threw "CUDAQ_GPU_ROCE_AVAILABLE is not set" at runtime)
while the sibling `decoding_server` target — which reads the cache — reported
"gpu_roce transport enabled". Fixed by setting the normal variable alongside the cache entry.

### DIFF
--- a/libs/qec/lib/realtime/decoding-server-cqr/CMakeLists.txt
+++ b/libs/qec/lib/realtime/decoding-server-cqr/CMakeLists.txt
@@ -98,14 +98,14 @@ if(HOLOSCAN_SENSOR_BRIDGE_BUILD_DIR AND CUDAQ_REALTIME_INCLUDE_DIR)
       AND DOCA_COMMON_LIB AND CUDAQ_REALTIME_BRIDGE_HOLOLINK_LIBRARY
       AND CUDAQ_REALTIME_DISPATCH_LIBRARY AND CUDAQ_REALTIME_HOST_DISPATCH_LIBRARY)
     set(CUDAQ_GPU_ROCE_AVAILABLE TRUE CACHE INTERNAL
-      "GPU RoCE transport compiled into cudaq-qec-decoding-server")
+      "GPU RoCE transport component cudaq-qec-decoding-server-gpuroce available")
     # Under CMP0126 (NEW) the CACHE set above does NOT update the normal
     # variable initialized FALSE at the top of this file, and the normal
-    # variable shadows the cache in this scope -- so the compile-definition
-    # gate below would never fire.  Set the normal variable explicitly.
+    # variable shadows the cache in this scope -- so the component gate
+    # below would never fire.  Set the normal variable explicitly.
     set(CUDAQ_GPU_ROCE_AVAILABLE TRUE)
     message(STATUS
-      "cudaq-qec-decoding-server: GPU RoCE ENABLED (core=${HOLOLINK_CORE_LIB})")
+      "cudaq-qec-decoding-server-gpuroce: GPU RoCE ENABLED (core=${HOLOLINK_CORE_LIB})")
   else()
     message(STATUS
       "cudaq-qec-decoding-server: GPU RoCE DISABLED "
@@ -125,6 +125,12 @@ message(STATUS
 # ---------------------------------------------------------------------------
 # Core decoding server (always built): transport abstraction, session registry,
 # per-decoder worker threads, RPC dispatcher.  No dependency on CUDAQ_REALTIME.
+#
+# Deliberately excludes GpuRoceTransceiver: consumers of this library (unit
+# tests, the CQR plugin) must stay loadable on driverless machines, and the
+# GPU RoCE deps (DOCA, the hololink bridge) require libcuda.so.1 at load
+# time.  gpu_roce lives in the cudaq-qec-decoding-server-gpuroce component
+# below, reached through a weak factory symbol (see DecodingServer.cpp).
 # ---------------------------------------------------------------------------
 add_library(cudaq-qec-decoding-server STATIC
   RoundAccumulator.cpp
@@ -133,7 +139,6 @@ add_library(cudaq-qec-decoding-server STATIC
   RpcDispatcher.cpp
   DecodingServer.cpp
   CpuRoceTransceiver.cpp
-  GpuRoceTransceiver.cpp
 )
 
 target_compile_features(cudaq-qec-decoding-server PUBLIC cxx_std_20)
@@ -156,31 +161,6 @@ target_link_libraries(cudaq-qec-decoding-server
     cudaq-qec-realtime-decoding
 )
 
-if(CUDAQ_GPU_ROCE_AVAILABLE)
-  target_compile_definitions(cudaq-qec-decoding-server PRIVATE CUDAQ_GPU_ROCE_AVAILABLE)
-  target_include_directories(cudaq-qec-decoding-server PRIVATE
-    "${CUDAQ_REALTIME_INCLUDE_DIR}"
-    "${_doca_root}/include"
-  )
-  target_link_libraries(cudaq-qec-decoding-server PRIVATE
-    "${CUDAQ_REALTIME_BRIDGE_HOLOLINK_LIBRARY}"
-    "${GPU_ROCE_TRANSCEIVER_LIB}"
-    "${BASE_RECEIVER_OP_LIB}"
-    "${HOLOLINK_CORE_LIB}"
-    "${HOLOLINK_COMMON_LIB}"
-    "${DOCA_VERBS_LIB}"
-    "${DOCA_GPUNETIO_LIB}"
-    "${DOCA_COMMON_LIB}"
-    $<$<BOOL:${IBVERBS_LIB}>:${IBVERBS_LIB}>
-    CUDA::cudart
-    # libdoca_gpunetio.so and libcudaq-realtime-bridge-hololink.so reference
-    # the CUDA Driver API (cu*); consumers of this static lib inherit those
-    # deps, so link the driver here.  CUDA::cuda_driver resolves to the
-    # toolkit's stubs in driverless build environments (CI containers).
-    CUDA::cuda_driver
-  )
-endif()
-
 set_target_properties(cudaq-qec-decoding-server PROPERTIES
   ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
   POSITION_INDEPENDENT_CODE ON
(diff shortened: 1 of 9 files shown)


libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
```diff
@@ -82,6 +82,22 @@ GEN_ROUNDS=4
...
+TRANSPORT=""
+# GPU for the gpu_roce scheduler + decode graph (HOLOLINK_GPU_ID).
+GPU_ID=0
+
+# gpu_roce build artifacts from the proprietary cuda-qx tree: the cudevice
+# archive (DEVICE_CALL handlers + populate shims, WHOLE_ARCHIVE-linked and
+# device-linked into the server) and the nv-qldpc plugin .so (dlopen'd;
+# capture_decode_graph builds the RelayBP decode graph).
+PROPRIETARY_ARCHIVE="/workspaces/cuda-qx/build/lib/libcudaq-qec-realtime-cudevice-proprietary.a"
+NV_QLDPC_PLUGIN="/workspaces/cuda-qx/build/lib/decoder-plugins/libcudaq-qec-nv-qldpc-decoder.so"
```
bmhowe23: The .a file and .so files should be installed/delivered artifacts. These bash env vars should not be pointing to a private build area. If the plan is to put these in the Docker container as example tests, the paths should be to the installed area. @anjbur for viz.

libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
```diff
@@ -624,23 +668,34 @@ do_build() {
         "$hsb_cmake" "${hsb_common_args[@]}" 2>&1 | tail -5
     fi
 
+    # hololink_core is enough for playback + the emulator; the gpu_roce
+    # server additionally needs the HSB gpu_roce_transceiver archive.
+    local hsb_targets=(hololink_core)
+    local hsb_gpu_roce_lib="${hsb_build}/src/hololink/operators/gpu_roce_transceiver/libgpu_roce_transceiver.a"
```
bmhowe23: @anjbur for viz. @cketcham2333 does this mean that the decoding server docker image needs to have the HSB source code? Is the .a in that source code? I'm just trying to think about how this is productized. (This comment assumes that we want this script to be included as some sort of a self-test include in the container.)

libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
```diff
@@ -649,6 +704,28 @@ do_build() {
...
+        )
+        _info "gpu_roce server support: using $PROPRIETARY_ARCHIVE"
+    else
+        _info "gpu_roce server support: DISABLED (no proprietary archive at" \
+              "$PROPRIETARY_ARCHIVE)"
+    fi
+
+    # Clear stale cmake cache entries (find_library caches NOTFOUND
+    # permanently, and CUDAQ_GPU_ROCE_AVAILABLE is CACHE INTERNAL).
+    rm -f "$cudaqx_build/CMakeCache.txt"
```
bmhowe23: Is this still needed? This seems a bit kludgy.

libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
```diff
@@ -649,6 +704,28 @@ do_build() {
...
+
```
bmhowe23: The `cmake` and build of cudaqx itself is making me reconsider whether this script is the right thing to put in the Docker container. That is - ideally we wouldn't have to have a cudaqx build script running in an already prebuilt Docker container for the decoding server.

bmhowe23 (commented):
Thanks, Chuck. Sharing some comments now ... I'm not sure to what extent we want to handle my comments in this PR vs a follow-up PR. We can discuss offline.

libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
cketcham2333: Resolved in commit 3d4354779d55d8c6f679113077c01c7b908407dd. The proprietary-archive path is **gone from the script entirely** — it was only ever a build-time input to the gpu_roce daemon, and the script no longer builds the daemon (it consumes it prebuilt).

libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
cketcham2333: Resolved in commit 3d4354779d55d8c6f679113077c01c7b908407dd. No — and stronger now: the script **no longer builds HSB or the decoder server at all**. It builds only the `surface_code-4` generator (the one artifact not shipped in the image) and consumes the decoder server, HSB / cudaq-realtime libs, and plugins **prebuilt/installed**. So the container needs no HSB source

libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
cketcham2333: Resolved in commit 3d4354779d55d8c6f679113077c01c7b908407dd. Gone. That cache clear only existed to reconfigure the **gpu_roce daemon build** between runs (find_library NOTFOUND + a CACHE INTERNAL var). Since the script no longer builds the daemon, there's no reconfigure and no cache surgery — `do_build` is now just `cmake --build --target surface_code-4-yaml` against the already-configured tree.

libs/qec/unittests/utils/hsb_fpga_decoding_server_test.sh
cketcham2333: Resolved in commit 3d4354779d55d8c6f679113077c01c7b908407dd.

libs/qec/lib/realtime/decoding-server-cqr/CMakeLists.txt
```diff
@@ -185,6 +175,77 @@ install(TARGETS cudaq-qec-decoding-server
...
+# Optional GPU RoCE component: GpuRoceTransceiver plus the strong definition
+# (GpuRoceFactory.cpp) of the factory that DecodingServer.cpp references
+# weakly.  Kept out of the core library so core consumers carry no DOCA /
+# Hololink / CUDA-driver runtime dependencies -- those .so's need
+# libcuda.so.1 at load time, which driverless machines (CI containers) lack.
+#
+# Consumers must link this WHOLE_ARCHIVE: the only reference to the factory
+# is weak, which does not pull archive members on its own.
+# ---------------------------------------------------------------------------
+if(CUDAQ_GPU_ROCE_AVAILABLE)
```
bmhowe23, line 188: Just noting (and we can discuss more offline), I'd like to move away from the decoding server having any awareness of what type of transport is being used, so we should have a follow-up pass where we remove as much "GPU RoCE" stuff as we can for this release. I'm not requesting any changes right now, so I'll mark this as resolved.

bmhowe23 (approved):

libs/qec/lib/realtime/decoding-server-cqr/CMakeLists.txt
cketcham2333: Noted.

PR #671, opened 2026-07-10
Added bundled decoder_context struct for measurement extraction

eliotheinrich:
## Description

In order to process enqueued syndromes, the decoder needs access to a measurement map so it knows how to form detectors. Right now, this comes from `generate_timelike_sparse_detector_matrix`, which does not include boundary detectors. 

Rather than refactor this function (which would require it to be aware of the stabilizer structure of the code), I include the `m2d` map calculated by `dem_from_kernel`, and bundle it into a new struct `decoder_context`, and provide a new helper function `decoder_context_from_memory_circuit` with a similar signature to `dem_from_memory_circuit`, which it essentially supersedes. This new function returns the DEM as well as the measurement map needed by the decoder. The `decoder_context` struct also provides helper methods for extracting the X/Z components, alleviating the need for helper functions (in the style of `*_dem_from_memory_circuit`) as well as allowing a single circuit analysis to provide both the X and Z decoder contexts. 

The PR also applies some changes to the `real_time_complete` documentation examples to use this function.

New tests:
  `test_dem.py::test_decoder_context_from_memory_circuit_requires_noise_model` - asserts a throw on the `decoder_context_from_memory_circuit` if no noise model is provided
  `test_dem.py::test_decoder_context_and_components` - asserts that `decoder_context_from_memory_circuit` matches output of superceded `*_dem_from_memory_circuit`
  `test_dem.py::test_decoder_context_d_sparse_layout()` - asserts that `.d_sparse()` on `decoder_context` flattens m2d map into correct -1-terminated sparse vector
  `test_dem.py::test_decoder_context_single_type_code_empty_component` - check that empty component (i.e. x-component of a repetition code) is consistent
  `test_qec_stim.cpp::checkTealtimeDecodeFromMemoryCircuit` - end-to-end check of realtime API using decoder_context
  `test_qec_stim.cpp::checkDecoderContextAndComponents` - Verify DEM output by `decoder_context_from_memory_circuit` matches superseded `*_dem_from_memory_circuit`, and that extracting `x/z` components is idempotent and consistent.

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


libs/qec/lib/experiments.cpp
```diff
@@ -432,17 +432,15 @@ sample_memory_circuit(const code &code, std::size_t numShots,
...
-                        bool decompose_errors) {
-  if (!keep_x_stabilizers && !keep_z_stabilizers)
-    throw std::runtime_error("dem_from_memory_circuit error - no stabilizers "
-                             "to keep.");
+/// @brief Fully characterize a memory-circuit experiment, returning the
+/// decoder_context. This is the shared driver behind all the public entry
+/// points: `dem_from_memory_circuit` keeps just the `.dem`, and the `x_`/`z_`
+/// variants slice the result via `x_component()`/`z_component()`.
+decoder_context decoder_context_from_memory_circuit(const code &code,
+                                                    operation statePrep,
```
melody-ren, line 517: This function currently serves as both the public factory and the shared internal driver, which means every call constructs the full two-basis result. That can be quite expensive as DEMs get bigger and in some sense, is a performance regression as well. Ideally, we want to only do work that's strictly needed and defer others until called for. I suggest that we restore a private selection-aware driver—keep_x/keep_z or equivalent—similar to the original implementation.

For example: the internal path should run `dem_from_kernel` once, select the requested DEM and m2d rows using the layout information already available here, canonicalize that selected model once, and only then construct the returned context. The public full/X/Z functions can be thin selectors over that driver.

libs/qec/lib/experiments.cpp
```diff
@@ -481,66 +479,141 @@ dem_from_memory_circuit(const code &code, operation statePrep,
...
+    out.num_x_stabilizers = 0;
+  if (!keep_z)
+    out.num_z_stabilizers = 0;
+
+  // A sliced context keeps a single stabilizer type, so the DEM is uniform
+  // (every detector layer has the same width) and canonicalizes per round.
+  const std::size_t numReturnSynPerRound =
+      (keep_z ? self.num_z_stabilizers : 0) +
+      (keep_x ? self.num_x_stabilizers : 0);
+  out.dem.canonicalize_for_rounds(numReturnSynPerRound,
```
melody-ren: This is the expensive part that I discussed in a previous comment. If I understand it correctly, the full DEM was already canonicalized above in decoder_context_from_memory_circuit(). For an X/Z request, a cheaper route is to select the requested detector rows from the uncanonicalized DEM and canonicalize the selected model exactly once.

libs/qec/lib/experiments.cpp
```diff
@@ -481,66 +479,141 @@ dem_from_memory_circuit(const code &code, operation statePrep,
...
+decoder_context slice_full(const decoder_context &self, bool keep_x,
+                           bool keep_z) {
+  const std::size_t numDetectors = self.dem.detector_error_matrix.shape()[0];
+  const std::size_t numErrorMechanisms =
+      self.dem.detector_error_matrix.shape()[1];
+  const auto ranges = details::stabilizer_detector_ranges(
+      self.num_rounds, self.num_z_stabilizers, self.num_x_stabilizers,
+      self.fixed_basis_is_z, keep_x, keep_z, numDetectors);
+
+  decoder_context out = self;
```
melody-ren: We're doing a deep copy here which can be expensive. I think you can build the output directly from the selected rows of the internal raw analysis.

libs/qec/include/cudaq/qec/experiments.h
```diff
@@ -159,6 +159,47 @@ std::tuple<cudaqx::tensor<uint8_t>, cudaqx::tensor<uint8_t>>
...
+  /// indices in order, followed by `-1`).
+  std::vector<std::int64_t> d_sparse() const;
+
+  /// @brief Restrict this decoder context to just the X-stabilizer
+  /// detectors
+  decoder_context x_component() const;
+
+  /// @brief Restrict this decoder context to just the Z-stabilizer
+  /// detectors
+  decoder_context z_component() const;
```
melody-ren: The struct is a public facing API. These fields are, for the most part, used for internal analysis (e.g., the describe how the context was constructed and exist so the returned object can slice and reinterpret itself later). I think the layout information should exist in the private/internal helpers. The public facing struct should carry just the finalized result, DEM and measurement maps.

libs/qec/lib/experiments.cpp
```diff
@@ -549,21 +622,21 @@ detector_error_model x_dem_from_memory_circuit(const code &code,
                                                std::size_t numRounds,
                                                cudaq::noise_model &noise,
                                                bool decompose_errors) {
-  return details::dem_from_memory_circuit(code, statePrep, numRounds, noise,
-                                          /*keep_x_stabilizers=*/true,
-                                          /*keep_z_stabilizers=*/false,
-                                          decompose_errors);
+  return decoder_context_from_memory_circuit(code, statePrep, numRounds, noise,
```
melody-ren, line 617: This wrapper currently constructs and canonicalizes the full two-basis context, then asks it to construct and canonicalize the X projection. Ideally it should directly go through the X-selected internal driver so it performs one circuit analysis and one X-only canonicalization, with no full-context materialization.
The same applies to z_dem_from_memory_circuit() below. The full dem_from_memory_circuit() wrapper can use the full-selection path.

libs/qec/lib/experiments.cpp
```diff
@@ -481,66 +479,141 @@ dem_from_memory_circuit(const code &code, operation statePrep,
...
-                  r.length * numCols * sizeof(uint8_t));
-      dst_data += r.length * numCols;
-    }
+  decoder_context result;
+  result.num_rounds = numRounds;
+  result.num_x_stabilizers = numXStabs;
+  result.num_z_stabilizers = numZStabs;
+  result.fixed_basis_is_z = is_z_prep;
+
+  result.m2d = std::move(m2d.rows);
```
melody-ren: Could the context retain CUDA-Q’s M2DSparseMatrix and M2OSparseMatrix directly instead of extracting their row vectors and storing the measurement count separately?
Moving .rows is probably inexpensive, so my concern is not the move itself. The issue is that this decomposes an existing domain type into independently stored fields and weakens the matrix-shape invariant. Callers that only need the rows can access them from the native sparse map.
The genuinely optional conversion is flattening m2d into the realtime -1-terminated D_sparse representation. That allocation should remain conditional and ideally be expressed as a helper operating on M2DSparseMatrix, so generic dem_from_kernel consumers can use it without first repackaging their map into a decoder_context.

libs/qec/lib/experiments.cpp
```diff
@@ -481,66 +479,141 @@ dem_from_memory_circuit(const code &code, operation statePrep,
...
+  const uint32_t numBoundary = is_z_prep ? numZStabs : numXStabs;
+  dem.canonicalize_for_rounds_with_boundary(
+      numXStabs + numZStabs, numBoundary,
+      /*remove_zero_syndrome_errors=*/true);
+
+  result.dem = std::move(dem);
+  return result;
+}
+
+std::vector<std::int64_t> decoder_context::d_sparse() const {
```
melody-ren: This is the actual flattening/allocation step, and keeping it on demand is correct. I think we can make it a generic helper accepting M2DSparseMatrix (no longer just a member of the decoder_context). That keeps realtime serialization out of the analysis result and makes the conversion reusable by other dem_from_kernel callers.

melody-ren (commented):
Thanks @eliotheinrich! This a great upgrade to the decoder interface. I have some comments regarding API and performance. Please see inline comments.  

Probably out of scope for this PR given the timeframe, I think we can consider deprecating the legacy APIs.

libs/qec/lib/experiments.cpp
eliotheinrich: Yes, I agree, this was doing too much work, although I think having separate `*_decoder_context_from_memory_circuit` functions also introduces some overhead; I expect that usually both the X and Z errors models are relevant, and needing to call `dem_from_kernel` twice is likewise wasteful. The version at 88908c1 addresses this by performing `dem_from_kernel` once, and then providing access to a thin handle which provides x/z/full DEM+m2d+m2o, performing canonicalization lazily. The pattern is then
```
ctx_handle = decoder_context_from_memory_circuit(...)
ctx_x = ctx_handle.x_component()
ctx_z = ctx_handle.z_component()
```
or, for the both-basis context,
```
ctx = decoder_context_from_memory_circuit(...).full_component()
```

where canonicalization is called at the site of `*_component`. I think this is the best of both worlds; let me know if you have any thoughts about the API.

libs/qec/lib/experiments.cpp
eliotheinrich: Yes, agreed; see my comment above.

libs/qec/lib/experiments.cpp
eliotheinrich: Good point; fixed in 88908c1!

libs/qec/include/cudaq/qec/experiments.h
eliotheinrich: Yes, good point. The version at 88908c1 returns a handle which stores the internals as private fields, and then the provided `decoder_context` object which provide direct access to the DEM etc are now just the finalized result.

libs/qec/lib/experiments.cpp
eliotheinrich: I think the main comment above also addresses this

libs/qec/lib/experiments.cpp
eliotheinrich: Yes, good point. Updated in 88908c1. The m2d and m2o objects are now stored directly on the `decoder_context` object without casting to a 2d array. `d_sparse` is also now a public function, which can be used by both the internal helpers and consumers which have access to an `m2d` matrix in other contexts.

bmhowe23 (commented):
The PR introduces two related types whose names make the relationship harder to follow than it needs to be in my opinion:

- decoder_context - the result struct (dem + m2d + m2o), returned by component methods
- decoder_context_handle - the lazy factory returned by decoder_context_from_memory_circuit; call .x/z/full_component() to get a decoder_context

A few specific issues:

1. Python class is named DecoderContext, not DecoderContextHandle (py_code.cpp:534)
In Python, DecoderContext is what C++ calls decoder_context_handle - the lazy factory, not the final result. The C++ decoder_context struct has no Python binding (component methods return tuples). So Python users see DecoderContext as the object with .x/z/full_component() methods, which reads as "this is the context" when it's really "this is the thing that produces contexts." Either name the Python class DecoderContextHandle to match C++, or flip the naming so the more user-facing object (the handle) gets the simpler name.

2. Stray term memory_circuit_handle in two test comments
test_qec_stim.cpp:854 and test_dem.py:684 both say "decoder_context_from_memory_circuit returns a memory_circuit_handle" - a type that doesn't exist in the API. Should say decoder_context_handle.

3. Possible naming inversion to consider
If the intent is that users mostly interact with the factory object (decoder_context_from_memory_circuit(...)) and rarely hold a decoder_context directly, it might read more naturally to give the factory the simpler name (decoder_context) and add a suffix to the result struct (decoder_inputs, decoder_config, or canonicalized_decoder_context). As written, the prominent user-facing object has the longer name with _handle, while the plain data struct has the cleaner name.

eliotheinrich:
Good points, thank you Ben; I went with calling the factory object the `decoder_context`, and the returned struct a `decoder_input` (and fixed the stale comments) in [edf1b2c](https://github.com/NVIDIA/cudaqx/pull/671/commits/edf1b2c6750c3f51fc34074005789312b9c9cd33).

bmhowe23 (approved):
Thanks, Eliot,

PR #672, opened 2026-07-10
Fix surface-code hook errors with an orientation-aware CNOT schedule

bmhowe23:
## Description

The surface-code `stabilizer` kernel executed every plaquette's CNOTs in ascending data-qubit index order — the flattened parity matrices it receives carry no ordering information, so an optimized schedule could not even be expressed. With the same row-major order on both X and Z plaquettes, a mid-round ancilla fault ("hook error", Dennis et al., [arXiv:quant-ph/0110143](https://arxiv.org/abs/quant-ph/0110143)) propagates onto a data-qubit pair aligned *with* one of the logical operators, halving the effective circuit-level distance of one memory basis.

**Measured impact (before this PR, ZH orientation, `prep0`):** effective distance 2 instead of 3 at d=3, and 3 instead of 5 at d=5. Which basis is damaged depends on the code orientation; `prepp` was coincidentally unaffected for ZH.

This PR gives the surface code the standard zigzag CNOT schedule (Tomita & Svore, [arXiv:1404.3747](https://arxiv.org/abs/1404.3747)), selected per orientation so hook pairs land perpendicular to the same-type logical, and executes the X and Z checks interleaved over four shared timesteps (one depth-4 extraction round).

Design, with no breaking interface changes:

- `stabilizer_grid::get_cnot_schedule_x/z()`: schedule matrices with the same shape/support as the parity matrices; entry 0 = no support, k ∈ 1..4 = CNOT timestep. Rows match the sorted parity-matrix rows (enforced: a first-support tie throws instead of silently misaligning ancilla wiring). `get_cnot_schedule_pairs_x/z()` provides the flat (stabilizer, data) pair-list form used by the realtime examples.
- `code::get_stabilizer_schedule_x/z()`: new virtuals defaulting to the parity matrices, so the `stabilizer_round` kernel signature is untouched and repetition/Steane receive byte-identical inputs. The memory-circuit drivers pass these to the kernel.
- The surface kernel executes CNOTs timestep-major with X and Z checks interleaved; the zigzag pair is conflict-free and commutation-valid under interleaving (proven by the DEM tests — stim rejects non-deterministic detectors).
- The four realtime C++ app examples and the Python example consume the grid API instead of five copy-pasted ascending-order schedule builders; Python bindings return the schedule as numpy arrays consistent with `get_parity_x/z()`.
- Docs: the QEC introduction page documents the schedule-matrix kernel contract and the surface-code CNOT schedule, cross-referencing the Doxygen API docs as the single source of truth for the encoding.

Validation:

- New effective-distance test (`test_qec_stim`): builds the DEM from the actual kernel under two-qubit depolarizing noise and asserts the shortest undetectable logical error has weight d, for d=3,5 × both bases × all four orientations. Fails with weight ⌈d/2⌉ against the previous ordering.
- New schedule-property test (`test_qec`): d=3,5,7 × all orientations — parity-row alignment, valid distinct timesteps, hook-pair perpendicularity, no data qubit touched twice per timestep.
- The schedule matches ising-decoding's up to a left–right mirror (theirs starts each plaquette at NE, ours at NW) — same hook orientation and validity class; flipping the corner order is a one-line change if exact equivalence with predecoder training circuits is ever needed.

Monte Carlo (pymatching, CX depolarizing noise, 100k shots, rounds = d): at p=0.002 the `prep0` logical error rate improves **14×** at d=3 (0.030 → 0.0018) and **5.8×** at d=5 (0.0057 → 0.00098); `prepp` is statistically unchanged and the two bases are balanced after the fix. The pre-fix DEM also contained parallel edges with identical detector signatures but different observable effects — fundamentally undecodable mechanisms that pymatching's default merge strategy rejects outright; these disappear with the fix.

Build/test-time cost: the new tests add ~50 ms to the two unit-test binaries. Kernel classical overhead adds a constant-factor scan per stabilizer round (negligible on stim, which traces the kernel once).

### Before requesting review
- [ ] Full CI has been run.

### Scope and size
- [x] Refactoring-only changes are isolated in their own PR(s). *(the self-review hardening/dedup is a separate commit within this PR)*

### Tests
- [x] Tests fail if the new functionality is broken (including crashes), not just when it is missing. *(effective-distance test verified to fail against the previous ordering)*
- [ ] ~~Negative tests added where exceptions are expected.~~ The new no-tie guard in `build_cnot_schedule` is unreachable for any valid rotated surface code by construction (geometric argument in the code comment), so it cannot be exercised without an artificial grid; existing invalid-orientation/invalid-distance exception tests still cover the public entry points.
- [x] Truth data added where simple `EXPECT_*` / `assert` checks are insufficient for algorithmic correctness. *(effective distance == d via graphlike-DEM shortest-error search; Monte Carlo numbers above)*
- [x] CI runtime impact considered; team notified if significant. *(~50 ms added)*

### Documentation
- [x] User-visible behavior changes have public docs. *(QEC introduction page updated; full docs build verified locally — zero new warnings, new cross-references resolve)*

### Code style
- [x] Naming follows the existing convention (`snake_case`) for the area being modified.

### Dependencies

### DIFF
--- a/libs/qec/lib/codes/surface_code_device.cpp
+++ b/libs/qec/lib/codes/surface_code_device.cpp
@@ -49,38 +49,28 @@ __qpu__ void prepm(patch logicalQubit) {
 }
 
 __qpu__ std::vector<cudaq::measure_result>
-stabilizer(patch logicalQubit, const std::vector<std::size_t> &x_stabilizers,
-           const std::vector<std::size_t> &z_stabilizers) {
+stabilizer(patch logicalQubit,
+           const std::vector<std::size_t> &x_stabilizer_schedule,
+           const std::vector<std::size_t> &z_stabilizer_schedule) {
   for (std::size_t i = 0; i < logicalQubit.ancx.size(); i++)
     reset(logicalQubit.ancx[i]);
   for (std::size_t i = 0; i < logicalQubit.ancz.size(); i++)
     reset(logicalQubit.ancz[i]);
 
-  // The stabilizer matrices are CNOT schedules: entry 0 means the ancilla
-  // does not touch that data qubit, entry k >= 1 means the CNOT executes at
-  // timestep k. The X and Z checks share the timesteps, giving one depth-4
-  // extraction round. The order matters twice over: per plaquette it steers
-  // mid-round ancilla (hook) errors perpendicular to the logical operators,
-  // and across the X/Z interleaving it must be a valid schedule pair so the
-  // non-commuting CNOTs still measure the intended stabilizers (see
-  // stabilizer_grid::get_cnot_schedule_x).
-  std::size_t num_steps = 1;
-  for (std::size_t i = 0; i < x_stabilizers.size(); ++i)
-    if (x_stabilizers[i] > num_steps)
-      num_steps = x_stabilizers[i];
-  for (std::size_t i = 0; i < z_stabilizers.size(); ++i)
-    if (z_stabilizers[i] > num_steps)
-      num_steps = z_stabilizers[i];
+  // The X and Z checks share the timesteps, and the surface-code schedule
+  // always fits in one depth-4 extraction round (see
+  // stabilizer_grid::get_cnot_schedule_x), so num_steps is a constant.
+  const std::size_t num_steps = 4;
 
   h(logicalQubit.ancx);
   for (std::size_t step = 1; step <= num_steps; ++step) {
     for (std::size_t xi = 0; xi < logicalQubit.ancx.size(); ++xi)
       for (std::size_t di = 0; di < logicalQubit.data.size(); ++di)
-        if (x_stabilizers[xi * logicalQubit.data.size() + di] == step)
+        if (x_stabilizer_schedule[xi * logicalQubit.data.size() + di] == step)
           cudaq::x<cudaq::ctrl>(logicalQubit.ancx[xi], logicalQubit.data[di]);
     for (std::size_t zi = 0; zi < logicalQubit.ancz.size(); ++zi)
       for (std::size_t di = 0; di < logicalQubit.data.size(); ++di)
-        if (z_stabilizers[zi * logicalQubit.data.size() + di] == step)
+        if (z_stabilizer_schedule[zi * logicalQubit.data.size() + di] == step)
           cudaq::x<cudaq::ctrl>(logicalQubit.data[di], logicalQubit.ancz[zi]);
   }
   h(logicalQubit.ancx);
--- a/libs/qec/lib/codes/surface_code_device.cpp
+++ b/libs/qec/lib/codes/surface_code_device.cpp
@@ -56,18 +56,34 @@ stabilizer(patch logicalQubit, const std::vector<std::size_t> &x_stabilizers,
   for (std::size_t i = 0; i < logicalQubit.ancz.size(); i++)
     reset(logicalQubit.ancz[i]);
 
-  h(logicalQubit.ancx);
-  for (std::size_t xi = 0; xi < logicalQubit.ancx.size(); ++xi)
-    for (std::size_t di = 0; di < logicalQubit.data.size(); ++di)
-      if (x_stabilizers[xi * logicalQubit.data.size() + di] == 1)
-        cudaq::x<cudaq::ctrl>(logicalQubit.ancx[xi], logicalQubit.data[di]);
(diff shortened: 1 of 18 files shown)


libs/qec/lib/codes/surface_code_device.cpp
```diff
@@ -56,19 +56,35 @@ stabilizer(patch logicalQubit, const std::vector<std::size_t> &x_stabilizers,
...
+  // and across the X/Z interleaving it must be a valid schedule pair so the
+  // non-commuting CNOTs still measure the intended stabilizers (see
+  // stabilizer_grid::get_cnot_schedule_x).
+  std::size_t num_steps = 1;
+  for (std::size_t i = 0; i < x_stabilizers.size(); ++i)
+    if (x_stabilizers[i] > num_steps)
+      num_steps = x_stabilizers[i];
+  for (std::size_t i = 0; i < z_stabilizers.size(); ++i)
+    if (z_stabilizers[i] > num_steps)
+      num_steps = z_stabilizers[i];
```
eliotheinrich: `num_steps == 4`; don't think there's any need to calculate it at runtime here.

libs/qec/lib/codes/surface_code_device.cpp
eliotheinrich: The `*_stabilizers` variables now represent a cnot schedule. Should the variable names change to reflect that?

This change also slightly changes the semantics for user-provided stabilizer kernels in `introductions.rst:139`. Might be worth a note there.

libs/qec/python/bindings/py_surface_code.cpp
```diff
@@ -134,7 +135,37 @@ void bindSurfaceCode(nb::module_ &mod) {
            "Return the stabilizers as a list of cudaq::spin_op_term")
       .def("get_spin_op_observables", &stabilizer_grid::get_spin_op_observables,
            "Return the logical observables as [X, Z] cudaq::spin_op_term "
-           "entries");
+           "entries")
+      .def(
```
eliotheinrich, line 139: Should the new `get_stabilizer_schedule_*` methods on `code` be added as Python bindings?

It seems that there's no way for a user-defined Python code to provide a stabilizer schedule. This might be outside the scope of this surface-code PR but probably warrants a follow-up.

libs/qec/unittests/test_qec.cpp
```diff
@@ -1268,6 +1268,123 @@ TEST(QECCodeTester, checkStabilizerGrid) {
...
+          "surface_code", cudaqx::heterogeneous_map{
+                              {"distance", static_cast<std::size_t>(distance)},
+                              {"orientation", orientation_str}});
+      auto code_sched_x = code->get_stabilizer_schedule_x();
+      ASSERT_EQ(code_sched_x.shape(), sched_x.shape()) << label;
+      for (std::size_t s = 0; s < sched_x.shape()[0]; ++s)
+        for (std::size_t d = 0; d < num_data; ++d)
+          EXPECT_EQ(code_sched_x.at({s, d}), sched_x.at({s, d})) << label;
+    }
+  }
```
eliotheinrich, line 1390: It'd be nice peace of mind if these checks were run for the `sched_z` as well

eliotheinrich (commented):
Left a few suggestions, but looks good to me. Most substantial is probably the Python support, but that's maybe best suited for a follow-up.

libs/qec/lib/codes/surface_code_device.cpp
bmhowe23: Updated in 714131f45f4ccf5319e219cf3f2dcc83343dd06b

libs/qec/lib/codes/surface_code_device.cpp
bmhowe23: Good point. Updated in 714131f45f4ccf5319e219cf3f2dcc83343dd06b

libs/qec/unittests/test_qec.cpp
bmhowe23: Updated in 714131f45f4ccf5319e219cf3f2dcc83343dd06b

libs/qec/python/bindings/py_surface_code.cpp
bmhowe23: Added in 6a48445c49bb99235ae922cfdf68160c9b2bfeaf

kvmto (approved):
Looks good thanks, Ben!

PR #678, opened 2026-07-11
Follow up fixes for pinning hardware device

melody-ren:
Summary:

- Worker threads in the decoding server now pin themselves to the decoder's cuda_device_id before serving. Construction only pinned the registry thread, so every worker was decoding on the default device regardless of the pin prior to this patch.
- The direct call path (no realtime session) had the same problem: configure_decoders leaves the thread on the last decoder's device, so decoder 0 would decode on decoder N's GPU. Each decode now selects its own decoder's device first.
- If a pin can't be honored we fail instead of decoding on the wrong device: a worker that can't pin aborts server startup, and a cudaSetDevice failure during dispatch returns an error response instead of logging a warning and carrying on.
- Hololink: HOLOLINK_GPU_ID and cuda_device_id both name the GPU the FPGA is attached to, so they must agree. Either one alone selects the device; setting both to different values throws at transport creation. Single decoder only, same as the rest of the gpu_roce path.
- Nothing changes if cuda_device_id is not set.
- Decoder construction is transactional: if a plugin constructor throws after its device was selected, the calling thread's CUDA device is restored to its previous value instead of being left on the failed decoder's device. Originally first introduced in #687. Adopting the change here. 

Outdated: This is a follow-up PR to #664, which should be merged before this PR

This PR depends on #690

### DIFF
--- a/.cudaq_version
+++ b/.cudaq_version
@@ -1,6 +1,6 @@
 {
   "cudaq": {
     "repository": "NVIDIA/cuda-quantum",
-    "ref": "5439aff2c851a1b06ced9f24cdec7b5bfc3273ac"
+    "ref": "5f0ee5be3bf5cb33afb544d9f01af77b1d50b26e"
   }
 }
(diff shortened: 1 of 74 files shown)


bmhowe23 (approved):
Approved given that #692 will follow soon thereafter

PR #681, opened 2026-07-11
Consolidate decoder RPC wire format into a single header

bmhowe23:
## Summary

`decoder_rpc_ids.h` and `decoding-server-cqr/RpcWireFormat.h` carried duplicate copies of the decoder RPC wire format: the three FNV1a function IDs, the three packed request payload structs, and `bit_packed_bytes`. `RpcWireFormat.h` additionally mirrored cudaq-realtime's `RPCHeader`/`RPCResponse` structs and magic constants. The duplication existed only so the `cudaq-qec-decoding-server` static library could build without cudaq realtime headers. This PR drops that constraint and consolidates everything into one header.

## Changes

- Renamed `libs/qec/include/cudaq/qec/realtime/decoder_rpc_ids.h` to `decoder_rpc_wire_format.h` and made it the single source of truth for the decoder wire format (namespace `cudaq::qec::decoding::rpc`). It absorbs `RpcStatus` and `kMaxSyndromeBits` from `RpcWireFormat.h`, and pins `sizeof(cudaq::realtime::RPCHeader)`/`RPCResponse` to 24 bytes so a cudaq-side layout change cannot silently desynchronize the framing.
- Deleted `RpcWireFormat.h`. The decoding-server sources now use `cudaq::realtime::RPCHeader`/`RPCResponse`/`RPC_MAGIC_*` directly instead of layout-compatible mirrors, and the duplicate `EnqueuePayload`/`GetCorrectionsPayload`/`ResetPayload` structs are replaced by the shared `EnqueueRequestPayload`/`GetCorrectionsRequestPayload`/`ResetRequestPayload` types.
- `cudaq-qec-decoding-server` (and the `test_decoders_yaml`/`test_decoding_server_core` unit tests) is now built only when a realtime-enabled CUDA-Q install is found (`CUDAQ_REALTIME_INCLUDE_DIR`). Everything downstream (the `decoding_server` tool, the cqr plugin, the two-process test, the dependency-closure check) was already gated on this. Wheel builds, which don't provision cudaq realtime, simply skip these targets; they don't ship them anyway.


libs/qec/include/cudaq/qec/realtime/decoder_rpc_wire_format.h
bmhowe23, line 1: Note to self: this rename means we need accompanying revisions on the private repo.

libs/qec/lib/realtime/qec_realtime_session.cpp
melody-ren: The host_call path still uses hardcoded numbers. Should use the unified `RpcStatus` instead

bmhowe23: Updated in ac097a0c

melody-ren (approved):

PR #682, opened 2026-07-12
QEC decoding server: transport-agnostic bridge providers, one ring per decoder, mixed host + device_graph dispatch

bmhowe23:
Built on the companion cuda-quantum PR https://github.com/NVIDIA/cuda-quantum/pull/4915, **now merged upstream** (squash `2a7911f`, 2026-07-23); `.cudaq_version` pins that commit, so this branch builds against stock `NVIDIA/cuda-quantum` main.

## Summary

This PR re-architects the realtime decoding server around the CUDA-Q bridge-provider boundary. The server no longer contains any transport code: the wire is named in the deployment YAML (or a `--transport` fallback), loaded at runtime as a `libcudaq-realtime-bridge-<name>.so` provider, and every decoder gets its own ring buffer and its own consumer -- a host dispatcher thread for `dispatch: host` decoders, or the GPU device-graph scheduler for `dispatch: device_graph` decoders, both side by side in one server process. A partner transport library drops in as a single `.so` with zero decoding-server changes.

The guiding litmus test for the layering: cudaq-realtime and its transport providers never say "QEC" or "decoder"; the QEC library code never names a wire (UDP, RoCE, hololink). Wire names appear only where deployments are described: YAML configs, launch scripts, and tests.

## Before (main prior to #670; #670 landed the first step)

```
┌──────────────────────────────────────────────────────────────────────────┐
│                     decoding_server (pre-refactor)                       │
│  compiled-in wire branches, one per transport:                           │
│   #ifdef udp path      #ifdef cpu_roce path        gpu_roce fork         │
│   udp wrapper calls    RendezvousInfo + hsb_fpga   GpuRoceTransceiver =  │
│   wired directly       QP handshakes INLINE        hololink bring-up     │
│                        (byte-exact copy of the     FUSED with the QEC    │
│                        CpuRoceChannel structs)     dispatch engine       │
│  hand-rolled std::thread around cudaq_host_ring_dispatch_loop            │
│  ONE ring buffer + ONE dispatch loop shared by ALL decoders              │
│  two transport knobs that had to agree (--transport + per-decoder YAML)  │
└──────────────────────────────────────────────────────────────────────────┘
      │ link-time: udp/cpu_roce wrappers; DOCA + hololink + HSB + ibverbs
      ▼ adding a wire = editing this repo; no partner drop-in possible
```

- The server had per-transport code paths: udp and cpu_roce wired directly against transport wrapper headers, and a separate gpu_roce fork that linked hololink/DOCA/HSB/ibverbs into the QEC libraries (`DT_NEEDED` on the server binary). Adding a wire meant editing the server.
- Transport selection was two knobs that had to agree (`--transport` on the CLI AND a per-decoder YAML key), with silent misconfiguration when they disagreed.
- One ring buffer served all decoders: every RPC funneled through a single dispatcher, and per-decoder endpoints (the topology the product needs: caller routes with `device_id == decoder_id`) were impossible.
- The GPU device-graph path (`gpu_roce`) was hololink-only, named after one wire, and could not coexist with host-dispatch decoders in the same process.
- The CPU-RoCE rendezvous / HSB-FPGA QP handshake lived in the server itself instead of behind the transport boundary.
- #670 (now on main) took the first step -- the gpu_roce engine moved out of the core library behind a weak factory -- but the component still links hololink/DOCA/HSB directly, serves exactly one decoder, is named for one wire, and the per-decoder `transport:` key remains the selection knob.

## After

```
┌──────────────────────────────────────────────────────────────────────────┐
│                    decoding_server (transport-blind)                     │
│  YAML `dispatch:` picks the ENGINE     YAML `transport:` picks the WIRE  │
│   (per decoder)                        (per deployment; --transport is   │
│                                         a fallback, conflict = error)    │
│  ONE RING PER DECODER, each with its own consumer:                       │
│   host ─► dispatcher object over        <name>  ► libcudaq-realtime-     │
│           its provider ring (4869 API)            bridge-<name>.so       │
│   device_graph ─► DeviceGraphRing-      /path.so ► partner library,      │
│           Consumer (GPU scheduler)                verbatim, zero changes │
│  geometry + READY ring tokens derived FROM the providers (v2 queries)    │
└──────────────────────────────────────────────────────────────────────────┘
     │ links                            │ links (weak factory, WHOLE_ARCHIVE)
┌────────────────────┐   ┌────────────────────────────────────────────────┐
│ CQR plugin         │   │ DecodingServer core · device-graph component:  │
│ (HOST_CALL table)  │   │ DeviceGraphTransceiver = dispatch ENGINE only  │
└────────────────────┘   └────────────────────────────────────────────────┘
     │ links (the ONLY realtime link dependency)
┌──────────────────────────────────────────────────────────────────────────┐
│  libcudaq-realtime.so — bridge loader (iface v2) + dispatcher object     │
└──────────────────────────────────────────────────────────────────────────┘
     ◌ dlopen ─────────────── runtime plug-in seam ────────────── ◌ dlopen
┌────────────┬──────────────────┬─────────────────┬────────────────────────┐
│ bridge-udp │ bridge-cpu-roce  │ bridge-hololink │ partner transport .so  │
│ .so        │ .so (rendezvous  │ .so (built on   │ (out of tree, ~9 C     │
│            │ + hsb_fpga)      │ the HSB rig)    │ functions)             │
└────────────┴──────────────────┴─────────────────┴────────────────────────┘
```

- **Bridge-provider-only server.** `decoding_server` speaks only YAML + the `cudaq_bridge_*` C API. Provider libraries resolve by name next to the cudaq-realtime install (`QEC_BRIDGE_PROVIDER_DIR`) or verbatim by path (partner drop-in). The QEC libraries link no transport libraries; the rendezvous/hsb_fpga handshake moved down into the cpu_roce provider (companion cuda-quantum PR).
- **Engine vs. wire, each named once.** Per-decoder YAML `dispatch: host|device_graph` picks HOW RPCs execute; the top-level, shape-keyed `transport:` section (`provider`, `args`, plus a `device_graph:` override for the rings that must be GPU-pollable) picks the wire for the whole deployment. `--transport` is a fallback for configs that intentionally leave the wire unspecified (one YAML reused across wires, selected per launch); a YAML that names a provider combined with an explicit `--transport` is a startup error, never a silent precedence decision. All pre-release aliases (`transport:` as a per-decoder key, `--transport=gpu_roce`, `HOLOLINK_*` env fallbacks) are removed -- this component is new this cycle, so there is no compatibility surface to keep.
- **One ring per decoder.** The server opens one bridge instance + one ring + one consumer per YAML decoder; ring geometry and endpoint identity come from the provider's v2 queries instead of re-parsed CLI. Readiness publishes every endpoint in one line (`QEC_DECODING_SERVER_READY port=<p0> ... ring<id>=<port>`) before any blocking connect, and shutdown reports per-ring traffic (`QEC_DECODING_SERVER_RING decoder=<id> dispatched=<n>`). Callers route with `cudaq::device_call(decoder_id, ...)` and per-device endpoint args (`udp-port.<id>=<port>`).
- **Mixed dispatch in one process.** `DeviceGraphRingConsumer` runs the CUDA-Q device-graph scheduler (self-relaunching GPU dispatch graph + the decoder's captured decode graph) as a ring consumer over any GPU-pollable ring -- hololink DOCA rings on a rig, or pinned+mapped udp rings (`--pinned-rings`) on any CUDA box. `GpuRoceTransceiver` was renamed to what it actually is (`DeviceGraphTransceiver`, the dispatch engine, transport-blind) and now delegates to the ring consumer; the standalone all-device_graph path remains for the HSB flow.
- **Device-graph wedge root-caused and fixed.** The scheduler deadlock after the first decode trigger was cooperative-launch co-residency starvation: the decode graph was captured sized for every SM and could never become co-resident with the persistent dispatch graph, so the device-side fire queued forever at `grid.sync()`. `DecodingSession` now captures with `reserved_sms=1` (`QEC_DEVICE_GRAPH_RESERVED_SMS` to raise it on rigs where hololink RX/TX kernels are also resident). The same bug class affects host-path GPU-cooperative decodes when a scheduler is resident -- documented, with the mixed-deployment guidance of a CPU decoder on host rings until the plugin plumbs reservation into its host path.
- **Works with the session stats from main.** `QEC_DECODING_SERVER_STATS=1` prints per-decoder counters in the mixed server too; they are host-session counters, so a `device_graph` ring legitimately reports `decodes=0` -- its execution evidence is the trigger diagnostics (`fires == tail_relaunches`) and the per-ring `dispatched=` line.
- **Tests and config schema.** New coverage: transport-section parse/round-trip with a mixed host+device_graph decoder set, a two-process decode where the YAML alone names the wire, CLI/YAML conflict rejection, a per-decoder-rings two-process test asserting traffic on each ring, and `app_examples.surface_code-4-yaml-mixed-dispatch` -- the flagship mixed flow as a gated ctest (registered only when the server links the device-graph component; skips without a GPU or the nv-qldpc plugin) asserting scheduler health via the new trigger diagnostics (rc=0, fires == tail_relaunches > 0). The generated JSON config schema (`decoder_config_json_schema()`) advertises the new per-decoder `dispatch:` key and the top-level `transport:` section (the removed per-decoder `transport:` key is gone from the schema too), with a python `jsonschema` test that validates a populated `dispatch`/`transport` document against it and rejects the removed key. The python config bindings expose the new fields as well -- `decoder_config.dispatch`, `multi_decoder_config.transport`, and the `decoder_dispatch` / `transport_config` / `transport_shape_override` types.

## Sample configurations

Three representative deployment shapes (YAML config + exact launch line + matching caller config):

**Two host decoders, one udp ring per decoder** -- the YAML says nothing about the wire, so `--transport` (default udp) selects it per launch; the READY line publishes every ring's endpoint.

```yaml
decoders:
  - id: 0
    type: pymatching
    # block_size / syndrome_size / H_sparse / ...
  - id: 1
    type: pymatching
    # ...
```

```
decoding_server --config=decoders.yaml
# QEC_DECODING_SERVER_READY port=<P0> transport=udp ring0=<P0> ring1=<P1>
```

Caller routes per decoder with device-scoped endpoint args: `--cudaq-device-call=udp udp-host=127.0.0.1 udp-port=<P0> udp-port.1=<P1>`.

**Mixed dispatch (host CPU decoder + GPU device-graph decoder), runs on any CUDA box** -- the wire lives in the YAML transport section; the shape-keyed override adds `--pinned-rings` only to the device_graph ring so the GPU scheduler can poll it. No `--transport` on the command line (combining both is a startup error). This is exactly what the new gated ctest runs.

```yaml
transport:
  provider: udp
  device_graph:
    args: [--pinned-rings]
decoders:
  - id: 0
    type: multi_error_lut
    # ...
  - id: 1
    type: nv-qldpc-decoder
    dispatch: device_graph
    # ...
```

```
decoding_server --config=decoders.yaml
```

**Partner transport drop-in** -- an out-of-tree provider library is named by path, verbatim; its args are forwarded untouched. Zero decoding-server changes.

```yaml
transport:
  provider: /opt/partner/libpartner_bridge.so
  args: [--lane=3]
decoders:
  - id: 0
    type: pymatching
    # ...
```

```
decoding_server --config=decoders.yaml
```

## Validation

- Two-process suite 6/6 (udp; includes per-decoder rings and the YAML-transport-section tests), decoding-server core + YAML suites, per-decoder-rings app example.
- Re-validated after merging main with #670 landed (#656/#673/#674/#678/#680/#683/#690/#691/#692): two-process suite, DecoderYAMLTest 18/18, per-decoder-rings + mixed-dispatch app tests, python surface_code-1, and the full mixed E2E all green against an nv-qldpc plugin rebuilt for #674's virtualized decoder API (an ABI break for out-of-tree plugin builds).
- Full mixed E2E on a WSL2 laptop, two-process, wire named only by the YAML: multi_error_lut on a host ring + nv-qldpc RelayBP behind the GPU device-graph scheduler on a pinned-udp ring -- decode graph fired for every window (trigger rc=0, fires == tail_relaunches), correct corrections, both rings dispatched, clean teardown. The HSB rig is no longer required to exercise the device-graph dispatch path; rig runs remain the validation for the hololink and cpu_roce RDMA data planes.

## Breaking changes vs. main

All of these surfaces are new this release cycle, so no deprecation aliases are kept:

- The per-decoder `transport:` YAML key and the `DecoderTransport` enum are replaced by per-decoder `dispatch: host|device_graph` plus the top-level `transport:` section. A config using the old key fails to parse loudly. `cuda_device_id` (#690) is kept unchanged and now also places the device_graph rings and scheduler.
- The device-graph transceiver's environment surface is `QEC_DEVICE_GRAPH_*` (was `HOLOLINK_*`), and `--transport=gpu_roce` no longer exists: device_graph is a dispatch shape named in the YAML; `--transport` only ever names a wire provider. The HSB test script is updated to match (injects `dispatch: device_graph` + `cuda_device_id`, sets `QEC_DEVICE_GRAPH_*`, waits for `READY device_graph`).
- `CpuRoceTransceiver` (the always-throwing placeholder) is removed; cpu_roce is served by its bridge provider.

## Dependencies

- **Built on the companion cuda-quantum PR [#4915](https://github.com/NVIDIA/cuda-quantum/pull/4915)** ([realtime] pluggable transport providers / bridge interface v2 + per-device device_call sessions), **merged upstream 2026-07-23** (squash `2a7911f`) and itself built on cuda-quantum #4869. `.cudaq_version` pins that squash commit, so this branch builds against stock main. This PR consumes `cudaq_bridge_create_from_library`, the v2 endpoint/geometry queries, per-device sessions and device-scoped channel args, the udp provider's `--pinned-rings`, and `cudaq_dispatch_get_trigger_debug`.
- **Built on main with #670 landed** (origin/main is merged into this branch). Relative to landed #670, this PR completes the follow-up its review called for -- removing the server's transport awareness: `GpuRoce{Transceiver,Factory,LinkCheck}` become `DeviceGraph*` (same weak-factory / optional-component structure, now transport-blind), while #670's post-review hardening (ring-size and host-page-alignment validation, `cuda_device_id`-driven GPU placement threaded through the factory) is preserved in the renamed files.
- The nv-qldpc decoder plugin and the proprietary cudevice archive are consumed as prebuilt binaries (`CUDAQ_QEC_REALTIME_CUDEVICE_PROPRIETARY_ARCHIVE`); builds without them still compile and the device-graph paths fail at runtime with a clear not-linked error, and the gated ctest is simply not registered.


libs/qec/lib/realtime/decoding-server-cqr/DeviceGraphTransceiver.cpp
```diff
@@ -0,0 +1,327 @@
...
+    cudaq_bridge_destroy(bridge_);
+    bridge_ = nullptr;
+    throw std::runtime_error(
+        "DeviceGraphTransceiver: provider does not report endpoint info "
+        "(bridge interface v2 required)");
+  }
+  const std::string endpoint_info(info);
+  parse_endpoint_token(endpoint_info, "qp", qp_number_);
+  parse_endpoint_token(endpoint_info, "rkey", rkey_);
+  parse_endpoint_token(endpoint_info, "buffer_addr", buffer_addr_);
```
1tnguyen: These seem to be HSB-specific info. Perhaps we can drop them (look like we don't use these anymore)

libs/qec/lib/realtime/decoding-server-cqr/DeviceGraphTransceiver.cpp
```diff
@@ -0,0 +1,327 @@
...
+      "--remote-qp=" + std::to_string(config.remote_qp),
+      "--gpu=" + std::to_string(config.gpu_id),
+      "--page-size=" + std::to_string(page_size),
+      "--num-pages=" + std::to_string(config.num_pages),
+      "--payload-size=" + std::to_string(payload_size),
+  };
+  std::vector<char *> argv;
+  argv.reserve(args.size());
+  for (auto &a : args)
+    argv.push_back(const_cast<char *>(a.c_str()));
```
1tnguyen, line 157: I think we'd need to generalize this for non-HSB bridge implementation.

melody-ren:
Not sure if this is still relevant after this PR as I haven't read the code, but as a note to self: should use config as the single source of truth for parameters and remove env vars used on the realtime path (e.g., gpu roce accepts env var and config)

.cudaq_version
```diff
@@ -1,6 +1,6 @@
 {
   "cudaq": {
-    "repository": "NVIDIA/cuda-quantum",
-    "ref": "6cc5f87f6cc3440ba1a518e8e9111b96da5e6581"
+    "repository": "bmhowe23/cuda-quantum",
+    "ref": "26947049747ed77bb245080454b596bcdcdc2a78"
```
bmhowe23: Just making this comment obvious...we shouldn't merge until the corresponding upstream PR lands in the main repo first.

cketcham2333: Resolved in 2ae4d454f735a712db4b59afb82e1caf6f3c0ce5

libs/qec/lib/realtime/decoding-server-cqr/DeviceGraphTransceiver.cpp
cketcham2333: Will be fixed in a follow up PR. Refer to https://jirasw.nvidia.com/browse/CUDRR1FA-534

libs/qec/lib/realtime/decoding-server-cqr/DeviceGraphTransceiver.cpp
cketcham2333: Resolved in two separate commits. First one is https://github.com/NVIDIA/cudaqx/commit/49eb8e1d00376e2212c504baca881ba99c8c994b. Second one is https://github.com/NVIDIA/cudaqx/commit/44cda1ac9ea9f6438dbdaf324920d19f6aebc7d5

cketcham2333 (approved):

PR #683, opened 2026-07-13
Fix GB200 gpu_roce decoding-server validation issues

vedika-saravanan:
## Description

Follow-up fixes on top of the HSB/FPGA `gpu_roce` decoding-server path, found during end-to-end GB200 validation of RelayBP / `nv-qldpc-decoder` with the FPGA emulator.

Validated after rebase with an 85-shot / 425-frame `gpu_roce` RelayBP run against the FPGA emulator, using 512 playback windows. The emulator processed all windows successfully with 425 responses, 0 send errors, and 0 timeouts.

Changes:

- Pin decoding-server graph capture to the decoder `cuda_device_id`, and reconcile it with the transitional `HOLOLINK_GPU_ID` setting.
- Validate the server-side `gpu_roce` DOCA ring allocation against host page-size alignment, and make the HSB test harness choose an aligned server ring depth while keeping playback window capacity explicit.
- Link optional private `decoder_rpc_dispatch.cu` support into `decoding_server` when present; public builds remain unaffected.
- Avoid duplicate public/proprietary cudevice whole-archive linkage in `decoding_server`.
- Make the `gpu_roce` link check verify symbol/link availability without constructing a transceiver or selecting GPU 0.
- Fix HSB playback verification to wait for the expected ILA sample count instead of stabilization, and fail clearly if the capture times out before enough samples are collected.
- Add separate HSB harness controls for server `gpu_roce` ring pages and FPGA/emulator playback window pages.

Note: I left this local intentionally because the `HOLOLINK_GPU_ID` path is transitional and PR692 removes it. Sharing the env parser now would add churn and increase conflicts for code that is about to disappear.

na

## Self-review checklist

through and explain.

### Before requesting review
- [ ] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/libs/qec/lib/realtime/CMakeLists.txt
+++ b/libs/qec/lib/realtime/CMakeLists.txt
@@ -181,30 +181,6 @@ if(CMAKE_CUDA_COMPILER)
       ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
     )
 
-    # Private builds may provide decoder_rpc_dispatch.cu, which exports the
-    # decoder-state registration and DEVICE_CALL handler glue that graph-based
-    # decoders resolve via dlsym(RTLD_DEFAULT, ...). Keep this optional so the
-    # public tree still builds without proprietary sources.
-    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/decoder_rpc_dispatch.cu")
-      add_library(cudaq-qec-realtime-decoder-rpc-dispatch STATIC
-        decoder_rpc_dispatch.cu
-      )
-      set_target_properties(cudaq-qec-realtime-decoder-rpc-dispatch PROPERTIES
-        CUDA_SEPARABLE_COMPILATION ON
-        CUDA_STANDARD 17
-        POSITION_INDEPENDENT_CODE ON
-        ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
-      )
-      target_include_directories(cudaq-qec-realtime-decoder-rpc-dispatch PUBLIC
-        $<BUILD_INTERFACE:${CUDAQX_QEC_INCLUDE_DIR}>
-        $<BUILD_INTERFACE:${_core_include_dir}>
-        $<BUILD_INTERFACE:${CUDAQ_REALTIME_INCLUDE_DIR}>
-        $<INSTALL_INTERFACE:include>
-      )
-      target_link_libraries(cudaq-qec-realtime-decoder-rpc-dispatch PUBLIC
-        CUDA::cudart
-      )
-    endif()
   else()
     message(WARNING "cuda-quantum realtime headers not found. "
                     "Device library cudaq-qec-realtime-device will not be built. "
--- a/libs/qec/lib/realtime/CMakeLists.txt
+++ b/libs/qec/lib/realtime/CMakeLists.txt
@@ -204,10 +204,6 @@ if(CMAKE_CUDA_COMPILER)
       target_link_libraries(cudaq-qec-realtime-decoder-rpc-dispatch PUBLIC
         CUDA::cudart
       )
-      install(TARGETS cudaq-qec-realtime-decoder-rpc-dispatch
-        COMPONENT qec-lib
-        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-      )
     endif()
   else()
     message(WARNING "cuda-quantum realtime headers not found. "
--- a/libs/qec/lib/realtime/CMakeLists.txt
+++ b/libs/qec/lib/realtime/CMakeLists.txt
@@ -180,6 +180,35 @@ if(CMAKE_CUDA_COMPILER)
       COMPONENT qec-lib
       ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
     )
+
+    # Private builds may provide decoder_rpc_dispatch.cu, which exports the
+    # decoder-state registration and DEVICE_CALL handler glue that graph-based
+    # decoders resolve via dlsym(RTLD_DEFAULT, ...). Keep this optional so the
+    # public tree still builds without proprietary sources.
+    if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/decoder_rpc_dispatch.cu")
+      add_library(cudaq-qec-realtime-decoder-rpc-dispatch STATIC
+        decoder_rpc_dispatch.cu
+      )
+      set_target_properties(cudaq-qec-realtime-decoder-rpc-dispatch PROPERTIES
+        CUDA_SEPARABLE_COMPILATION ON
+        CUDA_STANDARD 17
+        POSITION_INDEPENDENT_CODE ON
+        ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+      )
+      target_include_directories(cudaq-qec-realtime-decoder-rpc-dispatch PUBLIC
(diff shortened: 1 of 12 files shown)


libs/qec/lib/realtime/decoding-server-cqr/DecodingSession.cpp
```diff
@@ -12,13 +12,55 @@
...
+  } catch (const std::exception &) {
+    throw std::runtime_error(std::string("invalid ") + name +
+                             " value: " + value);
+  }
+}
+
+int graph_capture_device(const cudaq::qec::decoder &decoder) {
+  const int decoder_pin = decoder.get_cuda_device_id();
+  const auto hololink_gpu_id = env_int_optional("HOLOLINK_GPU_ID");
+  if (hololink_gpu_id && decoder_pin >= 0 && *hololink_gpu_id != decoder_pin)
```
bmhowe23: Noting: I expect Melody's PR #692 to change this. I think we should be removing all references to HOLOLINK_GPU_ID and using cuda_device_id instead. Let me know if you disagree. Otherwise, I expect Melody to remove it in the next PR.

vedika-saravanan: Agreed. I kept the HOLOLINK_GPU_ID reconciliation here only as a transitional guard for the current gpu_roce path while PR683 is stacked before PR692.

Once PR692 lands, cuda_device_id should be the single source of truth and this HOLOLINK_GPU_ID path can go away. I’ll avoid doing more cleanup here to keep PR683 scoped and reduce conflicts with Melody’s PR.

libs/qec/lib/realtime/CMakeLists.txt
```diff
@@ -180,6 +180,35 @@ if(CMAKE_CUDA_COMPILER)
...
+        $<BUILD_INTERFACE:${_core_include_dir}>
+        $<BUILD_INTERFACE:${CUDAQ_REALTIME_INCLUDE_DIR}>
+        $<INSTALL_INTERFACE:include>
+      )
+      target_link_libraries(cudaq-qec-realtime-decoder-rpc-dispatch PUBLIC
+        CUDA::cudart
+      )
+      install(TARGETS cudaq-qec-realtime-decoder-rpc-dispatch
+        COMPONENT qec-lib
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
bmhowe23: Are you sure this is really needed? I don't think we should need to "install" any .a files if they are being linked into the decoding server.

bmhowe23: See #694 for other examples that I am trying to remove.

vedika-saravanan: Good point. I originally added this while making the optional private dispatch target explicit, but the archive is only a build/link input for decoding_server , it does not need to be part of install packaging. I removed the install rule https://github.com/NVIDIA/cudaqx/pull/683/commits/b7c096309698f0e71765f4997763f2668d6723d1.

libs/qec/lib/realtime/decoding-server-cqr/DecodingSession.cpp
cketcham2333: Instead of adding `graph_capture_device`, could you just call the existing `reconcile_gpu_roce_device` (from `DecodingServer.h`), passing the local `HOLOLINK_GPU_ID` value and `decoder.get_cuda_device_id()`? That's less code, and it leaves one function deciding which GPU to use instead of two. Today the two can disagree — if nothing is pinned, one picks GPU 0 while the other stays on the current GPU — which is the exact problem the pinning is meant to prevent (recording the graph on one GPU and running it from another).

It also helps #692: once #692 changes `reconcile_gpu_roce_device` to take just `(int)`, this line won't compile until it's updated — so #692 can't accidentally skip `DecodingSession` and leave a `HOLOLINK_GPU_ID` read here (its current diff doesn't touch this file).

The one thing to leave for #692 is the rename. This function isn't really gpu_roce-specific — it just answers "which GPU is this decoder pinned to?", and `DecodingSession` uses it for *every* graph decoder (udp and cpu_roce too), not only gpu_roce. After #692 removes `HOLOLINK_GPU_ID`, nothing about it is gpu_roce at all. Keeping a transport-specific name on general decoder-device logic works against making the decoding server transport-agnostic, so it should get a neutral name (e.g. `resolve_decode_device`) — but that fits naturally in #692, which is already reworking this function.

melody-ren: That's a good point about the name and I agree with which PR that renaming should live. Will make aforementioned name change in #692 .

libs/qec/tools/decoding-server/CMakeLists.txt
```diff
@@ -135,14 +137,30 @@ if(CUDAQ_REALTIME_INCLUDE_DIR AND TARGET cudaq-qec-realtime-decoding-server-cqr)
...
+        if(NOT _decoding_server_linked_proprietary_cudevice)
+          target_link_libraries(decoding_server PRIVATE
+            $<LINK_LIBRARY:WHOLE_ARCHIVE,cudaq-qec-realtime-cudevice-proprietary>
+          )
+        endif()
+      elseif(TARGET cudaq-qec-realtime-cudevice)
+        target_link_libraries(decoding_server PRIVATE
+          $<LINK_LIBRARY:WHOLE_ARCHIVE,cudaq-qec-realtime-cudevice>
+        )
+      endif()
```
cketcham2333, line 158: This lib guarantees a duplicate-symbol link failure in the private (cuda-qx) build — it should be dropped.

`decoder_rpc_dispatch.cu` is proprietary, so it exists only in the private tree. The `if(EXISTS ...)` guard that builds this lib (realtime/CMakeLists.txt:188) is therefore true only in the private build — which is exactly where the proprietary archive is built from that same file (it holds that one object, `decoder_rpc_dispatch.cu.o`). So `decoding_server` whole-archives the same object twice: the archive at line 124 and this lib at line 158 → multiple-definition link error. Public CI can't catch it — the source isn't there, so this lib is never built in public.

Dropping it (rather than guarding the link) is the right fix because the lib is redundant: in the only build where it exists, the archive already whole-archives these symbols with `--export-dynamic`, so the `dlsym(RTLD_DEFAULT, ...)` path it feeds already works.

Fix: remove the whole-archive here (lines 158-162) and the `add_library` that creates it (realtime/CMakeLists.txt:188-208).

bmhowe23 (approved):
LGTM once Chuck's threads are marked as resolved.

libs/qec/tools/decoding-server/CMakeLists.txt
vedika-saravanan: Thanks Chuck, fixed in https://github.com/NVIDIA/cudaqx/pull/683/commits/272b84dcabc6813a0e90b6530967de38383dd329. I removed the redundant `decoder_rpc_dispatch.cu` archive target and the matching whole-archive link, so the private build now relies on the proprietary cudevice archive as the single source for those symbols.

libs/qec/lib/realtime/decoding-server-cqr/DecodingSession.cpp
vedika-saravanan: agreed. I removed the separate `graph_capture_device` helper and now call `reconcile_gpu_roce_device(env_int_optional("HOLOLINK_GPU_ID"), decoder.get_cuda_device_id())` from `DecodingSession`. I left the rename for PR692 as suggested. https://github.com/NVIDIA/cudaqx/commit/272b84dcabc6813a0e90b6530967de38383dd329

PR #684, opened 2026-07-13
fix: To correct the ADAPT-VQE default return energy to the initial state energy

kaiqiy-nv:
This PR tries to fix [[B] 6445945](https://nvbugspro.nvidia.com/bug/6445945) .
It also adds some tests to expose this bug and verify the fix.
Thanks for looking this PR.

This suggested fix computes `energy` from `observe(prepare_state, H, state).expectation()` immediately after preparing the initial state, and initialises `latestEnergy` from the same value.
In this case, all early exits return the correct energy and energy-difference convergence uses a valid baseline.


kvmto (approved):
LGTM!

kaiqiy-nv:
As is discussed, we may not handle this issue for now. Closing this PR.

PR #685, opened 2026-07-13
Surface code demo 1 refactor

eliotheinrich:
## Description

This PR updates the surface code demo (/libs/qec/unittests/realtime/app_examples/surface_code-1.cpp) to use the new syndrome format and decoder context objects. Semantically, the demo should remain very similar to the original version, but I have aimed to remove the hand-rolled sections and replace them with CUDA-QX features as much as possible. I have also refactored for readability.

### Notes for reviewers
The return format of the entry QPU kernel changed; now, the low 32 bits correspond to the values of the logical qubits, and the high 32 bits correspond to the total number of corrections applied by the decoders during the shot. The test can therefore now accommodate up to 32 logical qubits with no restrictions on distance.

Stationary window removed; now either monolithic decoding or sliding window. See the performance note below.

Depends on #656 and #671.

The removal of the stationary decoding window affected some multi-error-lut tests, as the LUT tables were too large. In particular, I had to reduce the number of rounds in the surface-code-1-quantinuum distance=5 test, although it still represents a performance regression on it and some similar tests. I'm open to re-introducing the decoder window if we want to keep that functionality, but as it doesn't represent a proper decoding flow, I prefer to remove it.

## Self-review checklist

through and explain.

### Before requesting review
- [ ] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/docs/sphinx/examples_rst/qec/realtime_relay_bp.rst
+++ b/docs/sphinx/examples_rst/qec/realtime_relay_bp.rst
@@ -375,7 +375,7 @@ Surface Code Test (Relay BP)
 
 The ``surface_code-1-local`` app example drives the device-graph scheduler
 through the in-process RPC path (``CUDAQ_QEC_REALTIME_MODE=inproc_rpc``) with
-the nv-qldpc-decoder configured for Relay BP (``--use-relay-bp``).  It simulates
+the nv-qldpc-decoder configured for Relay BP.  It simulates
 a surface code with ``stim`` and generates syndromes on the fly, so -- unlike
 the fixed-fixture CI unit test -- it can run an arbitrary number of shots.
 
@@ -398,20 +398,19 @@ loop through the scheduler:
    APP=./libs/qec/unittests/realtime/app_examples/surface_code-1-local
 
    # 1. Generate the Relay BP decoder config (DEM) for a distance-3 surface code
-   "$APP" --distance 3 --num_rounds 12 --decoder_window 6 \
-          --decoder_type nv-qldpc-decoder --use-relay-bp \
+   "$APP" --distance 3 --num_rounds 12 \
+          --decoder_type nv-qldpc-decoder \
           --num_shots 1000 --save_dem config.yml
 
    # 2. Run the decode loop through the device-graph scheduler
-   "$APP" --distance 3 --num_rounds 12 --decoder_window 6 \
-          --decoder_type nv-qldpc-decoder --use-relay-bp \
+   "$APP" --distance 3 --num_rounds 12 \
+          --decoder_type nv-qldpc-decoder \
           --num_shots 1000 --load_dem config.yml
 
 A clean run exits ``0`` and reports a small number of non-zero syndrome
 measurements alongside a larger number of corrections found.  The
 ``app_examples`` CTest ``surface_code-1-local-test-distance-3-inproc-rpc``
-wraps this flow (it sets ``CUDAQ_QEC_REALTIME_MODE=inproc_rpc`` and
-``EXTRA_CLI_ARGS=--use-relay-bp``).
+wraps this flow (it sets ``CUDAQ_QEC_REALTIME_MODE=inproc_rpc``).
 
 Emulated End-to-End Test
 ------------------------
--- a/libs/qec/unittests/realtime/app_examples/CMakeLists.txt
+++ b/libs/qec/unittests/realtime/app_examples/CMakeLists.txt
@@ -78,7 +78,7 @@ add_test(
     bash "${CMAKE_CURRENT_SOURCE_DIR}/surface_code-1-test.sh"
       ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-local
       ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-local
-      3 40 40 NULL 12 6
+      3 60 30 NULL 12 6
       ${CMAKE_BINARY_DIR}/lib
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
 )
@@ -91,7 +91,7 @@ if(_APP_CUDAQ_REALTIME_DISPATCH_LIB
       bash "${CMAKE_CURRENT_SOURCE_DIR}/surface_code-1-test.sh"
         ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-local
         ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-local
-        3 60 40 NULL 12 6
+        3 60 30 NULL 12 6
         ${CMAKE_BINARY_DIR}/lib
         nv-qldpc-decoder
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
@@ -113,7 +113,7 @@ add_test(
     bash "${CMAKE_CURRENT_SOURCE_DIR}/surface_code-1-test.sh"
       ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-local
       ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-local
-      3 40 40 NULL 12 6
+      3 60 30 NULL 12 6
       ${CMAKE_BINARY_DIR}/lib
       sliding_window 6 1
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
@@ -204,7 +204,7 @@ if(TARGET cudaq-qec-realtime-decoding-server-cqr)
       bash "${CMAKE_CURRENT_SOURCE_DIR}/surface_code-1-test.sh"
         ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-cqr
         ${CMAKE_CURRENT_BINARY_DIR}/surface_code-1-cqr
-        3 60 40 NULL 12 6
+        3 60 30 NULL 12 6
         ${CMAKE_BINARY_DIR}/lib
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   )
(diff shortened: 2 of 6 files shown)


libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
```diff
@@ -146,825 +203,517 @@ void save_dem_to_file(const cudaq::qec::detector_error_model &dem,
         nv_args.insert("srelay_config", srelay_args);
         nv_args.insert("gamma_dist", std::vector<double>{0.1, 0.2});
       } else {
-        // OSD post-processor
-        nv_args.insert("use_osd", true);
+        nv_args.insert("use_osd", true); // BP + OSD
```
bmhowe23: let's remove the BP+OSD else branch for now.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
```diff
@@ -146,825 +203,517 @@ void save_dem_to_file(const cudaq::qec::detector_error_model &dem,
...
+                           cudaq::qec::operation state_prep,
+                           const run_options &opts) {
+  experiment exp;
+  exp.num_data = code.get_num_data_qubits();
+  exp.num_ancx = code.get_num_ancilla_x_qubits();
+  exp.num_ancz = code.get_num_ancilla_z_qubits();
+  exp.num_rounds = static_cast<std::size_t>(opts.num_rounds);
+  exp.num_logical = static_cast<std::size_t>(opts.num_logical);
+
+  auto parity_x = code.get_parity_x();
```
melody-ren: I don't think we should be using the parity here. These two lines should be using get_stabilizer_schedule_x/z to match with the DEM

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
```diff
@@ -146,825 +203,517 @@ void save_dem_to_file(const cudaq::qec::detector_error_model &dem,
...
-          cudaq::x(logical.data[3]);
-        }
-#endif
-      }
-    }
-  }
+    std::stringstream config_text;
+    config_text << config_file.rdbuf();
+    auto cfg = config::multi_decoder_config::from_yaml_str(config_text.str());
+    config::configure_decoders(cfg);
```
melody-ren, line 378: Let's validate the loaded config against the code's geometry and so on. Right now, something like this can pass through: d=5/R=5 and d=3/R=15 both have 120 detectors but require 145 versus 129 raw measurements.

I think @kvmto first spotted this issue in surface code 4 example

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
```diff
@@ -146,825 +203,517 @@ void save_dem_to_file(const cudaq::qec::detector_error_model &dem,
...
   printf("Options:\n");
-  printf("  --distance <int>    Distance of the surface code. Default: 5\n");
-  printf("  --num_shots <int>   Number of shots. Default: 10\n");
-  printf(
-      "  --p_spam <double>   SPAM probability. Range[0, 1]. Default: 0.01\n");
-  printf("  --num_logical <int> Number of logical qubits. Default: 1\n");
-  printf("  --num_rounds <int>  Number of measurement rounds. Default: "
+  printf("  --distance <int>      Distance of the surface code. Default: 5\n");
+  printf("  --num_shots <int>     Number of shots. Default: 10\n");
+  printf("  --p_cnot <double>     Two-qubit depolarizing rate on CNOT gates. "
```
melody-ren, line 688: There seems to be some leftover mention of the now deleted/replaced p_spam in the doc .rst. Also, since this PR removes the stationary window, we should also remove the flag `decoder_window` from the driver, ctest and elsewhere

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
```diff
@@ -975,186 +724,130 @@ void show_help() {
...
+  if (!opts.save_dem)
     initialize_realtime_channel(argv[0]);
 #endif
 
-  demo_circuit_host(*code, distance, p_spam, cudaq::qec::operation::prep0,
-                    num_shots, num_rounds, num_logical, dem_filename, save_dem,
-                    load_dem, decoder_window, decoder_type, sw_window_size,
-                    sw_step_size, save_syndrome, load_syndrome,
-                    syndrome_filename, use_relay_bp);
+  demo_circuit_host(*code, opts);
```
melody-ren: I think we need to throw whenever there's an error along the workflow path. E.g., `main` always returns 0, and earlier failures such as capture syndrome will print an error and return, and therefore unless we check the log we won't see operational failures

libs/qec/unittests/realtime/app_examples/CMakeLists.txt
```diff
@@ -245,7 +246,7 @@ cudaqx_add_device_code(surface_code-1-quantinuum-emulate
   SOURCES
     surface_code-1.cpp
   COMPILER_FLAGS
-    --target quantinuum --emulate --quantinuum-machine Helios-Fake -DMANUALLY_INJECT_ERRORS=1 
+    --target quantinuum --emulate --quantinuum-machine Helios-Fake
```
melody-ren, line 312: This will reduce the test coverage for the Quantinuum path. If the same test is covered elsewhere by other examples in the repo, then I think we're ok with removing it here. But generally, I think we should try to preserve test coverage

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
```diff
@@ -146,825 +203,517 @@ void save_dem_to_file(const cudaq::qec::detector_error_model &dem,
...
+// Append the per-shot corrections so a later --load_syndrome run can verify the
+// replayed corrections against them, then close the capture file.
+void finish_syndrome_capture(const std::vector<std::int64_t> &run_result,
+                             const std::string &filename) {
+  if (!g_capture.file.is_open())
+    return;
+  cudaq::qec::decoding::host::_set_syndrome_capture_callback(nullptr);
+  g_capture.file << "CORRECTIONS_START\n";
+  for (auto shot : run_result)
+    g_capture.file << ((shot >> 32) > 0 ? 1 : 0) << "\n";
```
melody-ren, line 463: I think this narrows the previous per logical qubit correction to any correction

melody-ren (commented):
Thanks @eliotheinrich! I think there's a python counter part to this example which should also be updated. But that can come in a follow up to move things along quicker

docs/sphinx/examples_rst/qec/realtime_relay_bp.rst
```diff
@@ -398,20 +398,19 @@ loop through the scheduler:
...
+   "$APP" --distance 3 --num_rounds 12 \
+          --decoder_type nv-qldpc-decoder \
           --num_shots 1000 --load_dem config.yml
 
 A clean run exits ``0`` and reports a small number of non-zero syndrome
 measurements alongside a larger number of corrections found.  The
 ``app_examples`` CTest ``surface_code-1-local-test-distance-3-inproc-rpc``
-wraps this flow (it sets ``CUDAQ_QEC_REALTIME_MODE=inproc_rpc`` and
-``EXTRA_CLI_ARGS=--use-relay-bp``).
+wraps this flow (it sets ``CUDAQ_QEC_REALTIME_MODE=inproc_rpc``).
```
bmhowe23, line 413: inproc_rpc is useful for our own internal testing but should not be referenced in our docs because it is really too niche for users to care about.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
bmhowe23: I think I agree. This is something that #672 brought in to improve the error rates.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
eliotheinrich: It does, yes; this leaves room for readout of the per-logical observable values up to 32 logical qubits, since there are only 64 bits to work with. This is a semantic change from the original; I can change it back if we prefer the other way.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
eliotheinrich: Ah, good point, thanks. I've added exceptions to error paths in [b191511](https://github.com/NVIDIA/cudaqx/pull/685/commits/b1915119985aa9b7e06ad378c51d012342cc09c4).

bmhowe23 (commented):
This is gonna be great. I have 1 nit and 1 +1 on one of Melody's comments.

Note: I think GitHub is still showing my BP+OSD comment; I tried to delete it, but I started to make the comment and then didn't hit submit, and now that section of code is gone (good), so just ignore that one.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
melody-ren: This can go into a clean-up pass later

libs/qec/unittests/realtime/app_examples/CMakeLists.txt
melody-ren: This can be discussed later

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
melody-ren: Oh right. Forgot you wrote about it in the description. Agree with the semantic change. Closing now.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
melody-ren: This can go into a clean up pass later

docs/sphinx/examples_rst/qec/realtime_relay_bp.rst
bmhowe23: I see this was pre-existing (inproc_rpc), so this can wait until later. I'll create a Jira ticket for that.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
bmhowe23: Addressed in e63925ada3f7020e16c4142ec6aac3c6271d7ee0. Marking resolved.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
bmhowe23: Opened #704. Marking this thread as resolved.

libs/qec/unittests/realtime/app_examples/surface_code-1.cpp
bmhowe23: Opened #704. Marking this thread as resolved.

libs/qec/unittests/realtime/app_examples/CMakeLists.txt
bmhowe23: Opened #704. Marking this thread as resolved.

bmhowe23 (approved):

PR #687, opened 2026-07-13
Enforce decoder GPU ownership in the realtime server

kvmto:
## Description

Depends on #664.

This follow-up completes the realtime decoder-server integration for the `cuda_device_id` placement contract introduced by #664. It transfers device ownership from decoder construction to the runtime threads and resources that actually use each decoder.

Because the base PR comes from a fork, this draft temporarily targets `main` and includes the #664 changes until that PR merges. The follow-up-only delta is available at https://github.com/kvmto/cudaqx/compare/decoder-gpu-pinning...decoder-server-gpu-pinning.

The change:

- propagates `cuda_device_id` through YAML, Python bindings, and realtime decoder factory parameters;
- pins dedicated decoding-server workers before they accept work;
- selects the correct device in shared realtime host dispatch;
- creates and releases CUDA graph and scheduler resources on the owning device;
- destroys graph resources and decoder instances on the owning device;
- aligns GPU RoCE placement with decoder placement, inherits `HOLOLINK_GPU_ID` when needed, and rejects conflicts;
- rejects ambiguous placement across graph-dispatch decoders; and
- adds YAML, Python, server-worker, teardown, shared-dispatch, and multi-GPU coverage.

Dedicated decoder workers select their CUDA device once during startup. Shared host-dispatch and direct realtime paths query the current device per request and call `cudaSetDevice` only when a switch is required. No new third-party dependencies are introduced.

## Testing

- Fresh CMake and Ninja `all` build: 493/493 steps passed.
- `clang-format --dry-run --Werror`: passed.
- YAPF diff check: passed.
- `git diff --check`: passed.
- Multi-GPU `CudaDeviceId` suite: 15/15 passed on two visible GPUs.
- Seven two-process and CQR tests that were blocked by the restricted network namespace passed outside the sandbox.
- Focused Python decoder YAML tests: 4 passed, 2 optional-dependency skips.

One unrelated Python app test fails locally because the installed CUDA-Q Python module does not expose `cudaq.measure_handle`; this PR does not modify that application. GPU RoCE-enabled compilation also remains for CI because local HSB and DOCA dependencies are unavailable.

### Before requesting review
- [ ] Full CI has been run.

### Scope and size

### Tests
- [x] Tests fail if the new functionality is broken, not just when it is missing.
- [ ] Negative tests added where exceptions are expected.
- [x] Truth data is not required for this lifecycle and placement change.
- [x] CI runtime impact considered; the added tests are focused and hardware-gated where necessary.

### Documentation
- [x] Public-facing APIs have Doxygen documentation.
- [ ] User-visible documentation is deferred to a separate follow-up PR.

### Code style
- [x] Naming follows the existing convention for each modified area.

### Dependencies


libs/qec/lib/realtime/decoding-server-cqr/DecodingServer.cpp
```diff
@@ -24,16 +25,44 @@ namespace cudaq::qec::decoding_server {
 
 using cudaq::qec::decoding::config::DecoderTransport;
 
+namespace {
+
+int resolve_gpu_roce_device_id(
+    const cudaq::qec::decoding::config::decoder_config &decoder) {
+  const char *raw_env_id = std::getenv("HOLOLINK_GPU_ID");
```
bmhowe23: I think that after all of these cuda_device_id PRs, we should - in principle - no longer need any `HOLOLINK_GPU_ID` environment variables. I've asked @melody-ren to start removing them as much as possible. (And if there's any reason we can't remove them, we likely need to discuss.)

bmhowe23:
@kvmto - do you know which parts of this are still needed now that #690 and #678 have landed?

melody-ren:
> @kvmto - do you know which parts of this are still needed now that #690 and #678 have landed?

Please note that I posted #698 to use a single source of truth for pinning cuda device

kvmto:
@melody-ren is this PR needed?
Is the content already covered?

melody-ren:
> @melody-ren is this PR needed? Is the content already covered?

I picked the critical bug fix from this PR and included it in #678. But a lot has changed since then and I'm not sure how much in this PR is still update to date

kvmto:
If this is obsolete we can close it.

PR #696, opened 2026-07-15
Replace msm execution contexts with cudaq::dem_from_kernel in app examples

bmhowe23:
The `surface_code-2` and `surface_code-3` C++ app examples built their detector error models by running the demo kernels under the `"msm_size"`/`"msm"` execution contexts and hand-assembling the detector and observable matrices
from the raw measurement syndrome matrix. The `cudaq::dem_from_kernel` API supersedes that: the demo kernels now annotate themselves with `cudaq::detector`/`detectors` and `cudaq::logical_observable` when a `declare_detectors` flag is set (detectors pair each round against the previous one, with the lock-in round as the reference; prep0's deterministic first-round Z stabilizers are declared as singles where the old code used them), and the hosts parse the returned Stim DEM text with `dem_from_stim_text`. The runtime `D_sparse` now comes straight from the analysis' measurements-to-detectors
(`m2d`) map instead of `generate_timelike_sparse_detector_matrix`, remapped into each decoder's own enqueue stream for `surface_code-3`'s split Z/X decoders (which run one `dem_from_kernel` pass per decoder via `declare_detectors_z`/`_x`
flags).

Notable constraint baked into the kernels:
- `dem_from_kernel` rejects kernels tagged `qubitMeasurementFeedback`, so the final data measurements are packed branch-free instead of via `to_integer(to_bools(...))`, whose discriminate-into-call flow sets the tag.

The now-unused `compute_msm`/`construct_mz_table` Python bindings (`py_decoder.cpp`, and their aliases in `cudaq_qec/__init__.py`) are removed. No other code references them after `main`'s `surface_code-1` refactor.

**Note on the library API:** `main` now offers `decoder_context_from_memory_circuit(...)` (with `x_component()`/`z_component()`/ `full_component()`), which `surface_code-1` uses. `surface_code-2/3` keep the hand-rolled `dem_from_kernel` path because that library circuit models noise only through the attached `noise_model` (gate channels), whereas these two examples deliberately demonstrate **in-kernel SPAM** (`apply_noise` data-qubit depolarization at `p_spam`); `surface_code-3` additionally applies corrections as in-kernel transversal gates rather than the library's observable-flip model. Migrating them would change the modeled physics and measured logical error rate,
so it is intentionally out of scope here.


libs/qec/python/bindings/py_decoder.cpp
```diff
@@ -1233,50 +1232,6 @@ void bindDecoder(nb::module_ &mod) {
         [Internal] Load local simulation realtime decoder library.
       )pbdoc");
 
-  qecmod.def(
```
bmhowe23, line 1236: This was never intended to be a public API. It was not documented, so it is safe to remove without deprecation warning. Also, the new preferred way is to use `dem_from_kernel` from CUDA-Q anyway.

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
```diff
@@ -243,18 +282,25 @@ demo_circuit_qpu(bool allow_device_calls,
...
+      std::vector<cudaq::measure_result> no_prev(0);
+      auto syndrome = custom_memory_circuit_stabs(
           subData, subXstab_anc, subZstab_anc,
           /*numRounds=*/1, cnot_schedX_flat, cnot_schedZ_flat,
           /*enqueue_syndromes=*/allow_device_calls,
-          /*do_errors_after_non_last_rounds=*/false, p_spam, i);
+          /*do_errors_after_non_last_rounds=*/false, p_spam, i,
+          declare_detectors, no_prev);
+      if (i == 0)
+        lockin_syndrome = syndrome;
```
eliotheinrich: This sets the lockin syndrome for all logical qubits equal to the lockin syndrome of logical qubit 0. The lockin syndrome should be stored per-logical.

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
```diff
@@ -192,6 +212,19 @@ __qpu__ void custom_memory_circuit_stabs(
...
+        // own; the X stabilizers only project a random frame.
+        for (std::size_t k = 0; k < zstab_anc.size(); ++k)
+          cudaq::detector(combined_syndrome[k]);
+      }
+    }
+    prev = combined_syndrome;
+    have_prev = true;
 #if PER_SHOT_DEBUG
     debug_print_syndromes(syndrome_x_int, syndrome_z_int);
 #endif
```
eliotheinrich: This is a compile error if PER_SHOT_DEBUG; syndrome_x/z_int are not defined; pre-existing, but should probably be removed.

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
```diff
@@ -192,6 +212,19 @@ __qpu__ void custom_memory_circuit_stabs(
       cudaq::qec::decoding::enqueue_syndromes(
           /*decoder_id=*/logical_qubit_idx, combined_syndrome);
     }
+    if (declare_detectors) {
+      if (have_prev) {
+        cudaq::detectors(prev, combined_syndrome);
```
eliotheinrich: I believe detector annotations compile to no-ops on hardware targets, so I don't think these `declare_detector` flags (similar in the surface_code-3.cpp example) are needed.

eliotheinrich (commented):
Looks good!

Removing the `declare_detectors` flags would be good, since the detector annotations should be automatically removed during compilation.

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
melody-ren, line 2: Since we are updating these two files, let's also add a blurb about what each of the two example demonstrate. Like this one here: https://github.com/NVIDIA/cudaqx/blob/578242d0d5eb44ce36d7c466bf395672f07c2898/libs/qec/unittests/realtime/app_examples/surface_code-1.cpp#L11

melody-ren, line 1: Is this note still up to date? https://github.com/NVIDIA/cudaqx/blob/578242d0d5eb44ce36d7c466bf395672f07c2898/libs/qec/unittests/realtime/app_examples/surface_code-2-test.sh#L56

MR24 was merged

libs/qec/unittests/realtime/app_examples/surface_code-3.cpp
melody-ren, line 1: Same comments about the blurb and the walkaround comments as in surface code 2

melody-ren (commented):
Looks good. Thanks! Just some notes about blurbs/stale looking comments

Should these two examples be referenced in the docs too? Currently they are not

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
bmhowe23: Done for sc-2 in `1cac72d6` — always declaring now; confirmed the annotations are stripped on the emulate path (all tests still pass). Kept the flags in sc-3, though: there they aren't just no-op gating — `declare_detectors_z`/`_x` select which stabilizer type each of the two `dem_from_kernel` passes declares, so dropping them would collapse both DEMs into one.

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
bmhowe23: Fixed in `1cac72d6` — merged the lock-in / inject / rounds loops so each logical qubit uses its own lock-in syndrome as the reference round. Same fix applied to sc-3.

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
bmhowe23: Good catch — removed the dead `debug_print_syndromes` call in `1cac72d6`. (sc-3 has no such call site.)

libs/qec/unittests/realtime/app_examples/surface_code-2.cpp
bmhowe23: Added a short blurb to both sc-2 and sc-3 in `1cac72d6` describing what each demonstrates.

bmhowe23: Removed in `1cac72d6` — MR24 is merged and the workaround was already commented out, so I deleted the note from both test scripts.

libs/qec/unittests/realtime/app_examples/surface_code-3.cpp
bmhowe23: Done in `1cac72d6` — blurb added and the stale MR24 note removed here too.

bmhowe23:
Re: referencing these in the docs — leaving them out for now; they're realtime-decoding test examples rather than user-facing tutorials.

melody-ren (approved):

PR #697, opened 2026-07-15
fix: To Fix PyMatching decoding for observable indices above 31

kaiqiy-nv:
This PR tries to fix [[B] 6458564](https://nvbugspro.nvidia.com/bug/6458564) .
It also enhances one test case to expose this bug and verify the fix.
Thank you for reviewing this PR.

The observable mask now uses a 64-bit unsigned shift, preventing undefined behaviour and incorrect decoding for high observable indices.

We have also changed the test case `Decodes64ObservablePath` a little bit.
Originally, it used exactly 64 observables, but the faulty packed-mask branch is only entered for fewer than 64 observables, so the bug was not exposed.
Now, the renamed test covers both 33 and 64 observables. The 33-observable case tests bit 32 in the packed-mask path, while the 64-observable case retains coverage of the vector path. It also asserts the exact flipped observable rather than merely checking that results are binary.


bmhowe23 (approved):
Thanks!

PR #711, opened 2026-07-20
fix: To fix decoder lifetime during async decoding

kaiqiy-nv:
This PR tries to fix [[B] 6479862](https://nvbugspro.nvidia.com/bug/6479862).
It also adds 1 C++ test and 1 Python test to expose the bug and verify the fix.

In this suggested fix:
[1] We keep the decoder alive for the full asynchronous operation.
We make `decoder` support `shared_from_this()`, return `shared_ptr` from public decoder factories, and capture shared ownership in `decode_async` instead of raw `this`.
Plugin creators keep `unique_ptr` and convert ownership only at the factory boundary.
[2] We update decoder storage and Python bindings to preserve shared ownership.
This avoids use-after-free when callers release a decoder before retrieving its asynchronous result.

We also add two test cases to expose this bug and verify the fix:
- `DecoderAsyncLifetime.PendingAsyncKeepsDecoderAlive`: We block an instrumented decoder inside `decode()`, release the caller’s decoder handle, and check that the destructor has not run. We then unblock the worker and verify its result.
- `test_decode_async_temporary_keeps_decoder_alive`: We call `decode_async` on a temporary Python decoder and verify that `.get()` returns a valid result, proving that the temporary remains alive until decoding completes.

### DIFF
--- a/libs/qec/include/cudaq/qec/decoder.h
+++ b/libs/qec/include/cudaq/qec/decoder.h
@@ -135,7 +135,8 @@ public:
 /// decoder.
 class decoder
     : public cudaqx::extension_point<decoder, const decoder_init &,
-                                     const cudaqx::heterogeneous_map &> {
+                                     const cudaqx::heterogeneous_map &>,
+      public std::enable_shared_from_this<decoder> {
 private:
   struct rt_impl;
   struct rt_impl_deleter {
@@ -185,8 +186,10 @@ public:
   /// @brief Decode a single syndrome
   /// @param syndrome A vector of syndrome measurements where the floating point
   /// value is the probability that the syndrome measurement is a |1>.
+  /// The decoder must be shared-owned, normally via `decoder::get()`.
   /// @returns std::future of a vector of length `block_size` with soft
   /// probabilities of errors in each index.
+  /// @throws std::runtime_error if the decoder is not shared-owned.
   virtual std::future<decoder_result>
   decode_async(const std::vector<float_t> &syndrome);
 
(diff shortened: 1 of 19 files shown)


bmhowe23 (commented):
Thanks, @kaiqiy-nv. We are close to a code freeze. Since this changes interfaces and the problematic use case appears to be not a primary use case, I'd like to hold off on this one for at least this week. Feel free to poke us about this in an about a week, though. Thanks!

kaiqiy-nv:
Hi @bmhowe23 , shall we look at this fix now?

bmhowe23:
I am hesitant to convert all unique_ptr's to shared_ptr's because I think unique_ptr's are cleaner. While the bug could - in principle - affect C++ users, this is a common C++ pattern that could be covered with a simple docs update instructing users to not destroy objects while still waiting on previously issued decode_async calls. The bigger case seems to be Python, and I think it could be fixed with this 1 change instead:

```cpp
.def(
    "decode_async",
    [](decoder &dec,
       const std::vector<float_t> &syndrome) -> async_decoder_result {
      // Release the GIL while launching asynchronous work.
      nb::gil_scoped_release release;
      return async_decoder_result(dec.decode_async(syndrome));
    },
    "Asynchronously decode the given syndrome", nb::arg("syndrome"),
    nb::keep_alive<0, 1>())  // (NEW 1 LINE CHANGE) result keeps the decoder alive
```

If you're ok with this, I'd propose replacing all the C++ API changes with the one-line `nb::keep_alive<0, 1>()` change above.

kaiqiy-nv:
Hello @bmhowe23 ,
Yes I agree. I have reverted the C++ API to `unique_ptr` and dropped the C++ test.
The net change now is the one-line `nb::keep_alive<0, 1>()` plus the test case `test_decode_async_temporary_keeps_decoder_alive`, which runs `decode_async` on a temporary decoder. Without the fix, that aborts or deadlocks at `.get()`; with it, 63 Python and 51 C++ tests pass.

bmhowe23 (approved):

PR #728, opened 2026-07-22
(docs) Updated sliding window and measurement->detector docs (Follow up #636, #671)

eliotheinrich:
## Description
This PR provides some spot-cleaning to the sliding window decoder and C++/Python API to reflect the fact that `generate_timelike_sparse_detector_matrix` is no longer the standard way of extracting memory circuit m2d matrices, and the syndrome format output by memory circuits now includes boundary detectors. A handful of documentation files are affected, but the primary effect is to update the examples to use the new format and API.


docs/sphinx/api/qec/cpp_api.rst
```diff
@@ -124,8 +126,6 @@ rejects dimensions whose products overflow ``std::size_t``.
 .. doxygenfunction:: cudaq::qec::dense_to_sparse(const cudaqx::tensor<uint8_t> &)
 .. doxygenfunction:: cudaq::qec::generate_random_pcm(std::size_t, std::size_t, std::size_t, int, std::mt19937_64 &&);
 .. doxygenfunction:: cudaq::qec::generate_random_pcm_sparse(std::size_t, std::size_t, std::size_t, int, std::mt19937_64 &&);
-.. doxygenfunction:: cudaq::qec::generate_timelike_sparse_detector_matrix(std::uint32_t num_syndromes_per_round, std::uint32_t num_rounds, bool include_first_round = false)
```
melody-ren, line 127: Thanks Eliot for doing a clean-up pass of the docs. Regarding the not-the-preferred APIs, we should not be removing the doc entries since the code/API itself is still well and alive, and being used throughout the code base. The c++ and python doc entries for the still in-use APIs should be restored.

melody-ren (approved):

docs/sphinx/api/qec/cpp_api.rst
eliotheinrich: Gotcha, thanks for the note; resolved in 6060c63!

PR #759, opened 2026-07-31
Dynamic DEM Initial Infrastructure

tlshannon:
### Dynamic DEM construction for streaming decoders

**`css_code_matrices` / `css_noise_params`** (`code_matrices.h`) — types that encode a CSS code's parity-check and logical-operator matrices plus a depolarizing (optionally phenomenological) noise model as plain sparse-binary structs, decoupled from the `code` base class.

**`dem_from_css_matrices()`** (`dem_construction.h`) — builds a T-round `detector_error_model` directly from CSS matrices and noise, no stabilizer circuit or Stim round-trip required. A `code`-object overload (`dem_construction_code.h`) wraps the matrix version for callers that already have a `code` instance.

**`seam_id` / `phase_id`** (`extended_dem.h`) — lightweight `uint32_t` identifiers whose value is the FNV1a-32 hash of a name string, computed at compile time. The same string always produces the same ID; no manual numbering required. A runtime name registry (`seam_id::register_name` / `seam_id::name()`) provides human-readable error messages. Standard memory-experiment names:

```cpp
seam_name::prev_round   // incoming syndrome boundary
seam_name::next_round   // outgoing syndrome boundary
phase_name::dem_init    // string "init"
phase_name::dem_bulk    // string "bulk"
phase_name::dem_final   // string "final"
```

**`extended_dem`** (`extended_dem.h`) — a single H matrix holding all detector rows, plus a flat list of named seam descriptors that identify which row bands participate in stitching. Interior rows are the complement. O (observables) is separate.

```cpp
struct extended_dem {
  sparse_binary_matrix H;           // seam bands + interior rows
  sparse_binary_matrix O;           // observable-flip rows
  std::vector<double> error_rates;
  std::vector<seam> seams;          // {id, row_begin, row_end}
  std::vector<uint64_t> tags;
};
```

Supports an explicit stitching algebra:
- `extended_dem_from_css_matrices()` — build a one-round chunk.
- `dem_stitch(a, b, from_seam, to_seam)` / `dem_stitch_all(chunks)` — left-fold chunk composition, contracting named seam row bands.
- `dem_merge_duplicate_columns()` — collapse fault columns with identical support using either exact XOR-probability or linear-sum combination.
- `dem_close(dem, to_seam)` / `dem_close_all(chunks)` — convert a stitched chunk to a flat `detector_error_model` in O(T) time.
- `dem_chunks_spec` / `dem_chunk_spec` — declarative named phase specs (YAML-serialisable); `num_rounds` is optional and absent for streaming.
- Streaming helpers: `dem_chunks_to_pcm()`, `dem_chunks_to_o_sparse()`, `dem_chunks_to_d_sparse()`, `dem_chunks_to_detector_round()` — produce the exact inputs `decoder::set_O_sparse()` / `set_D_sparse()` expect, so no decoder plugin needs to know about `extended_dem` internals.

**`decoder_config` additions** (`realtime/decoding_config.h`) — the flat-form fields (`H_sparse`, `O_sparse`, `D_sparse`, `block_size`, `syndrome_size`) are complemented by an optional `dem_chunks` field. `num_rounds` moves inside `dem_chunks` and is optional (omit for streaming). `expand_dem_chunks()` rewrites a chunk-form config into the equivalent flat form at decoder construction time, so the rest of the realtime pipeline is unchanged.

---

### New YAML format

Decoders may now be configured with a compact phase description instead of pre-expanded flat matrices. The round count is optional — omit it for streaming decoders where the total is not known at configuration time.

```yaml
dem_chunks:
  seam: {from: next_round, to: prev_round}
  connections:
    - {from: init, to: bulk}
    - {from: bulk, to: bulk}     # self-loop → repeating phase
    - {from: bulk, to: final}
  num_rounds: 5                  # omit for streaming
  phases:
    - name: init
      spec:
        num_faults: 9
        H_sparse: [0, 1, 5, -1, 1, 2, 6, -1, 2, 3, 7, -1, 3, 4, 8, -1]
        O_sparse: [0, -1]
        error_rates: [0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02]
    - name: bulk
      spec:
        num_faults: 9
        H_sparse: [0, 1, 5, -1, 1, 2, 6, -1, 2, 3, 7, -1, 3, 4, 8, -1]
        O_sparse: [0, -1]
        error_rates: [0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02]
    - name: final
      spec:
        num_faults: 5
        H_sparse: [0, 1, -1, 1, 2, -1, 2, 3, -1, 3, 4, -1]
        O_sparse: [0, -1]
        error_rates: [0.02, 0.02, 0.02, 0.02, 0.02]
```

The `connections` list encodes the phase graph. A self-loop (`from: bulk, to: bulk`) identifies the repeating phase; `num_rounds` controls how many copies are inserted. `H_sparse` at chunk level is the memory-experiment shorthand (all seams share the same rows); a `seam_specs` list gives per-seam rows for non-memory circuits.

`expand_dem_chunks()` is called during `finalize_decoders()` and rewrites the chunk form into flat fields in-place; all downstream decoder construction sees the familiar flat representation. A config that already carries a nonempty `H_sparse` is left untouched.

---

### Future PR: decoder plugin updates

This PR does not modify any decoder plugin. A follow-up PR will wire `dem_chunks` awareness directly into individual decoders where it adds value.

### Lines of code

This PR exceeds 1000 lines due new functionality and its test suite.  Tests are roughly half the line count.

N/A

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
- [x] Tests fail if the new functionality is broken, not just when it is missing.
- [x] Truth data added where simple \`EXPECT_*\` checks are insufficient.
- [x] CI runtime impact considered.

### Documentation
- [x] User-visible behavior changes have public docs.

### Code style

### Dependencies


libs/qec/include/cudaq/qec/code_matrices.h
```diff
@@ -0,0 +1,231 @@
...
+  /// n_checks (= hz.num_rows() + hx.num_rows()); Z-checks first.
+  std::vector<double> pm_per_check;
+};
+
+// ---------------------------------------------------------------------------
+// Implementation details used by dem_construction.cpp and extended_dem.cpp.
+// Not part of the public API.
+// ---------------------------------------------------------------------------
+
+namespace detail {
```
bmhowe23: please move this into a `lib/code_matrices.cpp` file so that this file becomes struct/class declaration only (not implementations). If there are private (non-user) interfaces, please move them into a new .h file that isn't in the top-level `include` directory.

libs/qec/lib/CMakeLists.txt
```diff
@@ -43,7 +43,10 @@ set(DECODERS_SOURCES
   decoder.cpp
   decoder_config_payload.cpp
   decoder_config_schema.cpp
+  dem_chunks_memory.cpp
```
bmhowe23, line 46: I'm wondering if we should make this a new .so (or a .a file that we eventually link into the -decoders.so file. What do you think about this?
```
cudaq-qec-dem (OBJECT):
  sparse_binary_matrix.cpp
  code_matrices.cpp
  detector_error_model.cpp     # impl uses stim
  pcm_utils.cpp                # sparse-matrix/PCM manipulation — this is DEM math too
  extended_dem.cpp
  dem_construction.cpp
  dem_chunks_memory.cpp
  └─ PRIVATE: libstim, cudaqx-core (tensor)
```

libs/qec/lib/extended_dem.cpp
```diff
@@ -0,0 +1,1538 @@
...
+// seam_id name registry
+// ---------------------------------------------------------------------------
+
+static std::unordered_map<uint32_t, std::string> &seam_name_registry() {
+  static std::unordered_map<uint32_t, std::string> reg;
+  return reg;
+}
+
+void seam_id::register_name(seam_id id, std::string_view name) {
+  seam_name_registry().emplace(id.value, name);
```
bmhowe23: emplace silently no-ops when the key exists - so if two different names ever hash to the same 32 bits, they alias into one seam/phase with no diagnostic, and the failure mode is wrong stitching, not an error. Please update to throw an error if that is encountered.

bmhowe23 (commented):
Thanks, Tracy. I think this is looking great. Just a few comments below.

libs/qec/include/cudaq/qec/code_matrices.h
tlshannon: Fixed

libs/qec/lib/extended_dem.cpp
tlshannon: Added checks and error throw.

libs/qec/lib/CMakeLists.txt
tlshannon: I created a .a file that is linked in -decoders.so

tlshannon: Be aware detector_error_mode.cpp uses logger.cpp.  It is fine when included in the -decoders.so, but if we need to migrate it out farther, we will have to make updates.

bmhowe23: Acknowledged. Thanks.

bmhowe23 (approved):
Thanks, Tracy.

PR #765, opened 2026-08-04
Lift O/D into construction and enable Chromobius on decoding server

melody-ren:
## This PR aims to:
1. Introduce a single source of truth `decoder_inputs` for O/D at construction time
2. Add support for raw Stim DEM on decoding server path via YAML config, such that decoders which need a raw DEM can be constructed
3. Remove the reliance on call order of setters to ensure O and D are configured
4. Remove the implication that O's presence means the decoder is requested to return observables instead of errors

## Bug fix
When parallel H columns share a matching edge, the error frame must name one of them. This branch names the column whose parameters the graph actually holds after the merge — KEEP_ORIGINAL and INDEPENDENT keep the first column's observables, REPLACE adopts the last, SMALLEST_WEIGHT adopts the smaller weight. Main named the last column unconditionally, which contradicts the retained edge for the first two strategies: under KEEP_ORIGINAL the matcher never uses the last column at all. Covered by PyMatchingDecoder.ErrorOutputTracksMergedParallelEdgeColumn.

> [!IMPORTANT]
> **Update:** A similar behavioral fix was later merged into `main` in commit [`4f31707e`](https://github.com/NVIDIA/cudaqx/commit/4f31707e1f0065ec1de4f66c96ea1da517c2df1e) on August 27, 2026. 
>
> This branch deliberately retains its own implementation. Although both fixes produce the same correct PyMatching merge-strategy behavior, this version is integrated with the broader design of this PR: model data and error rates come from the immutable `decoder_init`, and the requested output form is selected explicitly at construction rather than inferred from the presence of an observable matrix. This PR retains main’s subsequent PyMatching scratch-buffer reuse and concurrent-decode protection from `3fe047b3`.

### PyMatching result selection is now explicit

Supplying an observable matrix `O` no longer causes PyMatching to return
observable corrections. `O` is model data and may be supplied for purposes
other than selecting the decoder’s public result type.

When no output type is specified, PyMatching now returns an error frame. To
receive observable corrections, callers must request observable output
explicitly:

```python
decoder = cudaq_qec.get_decoder(
    "pymatching",
    H,
    O=O,
    output="observables",
    error_rate_vec=error_rates,
)
```

```cpp
auto decoder = cudaq::qec::get_decoder(
    "pymatching", std::move(inputs),
    cudaq::qec::decode_result_type::observables, params);
```

The former O-implies-observables behavior is intentionally not preserved.
Inferring API behavior from the presence of model data made it impossible to
supply O while requesting an error frame and coupled two otherwise
independent parts of the decoder contract.

Callers that want an error frame may omit the output argument or select it
explicitly:

```python
decoder = cudaq_qec.get_decoder(
    "pymatching", H, O=O, output="errors")
```
The realtime decoding server continues to request observable output explicitly.

## Out-of-tree C++ decoder migration

This PR changes the C++ decoder construction interface and ABI. Partner and
out-of-tree decoder plugins must be rebuilt and migrated:

- Accept decoder_init by value instead of receiving H separately.
- Pass decoder_init and the requested decode_result_type to the decoder
  base constructor.
- Read H, O, D, and error rates through get_inputs().
- Update registered creator functions to accept decoder_init,
  std::optional<decode_result_type>, and the decoder parameter map.
- Remove uses of set_O_sparse, set_D_sparse, set_result_type,
  on_o_sparse_configured, on_d_sparse_configured, and
  make_pcm_decoder.
- Select the plugin’s default output when no output type is requested and
  reject unsupported explicit output types during construction.

A detailed old-to-new API mapping is included in the decoder framework
documentation.

### TensorRT decoder migration

`trt_decoder` callers must now provide `engine_output_format` describing the
model tensor’s actual output:
- errors
- residual_detectors
- observables
- observables_and_residual_detectors
This is independent of the decoder’s requested public result basis.

### YAML compatibility

Existing user-facing YAML that places `error_rate_vec` under
`decoder_custom_args` remains accepted and remains the emitted wire format.
Input YAML also accepts the top-level `error_rate_vec` spelling. Supplying it
in more than one location, including nested global or inner decoder parameter
maps, is rejected instead of silently choosing one.

Measured against the merge base `674cb8f2`. PyMatching throughout, three repetitions per configuration.

**Realtime path (the one that matters here).** App and `decoding_server` as two processes over a UDP loopback wire. Run with `QEC_DECODING_SERVER_SPIN_US=0` so the server blocks rather than busy-polls and its CPU time measures decode plus transport rather than poll loops.

| distance 5, 5 rounds, 1000 shots (8000 decodes) | main | this branch |
|---|---:|---:|
| server CPU per decode | 262.5 / 265.0 / 263.8 µs | 256.3 / 257.5 / 256.3 µs |
| server peak RSS | 418.1 / 418.5 / 418.5 MiB | 417.9 / 417.9 / 418.1 MiB |

At distance 9 the two are indistinguishable (751.7 / 745.0 µs against 738.3 / 746.7 µs). Both branches find the same number of corrections, so they decode identically rather than doing different work. This configuration carries its model as matrices and exercises both O and D per shot, so it is the direct test of whether lifting them out of the setters cost anything: it did not.

**Resolving a model from a DEM** (`H = 2184 x 47129`, distance-13 surface code): parse drops from 75.5 / 97.0 / 75.2 ms to 11.7 / 11.2 / 11.3 ms, and retained memory from 105.9 MiB to ~5.9 MiB. Main materializes dense `detectors x mechanisms` tensors for H and O; this branch builds the sparse arrays directly from the hit lists the parser already has. A Chromobius-on-the-server configuration is DEM-sourced by definition, so it pays this cost on every reload.


libs/qec/lib/decoder.cpp
```diff
@@ -271,34 +343,17 @@ static void
 set_sparse_from_vec(const std::vector<int64_t> &vec_in,
```
tlshannon: I think this function is dead code since set_O/D_sparse was removed.

libs/qec/lib/decoder.cpp
```diff
@@ -247,10 +316,13 @@ dem_default_values dem_defaults_for_missing_keys(
 static uint32_t calculate_num_msyn_per_decode(
```
tlshannon: I think this function is dead code since set_O/D_sparse was removed.

melody-ren:
@justinlietz Just wanted to put this on your radar since this PR enforces that the information about the problem to be decoded (DEM or H/O/D etc) be known at construction time. It was done with the dynamic DEM and streaming work in mind such that the change introduced in this PR can be extended later to account for streaming DEMs. But I'm sure rough edges remain.

tlshannon (approved):
LGTM

libs/qec/include/cudaq/qec/decoder_init.h
```diff
@@ -0,0 +1,160 @@
...
+  const std::vector<double> &error_rates() const;
+  const std::optional<std::vector<std::size_t>> &error_ids() const;
+
+  /// @brief Return D, or nullptr when input syndromes are already detectors.
+  const sparse_binary_matrix *measurement_to_detectors() const noexcept;
+
+  /// @brief Return the same inputs without D, for a decoder that is fed
+  /// detectors rather than a raw measurement stream. Everything else,
+  /// including the authoritative source, is preserved.
+  decoder_init decoder_init_without_d() const;
```
bmhowe23, line 124: Per offline discussion, this line in particular is pointing at a bigger issue. Namely, the `D` matrix probably never should've been part of our standard `decoder` class. Let's split this PR into multiple chunks. One of the future PRs should be to make a decoding task graph that contains `D` rather than a decoder containing `D`. Once that is done, that would affect all `D` parts of this PR.

Per our discussion, the other pieces should probably be moved into individual PRs and merged independently of the aforementioned decoding task graph object.

bmhowe23 (commented):
Sorry, this comment is old. I thought I hit submit a while back. It doesn't have to block this PR but submitting it anyway per your request.

libs/qec/include/cudaq/qec/decoder_init.h
melody-ren: As discussed, I agree with the direction of the proposed change for D. Since a proper cleanup would involve a moderate amount of work in both public and private repos, I opt for deferring this to a follow up PR and keep the scope of PR765 as stated. 

Tracking this in #832

PR #755, opened 2026-07-28
Gamma ensemble performance docs

eliotheinrich:
## Description

This PR adds a documentation page demonstrating the performance tuning using the ensemble gamma version of Relay BP. It introduces a new .rst example, along with figures.

PR #766, opened 2026-08-04
Gamma ensemble benchmarking scripts

### LINKED ISSUE #758: Upload benchmarking scripts for gamma ensemble docs
https://github.com/NVIDIA/cudaqx/pull/755 introduced figures to the documentation using benchmarking data. We need to upload the scripts used for generating/plotting this data so that the figures can be reproduced as needed.

eliotheinrich:
## Description

This PR adds the standalone scripts used for benchmarking the Relay BP gamma ensemble feature. The docs were introduced in https://github.com/NVIDIA/cudaqx/pull/755. Closes https://github.com/NVIDIA/cudaqx/issues/758.

The BB memory circuit .stim circuits are stored in the repo as LFS objects, avoiding a ~100k line diff.


assets/benchmarks/bb144_memory_Z.stim
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
melody-ren, line 1: It seems like these `stim` files are byte identical to the original except that `DETECTOR` lines are removed. If that's the case, could you please use a script to download and strip the stim files with pinned SHA from the source, instead of committing them into LFS?

benchmarks/qec/relaybp_gamma_ensemble/run_sweep.py
```diff
@@ -0,0 +1,161 @@
+# ============================================================================ #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+# Reproduces the data behind the gamma-ensemble Relay-BP user guide
+# (docs/sphinx/examples_rst/qec/nv_qldpc_gamma_ensemble_user_guide.rst).
```
melody-ren, line 10: To be truly reproducible, I think we need to add the used Stim version and qec version (or SHA) somewhere.

benchmarks/qec/relaybp_gamma_ensemble/README.md
```diff
@@ -0,0 +1,51 @@
+# Relay-BP gamma-ensemble benchmark
```
melody-ren: I propose that we put the content of this file into the codebase docs and remove the readme here. That way, we only need to maintain one file/location for the instructions and walkthrough

melody-ren (commented):
Thanks @eliotheinrich ! Just my two cents

assets/benchmarks/bb144_memory_Z.stim
eliotheinrich: The goal was to remove the explicit external dependency for the benchmark, but I'm happy to do it this way as well. @bmhowe23 maybe you can weigh in?

benchmarks/qec/relaybp_gamma_ensemble/run_sweep.py
eliotheinrich: Fair enough, although I think the goal of adding these scripts is so that we can regenerate the data for future QEC versions. I've included the SHAs in the script in [51a5bcd](https://github.com/NVIDIA/cudaqx/pull/766/commits/51a5bcd0023b7616f6abc6fde1cad163ac53f6ba) for now, but maybe it makes sense to include these tags as a comment in the docs where the figures appear rather than in the benchmark scripts?

benchmarks/qec/relaybp_gamma_ensemble/README.md
eliotheinrich: Good point; addressed in [51a5bcd](https://github.com/NVIDIA/cudaqx/pull/766/commits/51a5bcd0023b7616f6abc6fde1cad163ac53f6ba)

assets/benchmarks/bb144_memory_Z.stim
melody-ren: That's a good point I haven't thought of. Is removing external dependency something we want to achieve in general?

benchmarks/qec/relaybp_gamma_ensemble/run_sweep.py
melody-ren: Agreed

assets/benchmarks/bb144_memory_Z.stim
bmhowe23: I think that only *some* of the `DETECTOR` lines are removed, not all, right? If that's the case, we just want to keep this reduced-detector version here without having a complicated script to modify the files during testing. Since there is so much overlap between the original files and these, we should at least add a README.md file to this directory and give proper attribution to where the original ones came from.

bmhowe23 (approved):
Thanks, @eliotheinrich. Approved pending the resolution of the last open thread.

assets/benchmarks/bb144_memory_Z.stim
eliotheinrich: Yes, that's correct. 

Ok, I've moved the usage docs back to a README.md in 0be797b. I think, since this is the only place the usage details are included and it, along with the scripts, are top-level citizens of a named directory, this should address Melody's comment in the thread below.

PR #767, opened 2026-08-05
[CI] Give each build_wheels.yaml instance distinct artifact names

bmhowe23:
## Problem

`nightly_tests.yaml` invokes `build_wheels.yaml` **twice** within a single run:

- `wheels` — builds against the pinned `.cudaq_version`
- `cudaq-main-wheels` — builds against CUDA-Q `main`

Both instances uploaded artifacts under **identical names**. Every wheel artifact
appears exactly twice per nightly run — all four `cudaq-wheels-{arch}-cu1{2,3}`,
all twelve `wheels-py3.*`, and `metapackages`. For example, in run
[30883444365](https://github.com/NVIDIA/cudaqx/actions/runs/30883444365):

| Artifact ID | Name | Uploaded | Built against |
|---|---|---|---|
| 8882160104 | `cudaq-wheels-amd64-cu12` | 06:17:34 | pinned, CUDA-Q `90b0f4d8` |
| 8882752498 | `cudaq-wheels-amd64-cu12` | 06:45:11 | main, CUDA-Q `7f0f524a` |

`actions/download-artifact` resolves by name, so the test jobs pick up whichever
instance finished last. The pinned CPU test job downloaded `8882752498` —
main's CUDA-Q wheels — while testing CUDA-QX wheels built against the pin.

The ABI mismatch then surfaces as an `ImportError` on whatever symbol happens to
differ between the two CUDA-Q revisions. In the run above, CUDA-Q
[#5005](https://github.com/NVIDIA/cuda-quantum/pull/5005) changed a
`quantum_platform` member that is called from inline header code, so it leaks
into every downstream `.so`:

libcudaq-qec.so: undefined symbol: _ZNK5cudaq16quantum_platform13validateQpuIdEm

Because this is a race between two upload jobs, it fails **sporadically** across
architectures and Python versions rather than consistently — and it is not new.
The Aug 1 nightly failed the same way on a different symbol
(`_ZN14cudaq_internal8compiler13compileModuleE...`), simply because that was the
symbol that differed at the time.

## Fix

Add an optional `artifact_name_suffix` input to `build_wheels.yaml` and append it
to every artifact name (4 uploads and their 8 matching downloads across the CPU
and GPU test jobs). `nightly_tests.yaml` passes `-cudaq-main` from the CUDA-Q
main instance, so the two instances produce disjoint sets:

cudaq-wheels-amd64-cu12              <- pinned (.cudaq_version)
cudaq-wheels-amd64-cu12-cudaq-main   <- CUDA-Q main

The pinned instance keeps the default empty suffix, so its artifact names are
**unchanged** — this matters because `pr_workflow.yaml` and the
`artifacts_from_run` path both reference these names.

## Notes for reviewers

- **`artifacts_from_run` requires a matching suffix.** It reuses a prior run's
  artifacts by name, so re-running against a nightly that used `-cudaq-main`
  means passing the same `artifact_name_suffix`. This is documented in the
  `workflow_dispatch` input description, and the input is exposed there so it is
  settable from the UI — but it is a manual step.
- **A related latent issue is not addressed here.** In
  `get-cudaq-wheels/action.yaml` (and `get-cudaq-build`), `restore-keys` is the
  full cache key minus the `-nightly` suffix, so suffixed and unsuffixed callers
  share a cache pool. It is harmless today because the CUDA-Q ref is part of the
  key, and it is unrelated to these failures.

## Testing

YAML validated for both files. Full verification requires a nightly run or a
manual `workflow_dispatch`; this has not yet been exercised end to end.

### DIFF
--- a/.github/workflows/build_wheels.yaml
+++ b/.github/workflows/build_wheels.yaml
@@ -42,7 +42,7 @@ on:
         required: false
       artifact_name_suffix:
         type: string
-        description: Optional suffix to append to every artifact name. Required when more than one instance of this workflow runs within a single parent run, so the instances do not overwrite each other's artifacts. Must match the value used by the run referenced by artifacts_from_run.
+        description: Optional suffix to append to the CUDA-Q wheel, CUDA-QX wheel and metapackage artifact names. Required when more than one instance of this workflow runs within a single parent run, so the instances do not overwrite each other's artifacts. Must match the value used by the run referenced by artifacts_from_run.
         required: false
   workflow_call:
     inputs:
@@ -85,7 +85,7 @@ on:
         required: false
       artifact_name_suffix:
         type: string
-        description: Optional suffix to append to every artifact name. Required when more than one instance of this workflow runs within a single parent run, so the instances do not overwrite each other's artifacts.
+        description: Optional suffix to append to the CUDA-Q wheel, CUDA-QX wheel and metapackage artifact names. Required when more than one instance of this workflow runs within a single parent run, so the instances do not overwrite each other's artifacts.
         default: ''
         required: false
     secrets:
@@ -115,7 +115,7 @@ jobs:
       - name: Upload release assets
         uses: actions/upload-artifact@v4
         with:
-          name: release-assets${{ inputs.artifact_name_suffix }}
+          name: release-assets
           path: '*.tar.gz'
           if-no-files-found: error
           retention-days: 1
@@ -205,7 +205,7 @@ jobs:
         if: ${{ inputs.assets_repo != '' && inputs.assets_tag != '' }}
         uses: actions/download-artifact@v4
         with:
-          name: release-assets${{ inputs.artifact_name_suffix }}
+          name: release-assets
           path: .
 
       - name: Get required CUDAQ version
--- a/.github/workflows/build_wheels.yaml
+++ b/.github/workflows/build_wheels.yaml
@@ -138,9 +138,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.11', '3.12', '3.13']
-        platform: ['amd64', 'arm64']
-        cuda_version: ['12.6', '13.0']
+        # TEMP(PR767-test): trimmed to one combination. Restore
+        #   python: ['3.11', '3.12', '3.13']
+        #   platform: ['amd64', 'arm64']
+        #   cuda_version: ['12.6', '13.0']
+        # before merging.
+        python: ['3.11']
+        platform: ['amd64']
+        cuda_version: ['12.6']
         toolchain:
           - cc: gcc-12
             cxx: g++-12
@@ -297,8 +302,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: ['amd64', 'arm64']
-        cuda_version: ['12.6', '13.0']
+        # TEMP(PR767-test): trimmed to one combination. Restore
+        #   platform: ['amd64', 'arm64']
+        #   cuda_version: ['12.6', '13.0']
+        # before merging.
+        platform: ['amd64']
+        cuda_version: ['12.6']
     # Use 32 CPUs rather than 8 (above) because we are only spawning one job per
     # platform rather than one job per Python version per platform.
     runs-on: ${{ startsWith(github.repository, 'NVIDIA/cudaqx') && format('linux-{0}-cpu32', matrix.platform) || 'ubuntu-latest' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_wheels.yaml
```diff
@@ -106,7 +115,7 @@ jobs:
       - name: Upload release assets
         uses: actions/upload-artifact@v4
         with:
-          name: release-assets
+          name: release-assets${{ inputs.artifact_name_suffix }}
```
anjbur: There's no harm in adding the suffix here, but it is unnecessary. This `fetch-assets` job only runs on manual release builds that download artifacts from a draft release. To run into the race condition being fixed by this PR, there would need to be a workflow that called this wheel workflow twice, each downloading assets from different draft releases.

If we want to append something here, I'd suggest instead using `inputs.assets_tag`, as that is the unique identifier for the asset being downloaded/uploaded by this job.

.github/workflows/build_wheels.yaml
```diff
@@ -196,7 +205,7 @@ jobs:
         if: ${{ inputs.assets_repo != '' && inputs.assets_tag != '' }}
         uses: actions/download-artifact@v4
         with:
-          name: release-assets
+          name: release-assets${{ inputs.artifact_name_suffix }}
```
anjbur: Same as my comment on line 118. These lines should match, but I'd suggest `inputs.assets_tag` if we want a suffix here.

anjbur (commented):
This looks good to me! For testing, I'd suggest a pared-down manual run of the nightly build. We can comment out unnecessary parts of the workflow (i.e. every non-wheel job), then confirm that the correct artifacts were downloaded by the correct jobs and that any failures are not coming from this race condition. We can also double check that the wheels uploaded by the PR CI don't contain a suffix--that should be sufficient to confirm that `artifacts_from_run` usage will be unaffected.

.github/workflows/build_wheels.yaml
bmhowe23: Addressed in f2feabea39f831e66f7ec2d55375a49a14737208. I omitted the suffix because as you say, it doesn't seem to be needed.

.github/workflows/build_wheels.yaml
bmhowe23: Addressed in f2feabea39f831e66f7ec2d55375a49a14737208. I omitted the suffix because as you say, it doesn't seem to be needed.

bmhowe23:
> This looks good to me! For testing, I'd suggest a pared-down manual run of the nightly build. We can comment out unnecessary parts of the workflow (i.e. every non-wheel job), then confirm that the correct artifacts were downloaded by the correct jobs and that any failures are not coming from this race condition. We can also double check that the wheels uploaded by the PR CI don't contain a suffix--that should be sufficient to confirm that `artifacts_from_run` usage will be unaffected.

Thanks - I took your suggestion and ran a pared-down nightly. Details below.

**Setup** ([run 31066167590](https://github.com/NVIDIA/cudaqx/actions/runs/31066167590)): trimmed the nightly to just the two `build_wheels.yaml` instances — dropped `build-cudaq-main` and the lib/docs jobs, cut the matrices to `py3.11 / amd64 / cu12.6`, and disabled the GPU tests. I also pinned `resolve-cudaq-main` to `7f0f524a09` rather than resolving `main`. That's the CUDA-Q ref from the 2026-08-04 nightly, which is post-[#5005](https://github.com/NVIDIA/cuda-quantum/pull/5005) and therefore ABI-incompatible with our pinned `.cudaq_version` (`90b0f4d8`) — so the run recreates the exact skew that produced the original `undefined symbol` failure. Whole thing took ~35 min. The temp commit is reverted.

**Result: green.** Six artifacts, six unique names — three matched pairs:

```
cudaq-wheels-amd64-cu12          cudaq-wheels-amd64-cu12-cudaq-main
metapackages                     metapackages-cudaq-main
wheels-py3.11-amd64-cu12         wheels-py3.11-amd64-cu12-cudaq-main
```

On 2026-08-04 every one of those names appeared **twice** with two competing IDs.

**Download routing** — each test job resolved all three of its downloads to its own instance's IDs. The pinned job (the one that failed on 08-04) took `cudaq-wheels-amd64-cu12` = `8953910594`, *not* the main-built `8954227802` that was sitting in the same run under the colliding name previously. No `undefined symbol` anywhere in the log; suites green at 309 / 64 / 3 passed.

**On your second point** — confirmed the PR CI wheels carry no suffix. From this PR's own run 31065493462: `cudaq-wheels-{amd64,arm64}-cu1{2,3}` and `metapackages`, all unsuffixed, each appearing once. `pr_workflow.yaml` doesn't pass `artifact_name_suffix`, so it takes the `''` default and names are byte-identical to before — `artifacts_from_run` against a PR run is unaffected.

anjbur (approved):
LGTM!

PR #768, opened 2026-08-05
Remove `inproc_rpc`

melody-ren:
Removes `inproc_rpc`, the in-process RPC session that was an intermediary step
in the decoding-server work. The external decoding server (HOST_CALL over UDP)
and the direct-call path are the surviving decode paths.

- Delete `qec_realtime_session` and `rpc_producer`, and the tests that existed
  only to exercise them (`test_pymatching_realtime`,
  `test_realtime_qldpc_graph_decoding` + `qldpc_config_loader`, the
  `surface_code-1` inproc CTest).
- Drop `CUDAQ_QEC_REALTIME_MODE` (now unread) and the `REQUIRE_HOST_MODE`
  assertions, which only proved the inproc session was in use.
- Docs: excise the `realtime_relay_bp.rst` sections that documented the removed
  path; re-anchor comments naming the deleted session.
- Stop linking `cudaq-qec-realtime-decoding` against
  `libcudaq-realtime-host-dispatch`; that link existed for
  `qec_realtime_session.cpp`, and neither remaining source names a
  host-dispatch symbol.

## relay-BP coverage

Every `surface_code-4-yaml` case previously ran under
`CUDAQ_QEC_REALTIME_MODE=inproc_rpc`. pymatching and trt fall through to the
direct-call path; the nv-qldpc/relay-BP cases were parked behind `if(FALSE)`.
This PR deletes them rather than restoring them.

`surface_code-4-yaml-mixed-dispatch` already runs nv-qldpc RelayBP through
`decoding_server` and asserts on its per-decoder `logical_errors` line, under
the same device-graph + proprietary-archive gate those cases would have
required — so restoring them adds no configuration that is not already
covered.

## Testing

42/42 `app_examples.surface_code-4-yaml` tests pass
with the nv-qldpc plugin, proprietary cudevice archive, and python-onnx
configured. This covers the surviving direct-call and external decoding-server
paths after removing `inproc_rpc`, including
`app_examples.surface_code-4-yaml-mixed-dispatch`, where nv-qldpc RelayBP runs
on a device-graph ring alongside a host-dispatched decoder.


cketcham2333 (approved):

bmhowe23 (approved):

bmhowe23:
Most of the CI is green. The only ones that aren't are the ones that haven't run yet (blocked by GPU runners). Those have worked on previous CI runs before the latest merge from main. I'm making the judgement call to merge anyway.

PR #769, opened 2026-08-06
[QEC] Decoding server: serve HOST_CALL requests inline; retire the worker threads, queues, and CqrTransceiver

cketcham2333:
## Summary

Retires the cross-thread request path in the QEC decoding server ("fix A" of the
decoding-server redesign). Every host RPC previously crossed two server threads:
shim → CqrTransceiver::inject (frame copy, token, pending map) → RpcDispatcher
re-parse/re-route → WorkItem copy → bounded queue → per-decoder worker thread →
decoder → transport.send → pending lookup → tx-slot memcpy → promise wakeup of
the parked dispatcher — two thread crossings and ~5 buffers per get_corrections.

Now the shim resolves the `DecodingSession` by the payload's `decoder_id` and
calls its `handle_*` method **inline on the CUDAQ dispatcher thread**: the rx
slot is parsed in place, the decoder runs on the calling thread, and the
response is written straight into the tx slot (magic release-stored last).
Zero thread crossings, zero copies, zero steady-state allocations — the same
shape as the GPU device-graph dispatch path and the legacy direct path
(`host::enqueue_syndromes`), whose validation ladder, capture hook, and pin
helper `handle_enqueue` now shares.

Net: **+960 / −2,702 lines.**

## Measured latency (udp two-process rig, d3 surface code, pymatching, 500 shots)

| RPC | before (median) | after (median) | p99 after |
|---|---|---|---|
| enqueue_syndromes | 0.8 µs | **0.10 µs** | 0.86 µs |
| get_corrections | 1.6 µs | **0.08 µs** | 0.10 µs |
| reset_decoder | 4.8 µs | **0.10 µs** | 0.10 µs |

(`get_corrections` no longer contains the decode wait — the decode runs inline
in the volume-completing enqueue, whose tail carries the actual decode time.)

## What's new

- **RpcSlot.h** — single home for in-place slot parsing and in-place response
  writing; deliberately device-portable (no allocation/exceptions/STL) so the
  GPU dispatch kernel can later compile the same source.
- **DecodingSession** — payload-level cores plus inline `handle_*` entry
  points; corrections pack directly into the tx slot (no truncation: an
  oversized result fails with INTERNAL_ERROR).
- **Debug-only single-caller guard** — a scoped in-flight assert catches the
  invalid two-rings-one-decoder topology loudly in debug builds and compiles
  out entirely in release (invalid-config guards must not add hot-path
  latency). Always-on enforcement is deliberately omitted: the HOST_CALL ABI
  carries no ring identity to validate against, and multi-sender serialization
  belongs to the multi-ring decoding-unit design.
- **pin_decode_device_cached** — thread-local-cached CUDA pin; pin failures
  surface at server bring-up via a create-time probe.
- **HopStats** — slimmed to single-thread stage stats (parse/decode/respond);
  `QEC_DECODING_SERVER_HOP_STATS` and the `total` metric keep their semantics.

## What's deleted (~1,500 LOC)

- `CqrTransceiver.h` (singleton transceiver: inject, tokens, pending map,
  promises, inbox, frame builders)
- `RpcDispatcher` + `ResponseWriter` (second-stage re-parse/re-route)
- `DecodingSession` worker machinery (WorkItem, bounded queue, worker thread,
  `syndromes_dropped` latch — backpressure is the transport ring; the
  SYNDROMES_DROPPED/BUSY codes stay reserved on the wire but are no longer
  emitted)
- `RoundAccumulator` + `SyndromeMappingTable` (unreachable fragment-merge
  machinery; the `syndrome_mapping_id` wire field stays — id 0 accepted,
  others rejected with BAD_REQUEST)
- `SpinPolicy.h` + `QEC_DECODING_SERVER_SPIN_US` (the cross-thread waits they
  optimized no longer exist)
- `LoopbackTransceiver`, the recv loop, `install_dispatch_sink`

## Behavior notes

- Enqueue response moves to the handler tail; client-visible timing is
  unchanged (the tx flag publishes on handler return). The deferred-error
  contract is unchanged.
- A volume-completing enqueue runs its decode inline and blocks only its own
  ring: **decoder parallelism comes from per-decoder rings** (one dispatcher
  thread each). A shared ring still routes multiple decoders correctly but
  serializes their decodes.
- Test-harness updates accordingly: `surface_code-4-yaml-test.sh` exports the
  per-decoder ring ports (all tokens required, fail-fast), and
  `hard-patch-pymatching` runs 500 shots so its opportunistic overlap evidence
  is statistically certain (the deterministic proof is the three-way-overlap
  barrier test).
- `run_realtime_decoding.sh`: fixes a hololink→hsb rename leftover
  (`--hololink` → `--hsb-ip`) needed for `--source fpga` runs.


libs/qec/lib/realtime/decoding-server-cqr/decoding_server_cqr.cpp
```diff
@@ -119,46 +127,63 @@ static void write_error_response(const void *rx_slot, void *tx_slot,
...
+                           static_cast<int32_t>(RpcStatus::BAD_REQUEST));
+      return;
+    }
+    uint64_t decoder_id = 0;
+    if (!slot::peek_decoder_id(rx_slot, slot_size, decoder_id)) {
+      write_error_response(rx_slot, tx_slot, slot_size,
+                           static_cast<int32_t>(RpcStatus::BAD_REQUEST));
+      return;
+    }
+    DecodingSession *session = registry->find(decoder_id);
```
vedika-saravanan, line 168: I may be overlooking an existing topology guarantee, but is explicit enforcement of the current one-ring-per-decoder assumption planned as a follow-up? The inline path now mutates DecodingSession directly, while the concurrent-entry check is debug-only. It would be helpful to document the current ownership constraint, or reject unsupported sharing clearly, as multi-ring support evolves.

vedika-saravanan:
Thanks Chuck Overall, this is a strong simplification of the host dispatch path. I left one small question about documenting the current one-ring-per-decoder ownership assumption as multi-ring support evolves.

libs/qec/lib/realtime/decoding-server-cqr/decoding_server_cqr.cpp
melody-ren: Some stale comments here: This PR removes CqrTransceiver

libs/qec/lib/realtime/decoding-server-cqr/CMakeLists.txt
melody-ren: Stale comment about CqrTransceiver

libs/qec/lib/realtime/decoding-server-cqr/RpcSlot.h
```diff
@@ -0,0 +1,226 @@
...
+    return false;
+  std::memcpy(&out, static_cast<const uint8_t *>(rx_slot) + sizeof(RPCHeader),
+              sizeof(uint64_t));
+  return true;
+}
+
+/// Write a header-only RPCResponse (no result payload) into \p tx_slot.
+/// The magic is release-stored LAST so the CUDAQ runtime sees a complete
+/// response before observing the magic word.
+inline void write_response(void *tx_slot, uint32_t request_id,
```
melody-ren, line 182: You probably already have a plan for this one, but here's a function that does nearly the same thing: https://github.com/NVIDIA/cudaqx/blob/ec5fc992bf3b1622fce7eee42734f3f632af8d42/libs/qec/lib/realtime/decoding-server-cqr/decoding_server_cqr.cpp#L108

I could be mistaken though and they might be serving different purposes

melody-ren (commented):
Thanks Chuck! Looks good to me. Just some minor nits

libs/qec/lib/realtime/decoding-server-cqr/decoding_server_cqr.cpp
cketcham2333: Resolved in 30c64ed3e2d6ef503c3b425682889a929376ee42

libs/qec/lib/realtime/decoding-server-cqr/CMakeLists.txt
cketcham2333: Resolved in 30c64ed3e2d6ef503c3b425682889a929376ee42

libs/qec/lib/realtime/decoding-server-cqr/RpcSlot.h
cketcham2333: Resolved in 30c64ed3e2d6ef503c3b425682889a929376ee42

libs/qec/lib/realtime/decoding-server-cqr/decoding_server_cqr.cpp
cketcham2333: Good question. The constraint is "one consuming thread per session," and every supported topology satisfies it structurally — even a shared ring serves its decoders from one dispatcher thread. Two rings feeding one decoder is unsupported on deeper grounds than the race: without sender identity in the decoder API, interleaved rounds from two senders decode wrong regardless of locking.

Release-mode enforcement isn't possible today — the HOST_CALL ABI (`host_fn(rx, tx, slot_size)`) carries no ring identity to validate against; that needs a cudaq-side ABI addition, tracked with the de-globalization follow-up. And we're holding the line that invalid-config guards must cost zero on the release hot path, so detection is the debug-only concurrent-entry assert. The ownership rule is documented in both places a reader would look: the `DecodingSession` class comment / `debug_in_flight` member in DecodingSession.h, and the plugin's file-header comment in decoding_server_cqr.cpp.

To be explicit about the follow-up: it will NOT enforce one ring per decoder — it goes the other way. Rings become sender-scoped, so a decoder will have multiple rings (one per sender), with a single dispatcher thread per decoder round-robining its ring group. So one-ring-per-decoder is a today-only description, while the invariant that actually matters — exactly one consuming thread per session — is preserved by construction.

vedika-saravanan: Thanks, that makes sense. I had framed the current setup too narrowly as one ring per decoder, the important invariant is one consuming thread per session. The planned multi-ring design preserves that invariant, so this answers my question. Resolving this comment.

melody-ren (approved):

PR #781, opened 2026-08-10
[Core] Add rvalue insert overload to heterogeneous_map

bmhowe23:
- Adds a move-enabled insert(key, T&&) overload to heterogeneous_map that avoids an unnecessary deep copy when inserting rvalues (e.g. large vectors, nested containers)
- The overload uses SFINAE to restrict to rvalues only — lvalues continue to resolve to the existing copying overload, so the change is fully backward-compatible
- Adds four unit tests covering: move-from-vector, large payload move, overwrite-via-move, and lvalue-not-moved

Motivation

When inserting large payloads (e.g. syndrome history, LLR buffers) the existing copy-only insert forced an extra deep copy. The new overload lets callers pass std::move(large_object) to avoid that allocation.

### DIFF
--- a/docs/sphinx/conf.py.in
+++ b/docs/sphinx/conf.py.in
@@ -250,6 +250,8 @@ nitpick_ignore = [
     ('cpp:identifier', 'details'),
     ('cpp:identifier', 'spin_op'),
     ('cpp:identifier', 'heterogeneous_map'),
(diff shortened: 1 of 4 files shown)


melody-ren (approved):
Looks good. Thanks Ben!

PR #787, opened 2026-08-11
[QEC] Hardware CI for the realtime decoding server (NVQLink lab: DGX Spark + GB200)

cketcham2333:
## What

  A self-contained hardware CI runner (`docker/decoding-server/hw_ci/`) that
  exercises the realtime decoding server on real RDMA/FPGA hardware. Given a
  commit, it clones it, builds a layer-cached dev image
  (`dev.Dockerfile` on top of the pin-matched `cudaqx-dev` base), compiles
  cudaq-realtime + cudaqx + the demo binaries inside a privileged container,
  and runs the test lanes. Every lane ends in the summary table as PASS,
  FAIL, or a **named** SKIP, so lost coverage is always visible.

  ## Lane coverage

  - All four decoders (pymatching, multi_error_lut, nv-qldpc, trt) over
    **udp**, **cpu_roce** (two-process), and the **FPGA** syndrome source
  - FPGA **device-graph** GPU dispatch (incl. 64 KiB-host-page ring geometry)
  - `hsb-fpga-server` on both wires (cpu_roce + gpu_roce), the
    gpu-roce-qldpc-bridge cross-check, and the GB200-class GPU ctests
    unreachable in normal CI (qldpc-graph, mixed-dispatch)
  - **ising-prepare**: downloads the gated HF model and rebuilds the trt
    bundle on-GPU each run (`--hf-token-prompt` / `HF_TOKEN` / staged-bundle
    fallback for tokenless machines)

  ## SoftRoCE on DOCA-OFED hosts (GB200)

  Two gaps blocked `--roce-pair rxe` (the default) on hosts whose `ib_core`
  comes from DOCA/MLNX-OFED DKMS, and both are handled automatically:

  - the distro `rdma_rxe.ko` cannot bind OFED's `ib_core` (symbol CRC
    mismatch) and OFED dropped the rxe driver, so the image stages a
    3-line-patched upstream v6.17 rxe source (`hw_ci/rxe-ofed/`); when
    `modprobe` fails, the runner builds it against the host's `ofa_kernel`
    tree (mounted RO) and insmods it from the privileged container
  - Mellanox `ibverbs-providers` ships no rxe userspace provider and its
    provider ABI differs from Ubuntu's, so the image builds `librxe` from
    the matching Mellanox rdma-core source release (sha256-pinned)


melody-ren:
Thanks Chuck! Tried this out on GB200 and it works. 
But I think we have a concurrent user hazard. The container's name is derived from the commit SHA. If two or more people happen to be testing the same commit on the same machine, wouldn't they collide? One of them might remove the other's container. Also, two people running the same script at the same time will also be competing for the same hardware resources. I think we need to add some guard around this and only allow one CI run at a time, or maybe loudly write this in the Readme - "one user at a time".

bmhowe23 (approved):

docker/decoding-server/hw_ci/rxe-ofed/makefile
```diff
@@ -0,0 +1,72 @@
+# SPDX-License-Identifier: GPL-2.0
```
bmhowe23, line 1: We cannot check GPL-code into this repo.

bmhowe23 (changes requested):

PR #788, opened 2026-08-12
Allow a decoder to be constructed from a raw Stim DEM

melody-ren:
A workaround solution to enable Chromobius on the decoding server while the proper solution is being worked out.

A decoder entry may name a Stim detector error model instead of the matrix form of the same model. The decoder is then constructed from the DEM text, which is the only construction a DEM-native decoder such as Chromobius accepts, and derives its own dimensions and observable mapping from it. The matrix keys become optional at the parser and the checks that compare against syndrome_size move to the constructed decoder; the matrix path is otherwise unchanged.

The model text also reaches a nested global decoder through global_decoder_params, so trt_decoder can build Chromobius rather than handing it a parity check matrix it cannot use.

### DIFF
--- a/libs/qec/include/cudaq/qec/realtime/decoding_config.h
+++ b/libs/qec/include/cudaq/qec/realtime/decoding_config.h
@@ -72,6 +72,11 @@ struct decoder_config {
   std::optional<int> cuda_device_id;
   uint64_t block_size = 0;
   uint64_t syndrome_size = 0;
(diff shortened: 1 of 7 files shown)


tlshannon:
This was merged as part of https://github.com/NVIDIA/cudaqx/pull/801 (while Melody was OOO).

bmhowe23:
Closing since it was part of #801

PR #796, opened 2026-08-25
Deprecate Solvers

bmhowe23:
Deprecate Solvers in lieu of https://nvidia.github.io/cudaq-algorithms/. Future updates to this repo will be targeted at QEC only, and CUDA-Q Algorithms development will occur in https://github.com/NVIDIA/cudaq-algorithms.

### DIFF
--- a/docs/sphinx/api/solvers/gqe_api.rst
+++ b/docs/sphinx/api/solvers/gqe_api.rst
@@ -1,5 +1,3 @@
-.. include:: /_solvers_deprecation.rst
-
 .. function:: gqe(cost, pool, config=None, **kwargs)
 
     Run the Generative Quantum Eigensolver algorithm.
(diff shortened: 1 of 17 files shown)


docs/sphinx/api/solvers/gqe_api.rst
```diff
@@ -1,3 +1,5 @@
+.. include:: /_solvers_deprecation.rst
+
```
wsttiger: This file is never built standalone — it's in `exclude_patterns` and is only pulled into `python_api.rst` via `.. include:: gqe_api.rst`. Since `python_api.rst` already includes the deprecation notice at the top, this include renders the attention box a second time in the middle of the built Python API page. It also sits right where the `.. _gqe_api_python:` label points, so cross-references to the GQE function now land on the deprecation box instead of the `gqe()` docs.

Suggest dropping the include here and relying on the one at the top of `python_api.rst`:

```suggestion
```

wsttiger (approved):

docs/sphinx/api/solvers/gqe_api.rst
bmhowe23: Thanks. Addressed in 229a1969.

PR #802, opened 2026-08-27
[QEC] Add common decoder stats functions for ease of logging and replay

tlshannon:
## Description

Adds `decoder_stats`, the shared machinery behind the realtime `[DecoderStats]` log stream, so any decoder can report latency and replay fields in the line format `libs/qec/utils/replay_decoder_logs.py` already consumes.

`libs/qec/lib/decoder_stats.h` is deliberately **not** a public header: it sits beside its implementation rather than in the installed include tree, so it adds nothing to the decoder API and costs nothing for decoders that do not opt in. A decoder that wants instrumentation includes it, holds a `decoder_stats` member and adds `libs/qec/lib` to its private include directories.

At `info` it emits one lifetime summary line covering two latencies per shot -- tail (last measurements to end of decode) and full-shot (first measurements to the same point) -- as count, mean, min, max and p50/p90/p99. At `debug` it adds a per-call line with counts, durations and sparse arrays for replay capture. Percentiles come from a fixed-size log-spaced histogram rather than a sample buffer, since a realtime decoder cannot size a buffer up front and growing one mid-run would stall the path being measured; count, mean, min and max stay exact.

N/A

## Self-review checklist

Please confirm each item before requesting review. Check `[x]` or strike through and explain.

### Before requesting review
- [ ] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/libs/qec/include/cudaq/qec/realtime/nvtx_helpers.h
+++ b/libs/qec/include/cudaq/qec/realtime/nvtx_helpers.h
@@ -25,6 +25,7 @@ struct NvtxRange {
   cudaq::qec::realtime::experimental::NvtxRange _nvtx_range_##__LINE__(name)
 #define NVTX_PUSH(name) nvtxRangePushA(name)
 #define NVTX_POP() nvtxRangePop()
+#define NVTX_MARK(name) nvtxMarkA(name)
 
 } // namespace cudaq::qec::realtime::experimental
 #else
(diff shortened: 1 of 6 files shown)


bmhowe23 (approved):

bmhowe23:
Overriding the CI checks since this has passed other places before.

PR #803, opened 2026-08-28
Add CUDAQ_QEC_DECODERS_ONLY: build and test decoders without CUDA-Q

bmhowe23:
cmake -S libs/qec -B build -DCUDAQ_QEC_DECODERS_ONLY=ON builds the decoder interface (libcudaq-qec-decoders.so), the built-in decoders, and the decoder plugins with no CUDA-Q install and no LLVM/MLIR required; with CUDAQX_QEC_BINDINGS_PYTHON=ON it also builds a decoders-only python module (_qec_decoders_standalone, bindings under module.qecrt). fmt is fetched at the pin NVQIRConfig.cmake uses; cuStabilizer, libcudaq-qec.so (codes, experiments, DEM sampling), tools, realtime, and unit tests are skipped because only they need CUDA-Q. When the option is OFF every guard reduces to the existing behavior.

To let py_decoder.cpp compile without a CUDA-Q install, the CUDA-Q result-type casters (spin_op, sample_result, observe_result) move from type_casters.h into a new cudaq_type_casters.h, included by their only two consumers (py_code.cpp, py_surface_code.cpp), and py_decoder.cpp drops its unused cudaq/platform.h include. No behavior change for the full cudaq_qec module.

In decoders-only mode libs/qec/python/CMakeLists.txt returns early, so no cudaq_qec package files or install rules are generated: the module is only importable as bare _qec_decoders_standalone from <build>/python. That is intentional for now.

CI: a decoders-only build job in lib_qec.yaml (amd64 + arm64, cu13 devcontainer, no CUDA-Q build step) plus a smoke test that decodes a repetition-code syndrome via get_decoder("pymatching") through the decoders-only module and asserts cudaq is never imported.

### DIFF
--- a/.github/workflows/lib_qec.yaml
+++ b/.github/workflows/lib_qec.yaml
@@ -17,6 +17,57 @@ on:
         required: false
 
 jobs:
+  # Build the decoder stack with CUDAQ_QEC_DECODERS_ONLY=ON: no CUDA-Q
+  # install, no LLVM/MLIR, no cuStabilizer. Guards the decoders-only
+  # configuration against rot and demonstrates a decoder call through the
+  # decoders-only python module.
+  decoders-only-build:
+    name: Decoders-only build (no CUDA-Q)
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: ['amd64', 'arm64']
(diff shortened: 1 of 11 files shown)


tlshannon (approved):
LGTM

PR #811, opened 2026-09-01
Make proprietary cudevice archive optional and don't ship decoding_server

bmhowe23:
…in release

The decoding server is not part of the official release yet, so also strip bin/decoding_server* from the install prefix and install manifest before zipping the release artifacts. Wheels are unaffected: the qec-tools component was already excluded from install.components.


anjbur (approved):
LGTM

PR #812, opened 2026-09-01
[QEC] Reduce decode latency: resolve detectors as measurements arrive and reuse PyMatching scratch buffers

tlshannon:
## Description

Removes work from the final round of the realtime decode path. No public API change.

- **`decoder.cpp`**: `enqueue_syndrome()` now folds each detector in as soon as the
  last measurement it reads has arrived, instead of XOR-reducing the whole
  measurement buffer on the final round. Detectors are pre-sorted once in
  `set_D_sparse()` by how many measurement columns they need, and a cursor walks
  that order per round. Sorting on the column requirement rather than bucketing by
  round stays correct when rounds differ in width, as they do for the boundary
  layers of a memory circuit. A drain at decode time keeps the detector buffer
  complete by construction rather than by assumption. Each round is also copied
  with `memcpy` instead of a byte-at-a-time loop.
- **`pymatching.cpp`**: the three per-decode temporaries (detection events,
  observable frame, edge list) become reused members, so a steady-state `decode()`
  allocates nothing. Each is cleared at its use site because PyMatching XORs into
  the observable frame and appends to the edge list. Also drops a timing step that
  measured an empty interval (4 -> 3).

## Test plan
- New `QECCodeTester.checkRealtimeStreamedDetectorReduction`: streams a Steane
  memory-circuit window a round at a time over 100 random measurement records and
  requires the detector vector handed to the decoder to match a direct `m2d`
  reduction. Scoped in the comment to what it can observe: `decode()` only runs on
  a complete window, so it pins the reduction, not the fold order.
- `PyMatchingDecoder.DecodesHighObservableIndicesAcrossPaths` now decodes four
  times per instance, firing a different detector each time and revisiting one at
  the end. A single decode passed even without clearing the reused observable
  buffer; this fails without it.

### PyMatching on GB200
**Batch decode() — latencies in µs** (d=5, rounds=5, shots=1000)

build | p50 | p90 | p99 | shots/s | LER
-- | -- | -- | -- | -- | --
before | 0.35 | 0.90 | 2.02 | 2,109,349 | 0.0000
after | 0.32 ▼9% | 0.86 ▼4% | 1.98 ▼2% | 2,133,834 ▲1% | 0.0000

**Streaming tail (last round + corrections) — latencies in µs**
build | p50 | p90 | p99 | shots/s | LER
-- | -- | -- | -- | -- | --
before | 0.64 | 1.15 | 2.27 | 756,842 | 0.0000
after | 0.38 ▼41% | 0.93 ▼19% | 2.08 ▼8% | 854,479 ▲13% | 0.0000

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
- x ] User-visible behavior changes have public docs, or a follow-up is
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.

### DIFF
--- a/libs/qec/lib/decoder.cpp
+++ b/libs/qec/lib/decoder.cpp
@@ -12,7 +12,9 @@
 #include "cudaq/qec/logger.h"
 #include "cudaq/qec/plugin_loader.h"
 #include "cudaq/qec/version.h"
+#include <algorithm>
 #include <cassert>
+#include <cstring>
 #include <cuda_runtime_api.h>
 #include <dlfcn.h>
 #include <filesystem>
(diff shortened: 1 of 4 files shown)


vedika-saravanan (approved):
thanks, Tracy. LGTM

bmhowe23 (approved):

PR #816, opened 2026-09-02
Include the standalone decoders module in the cudaq_qec wheel

bmhowe23:
Builds on #803

Add a small SKBUILD-only block that also builds _qec_decoders_standalone (the same cudaq-free module the decoders-only build produces) and installs it into the wheel at the site-packages root. It links only cudaq-qec-decoders + cudart (no cudaq), giving an undocumented way to use the decoders without importing cudaq: `import _qec_decoders_standalone`. It is installed top-level, not under cudaq_qec/, on purpose -- importing anything under cudaq_qec runs cudaq_qec/__init__ -> patch.py -> `from cudaq import qvector`.

Purely additive and gated on SKBUILD, so non-wheel and decoders-only builds are unchanged. Add a small smoke test to the wheel test (test_wheels.sh): in a fresh interpreter, import the standalone module, decode with pymatching, and assert cudaq was never imported.

Note: the wheel does not bundle cudart, so this still needs a CUDA runtime present (from cuda-quantum, nvidia-cuda-runtime-cuXX, or a system CUDA).

### DIFF
--- a/libs/qec/python/CMakeLists.txt
+++ b/libs/qec/python/CMakeLists.txt
@@ -126,3 +126,28 @@ install(TARGETS ${MODULE_NAME}
   COMPONENT qec-python
   DESTINATION cudaq_qec/
 )
+
+# Undocumented: also ship the cudaq-free decoders module (_qec_decoders_standalone,
+# the same one the decoders-only build produces) inside the wheel, so the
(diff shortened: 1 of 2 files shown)


kvmto (approved):
looks good, just the CI problem to fix with the deprecation :)

PR #833, opened 2026-09-16
Add cancellation semantics to decoder

eliotheinrich:
## Description

This PR adds a `cancellation_source` containing shared cancellation state for ensemble decoders, which can be queried via a `cancellation_token`. This will be used for orchestrating ensembles of decoders under deadlines. Decoders gain an overloaded `decode(syndrome, token)` call, allowing them to provide an appropriate behavior for polling the token and aborting early, returning a nullopt. Default behavior is to ignore the token and delegate to `decode(syndrome)`. 

The trt_decoder inference does not provide an interface for cancellation, but it does pass the cancellation token to the global decoder. The sliding_window decoder does the same.

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


libs/qec/lib/decoders/sliding_window.cpp
```diff
@@ -317,14 +337,25 @@ std::vector<decoder_result> sliding_window::decode_batch(
...
   window_rounds.erase(window_rounds.begin(), window_rounds.begin() + step_size);
   CUDA_QEC_DBG("Returning empty decoder_result");
-  return std::vector<decoder_result>(); // empty return value
+  return std::vector<decoder_result>{}; // not ready yet
+}
+
+void sliding_window::reset_stream() {
+  window_rounds.clear();
+  rounds_since_last_reset = 0;
+  num_windows_decoded = 0;
```
tlshannon, line 348: Should rw_results also be cleared here?

eliotheinrich: Probably; previously, the buffer was reset during `initialize_window`, but clearing the buffer here keeps the streaming state consistent after a call to `reset_stream`. Fixed in [54720ba](https://github.com/NVIDIA/cudaqx/pull/833/commits/54720bab4278b267bc176ad6c697391386f7b590)

tlshannon (approved):
LGTM

PR #834, opened 2026-09-17
Add CPU RoCE to playback emulator

eliotheinrich:
## Description

This PR adds a CPU RoCE `session` backend to the playback emulator. 

Docs will follow (https://github.com/NVIDIA/cudaqx/issues/835).

N/A

## Self-review checklist

through and explain.

### Before requesting review
- [x] Full CI has been run.

### Scope and size

### Tests
      just when it is missing.

### Documentation
      tracked.

### Code style
      the area being modified.

### Dependencies
      OSRB tickets filed.


libs/qec/tools/playback-emulator/lib/session.cpp
```diff
@@ -713,4 +723,422 @@ make_udp_sessions(
...
+  /// (Sends consume our recv WQEs FIFO) but are matched by request_id.
+  bool poll_reply() {
+    const std::uint32_t slot = cursor_;
+    if (__atomic_load_n(&rx_flags_[slot], __ATOMIC_ACQUIRE) == 0)
+      return false;
+    std::memcpy(reply_scratch_.data(),
+                rx_data_ + static_cast<std::size_t>(slot) * stride_, stride_);
+    __atomic_store_n(&rx_flags_[slot], 0ull, __ATOMIC_RELEASE); // re-arm
+    cursor_ = (cursor_ + 1) & slot_mask_;
+    slot_owner_[slot] = kNoOwner;
```
vedika-saravanan: CUDA-Q’s existing RingSlotChannel serializes response-bearing calls and exits the dispatch on a timeout, while this implementation frees timed-out slots and continues publishing requests. A late unmatched response could therefore clear slot_owner_[slot] after that slot has been assigned to a newer request. Would it make sense to update ownership only after matching the response ID and confirming that slot_owner_[it->second.slot] still equals that ID? It may also be worth adding a delayed-response-after-timeout-and-ring-wrap test.

libs/qec/tools/playback-emulator/lib/session.cpp
```diff
@@ -713,4 +723,422 @@ make_udp_sessions(
...
+    collector_ = &collector;
+    // The transceiver's busy-polling RX/TX threads run from start() to stop().
+    monitor_ = std::thread([this] {
+      monitor_entered_.store(true, std::memory_order_release);
+      cpu_roce_blocking_monitor(xcvr_);
+    });
+    worker_ = std::thread([this] { worker_loop(); });
+  }
+
+  void send(const frame &f, tag t) override {
```
vedika-saravanan, line 878: Could we also validate f.size <= stride_ in send()? The current playback paths check max_frame_bytes upstream, but this session is the component that owns the fixed RDMA slot size, and try_publish() later copies frame_len bytes into that slot without another bound check. Keeping the invariant at this boundary would make the backend safer to reuse and straightforward to test with a slot_size + 1 frame.

eliotheinrich: Addressed in 85e1b1f

libs/qec/tools/playback-emulator/lib/session.cpp
eliotheinrich: Yes, good catch; addressed in 85e1b1f

vedika-saravanan:
Thanks, Eliot, my original comments look addressed. Could you help me confirm two details from the follow-up changes? With eight slots and max_inflight_ set to seven, should the late-reply test still expect the first eight records to time out? Also, is QEC_TEST_PLAYBACK_CPU_ROCE intentionally meant to control compiling the backend itself, or only running the hardware tests? I may be missing the intended build setup.

